PR #1, opened 2022-11-09
feat: new implementation

vasco-santos:
This PR adds a skeleton for SST project for w3up-api.

Details:
- added eslint with `hd-scripts` (no pretty for now)
- used npm until pnpm properly supports sst
- added a `/hello` route to have something we can deploy and guarantee seed.run is properly setup - will be removed in favour of further routes in follow up PR
- project deployed in `us-west-2`

Things to decide:
- [ ] naming: should we rename to `upload-api`?

### DIFF
--- a/stacks/api-stack.js
+++ b/stacks/api-stack.js
@@ -1,0 +1,16 @@
+import { Api } from '@serverless-stack/resources'
+
+/**
+ * @param {import('@serverless-stack/resources').StackContext} properties
+ */
+export function ApiStack({ stack }) {
+  const api = new Api(stack, 'api', {
+    routes: {
+      'GET /hello': 'functions/hello.handler',
+    },
+  })
+
+  stack.addOutputs({
+    ApiEndpoint: api.url,
+  })
+}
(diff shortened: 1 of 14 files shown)


stacks/api-stack.js
```diff
@@ -0,0 +1,16 @@
+import { Api } from '@serverless-stack/resources'
+
+/**
+ * @param {import('@serverless-stack/resources').StackContext} properties
+ */
+export function ApiStack({ stack }) {
+  const api = new Api(stack, 'api', {
```
olizilla, line 7: minor: this resource are going get auto named `w3up-api-api`

vasco-santos: I did not use default service name because `service` is already used in a bunch of places to name the ucanto remote service. Any other naming suggestions we could use?

olizilla (commented):
Let's add tsconfig and a step to verify our jsdoc type comments are accurate.

stacks/api-stack.js
olizilla: `http-gateway` is the actual aws resource that we's be creating so we could go with that

olizilla (approved):
🚀

PR #4, opened 2022-11-11
feat: ucanto invocation router with bare minimal store add invocation

vasco-santos:
This PR adds ucanto invocation router with `store/add` invocation.

It integrates following AWS Resources in ApiStack:
- S3 Bucket to store CAR files (carpark)
  - we could have re-used the current bucket `w3up-cars`, but it is in wrong region, so let's just be coherent with CF naming and use `carpark-prod-0` ? :)
- Dynamo DB with `store_table` to track mapping between Users and stored CARs
  - keeping same namings as current schema. We will need to perform a migration, so you might decide to rename things. Let's consider it later on new PRs

Notes:
- previous `services` folder renamed into API, so that we can use `service` folder per ucanto framework current used names
- uses hardcoded service DID, we need in follow up PR to wire it up with env var - also see `authority.js` + `config.js` in w3up repo
- we add info to `store_table` like previously done in `w3up` repo. Probably worth considering as follow up to add a column that tracks wether the CAR already made it to the bucket. On S3 Put Event, we would update that boolean. Otherwise, we have things in the DB that we don't have immediate visibility if the given CARs ever made it

Follow up needs (will create issues):
- handle account from `store/add`
- wire up service DID with secrets


api/test/functions/ucan-invocation-router.test.js
```diff
@@ -0,0 +1,42 @@
...
+import { alice } from '../fixtures.js'
+
+test.before(async t => {
+  await new Container('amazon/dynamodb-local:latest')
+    .withExposedPorts(8000)
+    .start()
+})
+
+// TODO: Need to set ENV for dbEndpoint...
+test.skip('ucan-invocation-router', async (t) => {
```
vasco-santos, line 19: @olizilla should we just drop this test to the lambda directly?

It is quite hard to inject the needed resources here... I would suggest follow up PR with more integration kind of tests, where we run the API Stack with all routes and perform tests using HTTP Calls.

package.json
```diff
@@ -37,9 +52,13 @@
     ],
     "parserOptions": {
       "project": "./tsconfig.json"
+    },
+    "rules": {
+      "no-console": "off",
+      "no-warning-comments": "off"
```
vasco-santos, line 57: removes warnings for console.logs

api/service/store/add.js
```diff
@@ -0,0 +1,72 @@
...
+        return new Server.MalformedCapability(
+          invocation.capabilities[0],
+          new Server.Failure('Provided capability has no link')
+        )
+      }
+
+      // Only use capability account for now to check if account is registered.
+      // This must change to access account/info!!
+      // We need to use https://github.com/web3-storage/w3protocol/blob/9d4b5bec1f0e870233b071ecb1c7a1e09189624b/packages/access/src/agent.js#L270
+      const account = capability.with
```
vasco-santos, line 32: Will handle account/agent stuff in follow up PR

stacks/api-stack.js
```diff
@@ -1,12 +1,61 @@
...
+   * to associate the uploaders DID with a payload CID.
+   * You can also optionally indicate an application DID, origin and size
+   */
+   const storeTable = new Table(stack, 'store_table', {
+    fields: {
+      uploaderDID: 'string',
+      payloadCID: 'string',
+      applicationDID: 'string',
+      origin: 'string',
+      size: 'number',
```
olizilla, line 27: is it size in bytes?

stacks/api-stack.js
```diff
@@ -1,12 +1,61 @@
...
+  const stackConfig = getConfig(stack.stage)
+
+  /**
+   * This table takes a stored CAR and makes an entry in the store table
+   * to associate the uploaders DID with a payload CID.
+   * You can also optionally indicate an application DID, origin and size
+   */
+   const storeTable = new Table(stack, 'store_table', {
+    fields: {
+      uploaderDID: 'string',
```
olizilla, line 23: - Is this the account DID or the agent DID?
- camelCase?

stacks/api-stack.js
```diff
@@ -1,12 +1,61 @@
...
+    primaryIndex: { partitionKey: 'uploaderDID', sortKey: 'payloadCID' },
+    ...stackConfig.tableConfig,
+  })
+
+  /**
+   * This bucket is the car store used within the `store/*` invocations.
+   * Force name of bucket to be "car-park-prod-0" in prod.
+   */
+  const ingestBucketCdkConfig =
+    stack.stage === 'prod' ? { cdk: { bucket: { bucketName: 'car-park-prod-0' } } } : {}
```
olizilla, line 40: Some of our per env config is in config.js. We should try and be consistent

stacks/api-stack.js
```diff
@@ -1,12 +1,61 @@
...
+  
+  const carStore = new Bucket(stack, 'carStore', {
```
olizilla, line 42: in this file we have
- kebab-case
- camelCase
- snake_case

for resource ids. plz can we not.

stacks/api-stack.js
```diff
@@ -1,12 +1,61 @@
...
+    ...ingestBucketCdkConfig
+  })
+
   const api = new Api(stack, 'http-gateway', {
+    defaults: {
+      function: {
+        permissions: [storeTable, carStore],
+        environment: {
+          STORE_TABLE_NAME: storeTable.tableName,
+          CAR_STORE_BUCKET_NAME: carStore.bucketName
```
olizilla, line 53:
```suggestion
          STORE_BUCKET_NAME: carStore.bucketName
```

api/functions/ucan-invocation-router.js
```diff
@@ -0,0 +1,90 @@
+import * as Server from '@ucanto/server'
+import * as CAR from '@ucanto/transport/car'
+import * as CBOR from '@ucanto/transport/cbor'
+
+import getServiceDid from '../authority.js'
+import { create as createCarStore } from '../buckets/car-store.js'
```
olizilla, line 6: let's make the export from car-store be `createCarStore`

api/functions/ucan-invocation-router.js
```diff
@@ -0,0 +1,90 @@
...
+import { StoreTable } from '../database/store.js'
+import { createServiceRouter } from '../service/index.js'
+
+const AWS_ACCESS_KEY_ID = process.env.AWS_ACCESS_KEY_ID || ''
+const AWS_SECRET_ACCESS_KEY = process.env.AWS_SECRET_ACCESS_KEY || ''
+const AWS_SESSION_TOKEN = process.env.AWS_SESSION_TOKEN || ''
+const AWS_REGION = process.env.AWS_REGION || 'us-west-2'
+
+/**
+ * AWS API Gateway handler for POST / with ucan invocation router.
```
olizilla, line 16:
```suggestion
 * AWS HTTP Gateway handler for POST / with ucan invocation router.
```

api/functions/ucan-invocation-router.js
```diff
@@ -0,0 +1,90 @@
...
+      endpoint: dbEndpoint
+    }),
+    carStore: createCarStore(AWS_REGION, bucketName),
+    signingOptions: {
+      region: AWS_REGION,
+      secretAccessKey: AWS_SECRET_ACCESS_KEY,
+      accessKeyId: AWS_ACCESS_KEY_ID,
+      sessionToken: AWS_SESSION_TOKEN,
+      bucket: bucketName,
+    }
```
olizilla, line 48: pass in a UrlSigner or similar to follow the pattern of holding the state in an object that we set in StoreTable and CarStore

api/functions/ucan-invocation-router.js
```diff
@@ -0,0 +1,90 @@
...
+  })
+  const response = await server.request({
+    // @ts-ignore - type is Record<string, string|string[]|undefined>
```
olizilla, line 51:
```suggestion
    // @ts-expect-error - type is Record<string, string|string[]|undefined>
```

also i think we should review the type sig that ucanto server is using here. that seems like a reasonable thing to pass in.

api/database/store.js
```diff
@@ -0,0 +1,92 @@
...
+    } catch {
+      return false
+    }
+  }
+
+  /**
+   * Bind a link CID to an account
+   *
+   * @param {object} data
+   * @param {string} data.accountDID
```
olizilla, line 65: is really string type?

olizilla (approved):
Good start. Lots of nitpicks that we can iterate on together. GOOOOO VASCO!

stacks/api-stack.js
vasco-santos: Yes, see https://github.com/web3-storage/w3protocol/pull/160/files

PR #10, opened 2022-11-15
fix: tidy up types

vasco-santos:
This PR tidies up a bit types. Also:

-  `origin` is a Link (CID) per https://github.com/web3-storage/w3protocol/pull/160/files so updating it.
- augmented test to check Dynamo DB content was inserted


vasco-santos:
merging this as I still need it in my other PR and is just tidying up types

PR #14, opened 2022-11-15
feat: store star handlers

### LINKED ISSUE #11: Implement store/* providers
We have `store/add` implemented, let's get the remaining ones in: `store/remove` and `store/list`

vasco-santos:
This PR adds the remaining `store/*` handlers: list + remove.

As follow up, created issue https://github.com/web3-storage/upload-api/issues/16 to handle the cursor moving around - this needs to be updated in spec, [access](https://github.com/web3-storage/w3protocol/blob/main/packages/access/src/capabilities/store.js#L68) and upload-api

Closes #11

### DIFF
--- a/api/service/store/remove.js
+++ b/api/service/store/remove.js
@@ -1,0 +1,24 @@
+import * as Server from '@ucanto/server'
+import * as Store from '@web3-storage/access/capabilities/store'
+
+/**
+ * @typedef {import('@ucanto/interface').Link<unknown, number, number, 0 | 1>} Link
+ */
+
+/**
+ * @param {import('../types').StoreServiceContext} context
+ */
+export function storeRemoveProvider(context) {
+  return Server.provide(
+    Store.remove,
+    async ({ capability }) => {
+      const { link } = capability.nb
+      // Only use capability account for now to check if account is registered.
+      // This must change to access account/info!!
+      // We need to use https://github.com/web3-storage/w3protocol/blob/9d4b5bec1f0e870233b071ecb1c7a1e09189624b/packages/access/src/agent.js#L270
+      const account = capability.with
+
+      await context.storeTable.remove(account, link.toString())
+    }
+  )
+}
--- a/api/tables/store.js
+++ b/api/tables/store.js
@@ -78,16 +78,16 @@ export function createStoreTable (region, tableName, options = {}) {
      * @param {string} payloadCID
      */
     remove: async (uploaderDID, payloadCID) => {
-      const params = {
+      const cmd = new DeleteItemCommand({
         TableName: tableName,
         Key: marshall({
           uploaderDID,
           payloadCID,
         }),
         AttributesToGet: ['uploaderDID'],
-      }
+      })
   
-      await dynamoDb.send(new DeleteItemCommand(params))
+      await dynamoDb.send(cmd)
     },
     /**
      * List all CARs bound to an account
@@ -96,7 +96,7 @@ export function createStoreTable (region, tableName, options = {}) {
      * @param {import('../service/types').ListOptions} [options]
      */
     list: async (uploaderDID, options = {}) => {
-      const params = {
+      const cmd = new QueryCommand({
         TableName: tableName,
         Limit: options.pageSize || 20,
         KeyConditions: {
@@ -106,8 +106,8 @@ export function createStoreTable (region, tableName, options = {}) {
           },
         },
         AttributesToGet: ['payloadCID', 'size', 'origin', 'uploadedAt'],
-      }
-      const response = await dynamoDb.send(new QueryCommand(params))
+      })
+      const response = await dynamoDb.send(cmd)
 
       /** @type {import('../service/types').StoreListResult[]} */
       // @ts-expect-error
--- a/api/tables/store.js
+++ b/api/tables/store.js
@@ -83,8 +83,7 @@ export function createStoreTable (region, tableName, options = {}) {
         Key: marshall({
           uploaderDID,
           payloadCID,
-        }),
-        AttributesToGet: ['uploaderDID'],
+        })
       })
   
       await dynamoDb.send(cmd)
--- a/api/tables/store.js
+++ b/api/tables/store.js
@@ -2,8 +2,10 @@ import {
   DynamoDBClient,
   GetItemCommand,
   PutItemCommand,
+  DeleteItemCommand,
+  QueryCommand
 } from '@aws-sdk/client-dynamodb'
-import { marshall } from '@aws-sdk/util-dynamodb'
+import { marshall, unmarshall } from '@aws-sdk/util-dynamodb'
 
 /**
  * Abstraction layer to handle operations on Store Table.
@@ -12,6 +14,7 @@ import { marshall } from '@aws-sdk/util-dynamodb'
  * @param {string} tableName
  * @param {object} [options]
  * @param {string} [options.endpoint]
+ * @returns {import('../service/types').StoreTable}
  */
 export function createStoreTable (region, tableName, options = {}) {
   const dynamoDb = new DynamoDBClient({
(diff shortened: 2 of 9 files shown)


api/test/service/store.test.js
```diff
@@ -118,9 +118,161 @@ test('store add returns done if already uploaded', async (t) => {
...
+  t.truthy(item?.uploaderDID)
+})
+
+test('store remove does not fail for non existent link', async (t) => {
+  const server = await createStoreUcantoServer(t.context)
+  const account = alice.did()
+  const data = new Uint8Array([11, 22, 34, 44, 55])
+  const link = await CAR.codec.link(data)
+
+  const request = await CAR.encode([
```
alanshaw, line 132: What is the reason for using CAR encoder here? You lose the type validation from the capability if you manually create the invocation like this and also have to decode the response. For example, why not do this instead:

```js
import { connect } from '@ucanto/client'
import { CAR, CBOR } from '@ucanto/transport'
import * as StoreCapabilities from '@web3-storage/access/capabilities/store'

const conn = connect({
  id: t.context.serviceDid,
  encoder: CAR,
  decoder: CBOR,
  channel: server
})

const storeRemove = await StoreCapabilities.remove
    .invoke({
      issuer: alice,
      audience: parse(t.context.serviceDid.did()),
      capabilities: [{
        can: 'store/remove',
        with: account,
        nb: { link },
      }],
      proofs: [],
    })
    .execute(conn)
```

api/tables/store.js
```diff
@@ -67,6 +70,61 @@ export function createStoreTable (region, tableName, options = {}) {
...
+     *
+     * @param {string} uploaderDID
+     * @param {string} payloadCID
+     */
+     remove: async (uploaderDID, payloadCID) => {
+      const params = {
+        TableName: tableName,
+        Key: marshall({
+          uploaderDID: uploaderDID.toString(),
+          payloadCID: payloadCID.toString(),
```
alanshaw, line 85: Already strings according to the method params?

alanshaw (changes requested):

alanshaw:
😉😉😉 "store star" sounds cooler than "store asterisk"

api/test/service/store.test.js
vasco-santos: I know, the blocker for that is how to spin the SST resources in that way of doing tests. I am expecting to use client in integration tests (which performs capability invocation like you are suggesting).

However, for unit tests kind of interaction, we need to interact directly with the server instance and call request after creating it.

api/tables/store.js
vasco-santos: good catch, I moved things around and forgot these

api/test/service/store.test.js
```diff
@@ -72,9 +72,9 @@ test('store add returns signed url for uploading', async (t) => {
 
   const item = await getItemFromStoreTable(t.context.dynamoClient, alice, link)
   t.truthy(item)
-  t.truthy(item.uploadedAt)
-  t.truthy(item.proof)
-  t.truthy(item.uploaderDID)
+  t.truthy(item?.uploadedAt)
+  t.truthy(item?.proof)
+  t.truthy(item?.uploaderDID)
```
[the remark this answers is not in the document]
vasco-santos: Made type check and added size as well

api/tables/store.js
```diff
@@ -67,6 +70,61 @@ export function createStoreTable (region, tableName, options = {}) {
...
   
       return item
+    },
+    /**
+     * Unbinds a link CID to an account
+     *
+     * @param {string} uploaderDID
+     * @param {string} payloadCID
+     */
+     remove: async (uploaderDID, payloadCID) => {
```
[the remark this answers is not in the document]
vasco-santos: oh no, CI is happy with it :(

api/service/store/remove.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param {import('../types').StoreServiceContext} context
+ */
+export function storeRemoveProvider(context) {
+  return Server.provide(
+    Store.remove,
+    async ({ capability, invocation }) => {
+      const { link } = capability.nb
+
+      if (!link) {
+        return new Server.MalformedCapability(
```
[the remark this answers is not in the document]
vasco-santos: Oh ok, I did not know this. I was following old API code for it. That makes sense, I am removing

vasco-santos:
> 😉😉😉 "store star" sounds cooler than "store asterisk"

I like asterisk more actually, but renamed for consistency

api/test/service/store.test.js
olizilla: @vasco-santos i think Alan is saying, you can pass the uncanto server **object** directly to a uncanto client and then use the client to test different scenarios. No http requests occur, not http api gateway needed. The client object uses the server object directly.

vasco-santos: In that scenario, we get limited by the things we can test on having the client to support them (example: list with pagination, ...)

I think we should use the client, but in integration tests, not unit

api/test/service/store.test.js
```diff
@@ -72,9 +73,11 @@ test('store add returns signed url for uploading', async (t) => {
...
   const item = await getItemFromStoreTable(t.context.dynamoClient, alice, link)
   t.truthy(item)
-  t.truthy(item.uploadedAt)
-  t.truthy(item.proof)
-  t.truthy(item.uploaderDID)
+  t.is(typeof item?.uploadedAt, 'string')
+  t.is(typeof item?.proof, 'string')
+  t.is(typeof item?.uploaderDID, 'string')
+  t.truthy(DID.parse(item?.uploaderDID))
+  t.is(typeof item?.size, 'number')
```
olizilla, line 80: maybe check that it's the DID and size you expect it to be here?

api/test/service/store.test.js
```diff
@@ -117,10 +120,163 @@ test('store add returns done if already uploaded', async (t) => {
...
   const item = await getItemFromStoreTable(t.context.dynamoClient, alice, link)
-  t.truthy(item)
-  t.truthy(item.uploadedAt)
-  t.truthy(item.proof)
-  t.truthy(item.uploaderDID)
+  t.is(typeof item?.uploadedAt, 'string')
+  t.is(typeof item?.proof, 'string')
+  t.is(typeof item?.uploaderDID, 'string')
+  t.truthy(DID.parse(item?.uploaderDID))
+  t.is(typeof item?.size, 'number')
```
olizilla, line 127: check DID is the one you expect

olizilla (approved):
👍

api/test/service/store.test.js
```diff
@@ -117,10 +120,163 @@ test('store add returns done if already uploaded', async (t) => {
...
+})
+
+test('store remove does not fail for non existent link', async (t) => {
```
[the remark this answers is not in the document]
vasco-santos: Discussion time in https://github.com/web3-storage/upload-api/issues/21

vasco-santos:
Addressed Oli's nits and merging given it is approved and also addressed all Alan's comments. 

@alanshaw we can address [this](https://github.com/web3-storage/upload-api/pull/14#discussion_r1024156583) differently if you have more feedback in follow up :)

api/test/service/store.test.js
```diff
@@ -72,9 +72,13 @@ test('store add returns signed url for uploading', async (t) => {
 
   const item = await getItemFromStoreTable(t.context.dynamoClient, alice, link)
   t.truthy(item)
-  t.truthy(item.uploadedAt)
-  t.truthy(item.proof)
-  t.truthy(item.uploaderDID)
+  t.is(typeof item?.uploadedAt, 'string')
+  t.is(typeof item?.proof, 'string')
+  t.is(typeof item?.uploaderDID, 'string')
+  t.is(item?.uploaderDID, account)
```
olizilla, line 78: account is now a bit of a trigger word for me. I see `account` and i mentally swap it with `space` and that trips me up here. Can we do
```suggestion
  t.is(item?.uploaderDID, alice.did())
```
and drop the word `account` from the tests pls?

api/test/service/store.test.js
alanshaw: If you want to test capabilities that are not in the access client _yet_ then you can always create them here in the same way as the access client creates them. You also get the benefit of being able to copy/pasta them to the client after you've defined them here and tested them.

vasco-santos: Let's sync on this tomorrow and decide

olizilla: i make it so in #23 and i like it

PR #15, opened 2022-11-15
feat: validate store/add access

alanshaw:
This PR adds a request to w3access when a `store/add` invocation is handled.

The request to w3access ensures that the target space has been verified by a user via email.

### DIFF
--- a/api/functions/ucan-invocation-router.js
+++ b/api/functions/ucan-invocation-router.js
@@ -1,8 +1,10 @@
 import * as Server from '@ucanto/server'
 import * as CAR from '@ucanto/transport/car'
 import * as CBOR from '@ucanto/transport/cbor'
+import { DID } from '@ucanto/core'
 
 import getServiceDid from '../authority.js'
+import { createAccess } from '../access.js'
 import { createSigner } from '../signer.js'
 import { createCarStore } from '../buckets/car-store.js'
 import { createStoreTable } from '../tables/store.js'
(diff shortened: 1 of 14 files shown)


vasco-santos (commented):
Yes! I Like it

olizilla:
@alanshaw plz to add PR description or link to issue or both!

api/service/store/add.js
```diff
@@ -28,22 +28,25 @@ export function storeAddProvider(context) {
...
-      const account = capability.with
 
+      const resource = Server.DID.parse(capability.with).did()
       const [
+        verified,
         carIsLinkedToAccount,
         carExists
       ] = await Promise.all([
-        context.storeTable.exists(account, link.toString()),
+        context.access.verifyInvocation(invocation),
```
olizilla, line 38: do `context.access.verifyInvocation` before we do any other work? db and bucket reads have a smol $ cost.

api/functions/ucan-invocation-router.js
```diff
@@ -35,7 +39,9 @@ async function ucanInvocationRouter (request) {
     }
   }
 
-  const server = await createUcantoServer({
+  const id = await getServiceDid()
```
olizilla, line 42: plz long name like
```suggestion
  const uploadApiDid = await getServiceDid()
```

olizilla (approved):
👍

api/service/store/add.js
alanshaw: IDK, I'm optimistically hoping that requests are going to have access to do this.

vasco-santos (commented):
Can you extend this to `store/*`? and maybe `upload/*` if this lands after https://github.com/web3-storage/w3protocol/pull/184

api/test/utils.js
```diff
@@ -77,3 +96,81 @@ export function getSigningOptions(ctx) {
     bucket: ctx.bucketName,
   }
 }
+
+/** @returns {Promise<MockAccess>} */
+export async function createAccessServer () {
```
vasco-santos, line 101: Might be nice to consider creating a mock module for this as I expect it to be needed in more services. Not for now, just raising thoughts for visibility :)

vasco-santos (approved):
LGTM! 🎉  Thanks

olizilla:
I had to make similar changes to createUcantoServer in #35 so let's get this one merged and I can deal with the conflicts over there.

PR #26, opened 2022-11-17
feat: upload star handlers

### LINKED ISSUE #12: Implement upload/* providers

vasco-santos:
This PR adds `upload/*` handlers as well as their tests.

It includes
- new Table resource with upload listing (follows same schema as w3up)
- uploadTable abstraction like storeTable
- update access to fix `upload/add` type inference
- random car helper function stolen from client tests
- updated `store` tests to use `createSpace` and `getClientConnection` helpers

Note that:
- pagination will be handled as follow up
- removes unused fixtures folder
- dynamoDB does not support deleting items without providing entire key.
  - Upload table key is `primaryIndex: { partitionKey: 'uploaderDID', sortKey: 'sk' }` to be unique. However, when `upload/remove` is triggered we receive `uploaderDID` and `dataCID` and have no way to get `sk` ('dataCID#carCID'). 
  - We need to perform an initial query to get items from DB, and then batch deleteRequests. It is also important mentioning that Dynamo brings more [fun to the table](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/classes/batchwriteitemcommand.html). Batch has some limitations, I am not worried with size but we might have more than 25 items to batch delete. So, we need to guarantee that batches bigger than it are not created, iterating in loop if more
  - Note: same applies for batch insert on `upload/add`

Still todo:
- [x] figure out remove

Closes #12

### DIFF
--- a/api/tables/upload.js
+++ b/api/tables/upload.js
@@ -1,0 +1,169 @@
+import {
+  DynamoDBClient,
+  BatchWriteItemCommand,
+  GetItemCommand,
+  QueryCommand
+} from '@aws-sdk/client-dynamodb'
+import { marshall, unmarshall } from '@aws-sdk/util-dynamodb'
+
+// https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/classes/batchwriteitemcommand.html
+export const BATCH_MAX_SAFE_LIMIT = 25
+
+/**
+ * Abstraction layer to handle operations on Upload Table.
+ *
+ * @param {string} region
+ * @param {string} tableName
+ * @param {object} [options]
+ * @param {string} [options.endpoint]
+ * @returns {import('../service/types').UploadTable}
+ */
+export function createUploadTable (region, tableName, options = {}) {
+  const dynamoDb = new DynamoDBClient({
+    region,
+    endpoint: options.endpoint
+  })
+
+  return {
+    /**
+     * Check if the given data CID is bound to the uploader DID
+     *
+     * @param {string} uploaderDID
+     * @param {string} dataCID
+     */
+     exists: async (uploaderDID, dataCID) => {
+      const cmd = new GetItemCommand({
+        TableName: tableName,
+        Key: marshall({
+          uploaderDID,
+          dataCID,
+        }),
+        AttributesToGet: ['uploaderDID'],
+      })
+  
+      try {
+        const response = await dynamoDb.send(cmd)
+        return response?.Item !== undefined
+      } catch {
+        return false
+      }
+    },
+    /**
+     * Link an upload to an account
+     *
+     * @param {string} uploaderDID
+     * @param {import('../service/types').UploadItemInput} item
+     */
+    insert: async (uploaderDID, { dataCID, carCIDs }) => {
+      /** @type {import('../service/types').UploadItemOutput[]} */
+      const items = []
+      const batchItems = []
+
+      // Create returning array and array to batch
+      for (const carCID of carCIDs) {
+        const item = {
+          uploaderDID,
+          dataCID,
+          carCID,
+          uploadedAt: new Date().toISOString(),
+        }
+
+        items.push(item)
+        batchItems.push({
+          ...item,
+          // add sk property for Dynamo Key uniqueness
+          sk: `${item.dataCID}#${item.carCID}`
+        })
+      }
+
+      // Batch writes with max safe limit
+      while (batchItems.length > 0) {
+        const currentBatchItems = batchItems.splice(0, BATCH_MAX_SAFE_LIMIT)
+        const cmd = new BatchWriteItemCommand({
+          RequestItems: { [tableName]: currentBatchItems.map(item => ({
+            PutRequest: {
+              Item: marshall(item)
+            }
+          }))},
+        })
+        await dynamoDb.send(cmd)
+      }
+      
+      return items
+    },
+    /**
+     * Remove an upload from an account
+     *
+     * @param {string} uploaderDID
+     * @param {string} dataCID
+     */
+    remove:  async (uploaderDID, dataCID) => {
+      let lastEvaluatedKey
+      // Iterate through all carCIDs mapped to given uploaderDID
+      do {
+        // Get first batch of items to remove
+        const queryCommand = new QueryCommand({
+          TableName: tableName,
+          Limit: BATCH_MAX_SAFE_LIMIT,
+          ExpressionAttributeValues: {
+            ':u': { S: uploaderDID },
+            ':d': { S: dataCID }
+          },  
+          KeyConditionExpression: 'uploaderDID = :u',
+          FilterExpression: 'contains (dataCID, :d)',
+          ProjectionExpression: 'uploaderDID, sk'
+        })
+        const queryResponse = await dynamoDb.send(queryCommand)
+        // Update cursor if existing
+        lastEvaluatedKey = queryResponse.LastEvaluatedKey
+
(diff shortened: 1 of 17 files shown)


api/test/service/store.test.js
```diff
@@ -174,21 +171,9 @@ test('store remove does not fail for non existent link', async (t) => {
 
   // expect no response for a remove
   t.falsy(storeRemove)
-
-  const storeRemove2 = await StoreCapabilities.remove.invoke({
```
vasco-santos, line 178: this looks not needed added in previous PR?

vasco-santos: cc @olizilla

api/test/service/upload.test.js
```diff
@@ -0,0 +1,311 @@
...
+ * @param {import("@ucanto/principal/dist/src/ed25519/type.js").EdSigner<"key">} audience
+ * @param {`${string}:${string}`} resource
+ */
+async function getSpaceDelegationProof (issuer, audience, resource) {
+  return await UcantoClient.delegate({
+    issuer,
+    audience,
+    capabilities: [{ can: '*', with: resource }]
+  })
+}
```
vasco-santos, line 269: @olizilla feels way cleaner this way then having to replicate this across the board. what do you think?

api/test/service/store.test.js
olizilla: no it was deliberate. I wanted the test to show that trying to remove the same thing twice works.

api/test/service/upload.test.js
olizilla: yep. is good. I would go further and have the function create the `space` signer too, which is what the `createSpace` function of the agent would do.

api/test/service/upload.test.js
```diff
@@ -0,0 +1,311 @@
...
+    channel: await createTestingUcantoServer(context),
+  })
+}
+
+/**
+ * @param {import("@ucanto/principal/dist/src/ed25519/type.js").EdSigner<"key">} issuer
+ * @param {import("@ucanto/principal/dist/src/ed25519/type.js").EdSigner<"key">} audience
+ * @param {`${string}:${string}`} resource
+ */
+async function getSpaceDelegationProof (issuer, audience, resource) {
```
olizilla, line 263: something like
```js
async function createSpace (audience): { spaceDid, proof } 
```

olizilla: we dont need the space signer outside of the function, as we provide the proof that space delegates `*` to `audience`

api/test/service/store.test.js
vasco-santos: same non existent thing? not sure I understand what happens differently. Perhaps you wanted to remove after a successful remove?

api/test/service/upload.test.js
vasco-santos: yes!

api/test/service/store.test.js
olizilla: I see "dont error if delete called on missing thing" as "be idempotent". This test verifies that we can call remove twice with the same cid and get the same result.

api/tables/upload.js
```diff
@@ -0,0 +1,169 @@
...
+          uploadedAt: new Date().toISOString(),
+        }
+
+        items.push(item)
+        batchItems.push({
+          ...item,
+          // add sk property for Dynamo Key uniqueness
+          sk: `${item.dataCID}#${item.carCID}`
+        })
+      }
```
olizilla, line 77: pull up uploadedAt to create once so all definitely get the same timestamp, and this seems like a simple `map` so would be a little clearer as
```suggestion
      const uploadedAt = new Date().toISOString(),

      const items = carCIDs.map(carCID => ({
        uploaderDID,
        dataCID,
        carCID,
        uploadedAt
      })

      // items formatted for dynamodb
      const batchItems = items.map(item => ({
        ...item,
        // add sk property for Dynamo Key uniqueness
        sk: `${item.dataCID}#${item.carCID}`
      })
```

api/test/service/upload.test.js
```diff
@@ -0,0 +1,367 @@
...
+    // @ts-expect-error ʅʕ•ᴥ•ʔʃ
+  }).execute(connection)
+
+  // @ts-expect-error error is added by ucanto if it fails
+  t.not(uploadAdd.error, true, uploadAdd.message)
+  t.is(uploadAdd.length, shards.length)
+
+  // Validate shards result
+  for (const shard of shards) {
+    const shardResult = uploadAdd.find((s) => s.carCID === shard.toString())
```
olizilla, line 61: minor: we should assert what we expect the order of shards to be here. not because we really care... _(tho i would want a test to fail if I changed that without meaning to)..._ but more because we should be very explicit about what we expect to see in the results.

if we sent `[shardA, shardA, shardB]` and the response contained `[shardA, shardB, shardB]`, this test would still pass. the length matches and we can find the item we want in it, but it is not correct.

api/test/service/upload.test.js
```diff
@@ -0,0 +1,367 @@
...
+  // Validate shards result
+  for (const shard of shards) {
+    const shardResult = dbItems.find((s) => s.carCID === shard.toString())
+    t.is(shardResult?.dataCID, root.toString())
+    t.truthy(shardResult?.uploadedAt)
+  }
+})
+
+// TODO: this is current behavior with optional nb.
+// We should look into this as a desired behavior
```
olizilla, line 80: plz to make issue

api/test/service/upload.test.js
```diff
@@ -0,0 +1,367 @@
...
+}
+
+/**
+ * @param {import("@aws-sdk/client-dynamodb").DynamoDBClient} dynamo
+ * @param {`did:key:${string}`} spaceDid
+ */
+ async function getSpaceItems(dynamo, spaceDid) {
+  const cmd = new QueryCommand({
+    TableName: 'upload',
+    Limit: 30,
```
olizilla, line 334:
```suggestion
 async function getUploadsForSpace(dynamo, spaceDid, limit = 30) {
  const cmd = new QueryCommand({
    TableName: 'upload',
    Limit: limit,
```

api/test/service/upload.test.js
```diff
@@ -0,0 +1,367 @@
...
+    proofs: [ proof ]
+    // @ts-expect-error ʅʕ•ᴥ•ʔʃ
+  }).execute(connection)
+
+  t.is(uploadList.pageSize, cars.length)
+
+  // Validate entries have given CARs
+  for (const entry of uploadList.results) {
+    t.truthy(cars.find(car => car.roots[0].toString() === entry.dataCID ))
+  }
```
olizilla, line 303: plz validate expected order of results list

olizilla (approved):
approving to unblock. minor tweaks suggested

api/test/service/upload.test.js
vasco-santos: https://github.com/web3-storage/w3protocol/issues/186

api/test/service/upload.test.js
vasco-santos: I will merge this as this needs discussion. 

If you consider `primaryIndex: { partitionKey: 'uploaderDID', sortKey: 'sk' },` the sortKey is `sk` which is `'dataCID#carCID'`. So there is no guarantee that the order is the order of insertion. But, I also don't see any requirement or need to keep an order. So, I consider that seeing we have length, and all are there is the expectation.

api/test/service/upload.test.js
vasco-santos: see https://github.com/web3-storage/upload-api/pull/26#discussion_r1026314813

PR #28, opened 2022-11-18
feat: store and upload list with pagination support

### LINKED ISSUE #16: Add store/list cursor for pagination
Follow up from https://github.com/web3-storage/upload-api/pull/14 to allow pagination

vasco-santos:
Adds pagination for `store/list` and `upload/list` relying on a `cursor` and `size` capability invocation parameters.

Updates `access-client` to 6.0.0 https://github.com/web3-storage/w3protocol/releases/tag/access-v6.0.0 . Most notably, account was renamed to space and these changes are reflected here as well

Needs:
- [x] https://github.com/web3-storage/w3protocol/pull/184 - note that CI is green if locally we npm link dependency. This is because types are not generated with current dependency

Closes #16


alanshaw (approved):
The change here looks good to me, but I think we want the items to be ordered by the datetime they were added...right?

alanshaw:
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/SecondaryIndexes.html

Maybe a local secondary? https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LSI.html

vasco-santos:
@alanshaw good shout. I did not know about these local indexes but looks like we should add them. It is also super easy to use with https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/interfaces/querycommandinput.html#indexname

I am merging this and add in new PR, given this PR just paginates list and does not take into account sorting functionality

PR #30, opened 2022-11-18
feat: deploy to custom domain up.web3.storage

olizilla:
Adds custom domain config to api definition per https://docs.sst.dev/constructs/Api#customdomain

Adds `HOSTED_ZONE` env var. the value should match a hosted zone configured in route53 that your aws account has access to.

## TO DONE

- [x] Create hosted zone for `up.web3.storage` on route53
- [x] Add NS records to web3.storage zone on cloudflare to point `up.web3.storage` to route53
- [x] Update `HOSTED_ZONE=up.web3.storage` on seed.run

License: MIT

### DIFF
--- a/.env.tpl
+++ b/.env.tpl
@@ -1,1 +1,5 @@
 # These variables are only available in your SST code.
+
+# uncomment to try out deploying the api under a custom domain.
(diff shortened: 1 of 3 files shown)


vasco-santos (approved):
LGTM

PR #34, opened 2022-11-21
fix: types for service

vasco-santos:
Adds ucanto service types. Now we have fully type coverage within ucanto service and inferred in the tests when invoking capabilities.

Closes https://github.com/web3-storage/upload-api/issues/8


api/test/service/store.test.js
```diff
@@ -51,15 +51,17 @@ test('store/add returns signed url for uploading', async (t) => {
     with: spaceDid,
     nb: { link, size: data.byteLength },
     proofs: [proof]
-    // @ts-expect-error ʅʕ•ᴥ•ʔʃ
   }).execute(connection)
 
-  t.not(storeAdd.error, true, storeAdd.message)
+  if (storeAdd.error) {
```
vasco-santos, line 56: now we need to do this, which is the most annoying part. Because if error, other properties in store* object won't be there

api/test/service/upload.test.js
```diff
@@ -46,18 +52,17 @@ test('upload/add inserts into DB mapping between data CID and car CIDs', async (
...
     with: spaceDid,
     nb: { root, shards },
     proofs: [proof]
-    // @ts-expect-error ʅʕ•ᴥ•ʔʃ
   }).execute(connection)
 
-  // @ts-expect-error error is added by ucanto if it fails
-  t.not(uploadAdd.error, true, uploadAdd.message)
+  if (uploadAdd.error) {
+    throw new Error(uploadAdd.message)
```
alanshaw, line 64: IDK might be better to keep the whole error response than pick out the message?

```suggestion
    throw new Error('invocation failed', { cause: uploadAdd })
```

alanshaw (approved):

api/test/service/upload.test.js
vasco-santos: Yeah, I will add this in all of them 👍🏼

Issue #7, opened 2022-11-14
Wire up service DID with secrets


vasco-santos:
As follow up from https://github.com/web3-storage/upload-api/pull/4 , we need to look into integrating service DID (currently hardcoded) with secrets and environments. 

See https://github.com/web3-storage/upload-api/blob/17e1fdf7b7c5b5a40f3ed207cbf75f57aeea88a4/api/authority.js

We can take inspiration from authority.js + config.js in w3up repo

olizilla:
SST supports Secret managment via SSM now, with a nicer api so i'm gonna use that
https://docs.sst.dev/config

olizilla:
the base64 encoded edsigner private+public keypair is called:
- [`PRIVATE_KEY`](https://github.com/web3-storage/w3protocol/blob/aef55e718eb8756ed95beb237780ef7125b3aace/packages/access-api/src/utils/context.js#L45) in w3access cf worker secrets
- [`prod/uploads_v2/ingestion_service_key`](https://github.com/web3-storage/w3up/blob/b124739c30d44f9ba0e5f1b2c7b721005af111aa/src/config.js#L15) in w3up aws SSM.

PR #35, opened 2022-11-21
feat: get `PRIVATE_KEY` from `SST.Config`

olizilla:
Uses SST.Config to get our base64 encoded EdSigner keypair for our service.

Called it `PRIVATE_KEY` to match what this string is called in [w3 access-api](https://github.com/web3-storage/w3protocol/blob/aef55e718eb8756ed95beb237780ef7125b3aace/packages/access-api/src/utils/context.js#L45)

`Config` magically loads properties with fallbacks from aws SSM. I've refactored things so that we don't use it in tests to keep things simpler.

Note: We can use Config in tests, but it means our tests would call out to SSM, and fetch a fallback value for any secrets we use, so i've opted for the conceptually cleaner option of only using `Config` magic at runtime, and not in tests.

Documents how to set secrets, and I've set up the prod PRIVATE_KEY and fallback for staging and PRs.

fixes: #7 

License: MIT


vasco-santos (commented):
This looks good! I think moving ucanto server creation to `service/index.js` makes a lot of sense (even without considering the tests needs here).
Please rebase to address conflicts @olizilla

vasco-santos (approved):
LGTM

PR #36, opened 2022-11-21
feat: carpark stack with bucket and eventbus

### LINKED ISSUE #13: Integrate AWS EventBridge with S3 putObject events

vasco-santos:
Carpark is one of the centre building blocks of our infrastructure. It is a CAR store and several events are triggered from operations happening to the bucket. Our HTTP API runs the UCANTO server and is an abstraction layer to DynamoDB operations (and validating given CAR is already in Carpark). With that said and aiming to keep SST Stacks tidy, we should move the bucket from the API Stack into its own stack where we can define all the resources and triggers. API Stack can access CARPARK Stack to get Bucket reference (to get permissions to perform head operation on it).

CARPARK stack includes:
- `carpark` bucket - previously on API Stack
- Event Bus responsible for handling S3 PUT Object requests and forwarding them to appropriate targets
- Event Bus targets - for now focusing on minimal needs for MVP, so it includes SQS Event to Elastic IPFS indexer and SQS Event to have CAR file backup (and side index created)

The core code logic for CARPARK operations lives inside `api/carpark` and is tested in `test/carpark`. Currently, it includes `car backup + creating side index` and `E-IPFS indexing`

Other details:
- using node16 now (`cardex` was not being installed correctly and lead me to find that)
- I started with structure where I would have most of the code in `functions/*.handler` for `carBackupAndIndex` and `notifyCarparkBus`. This would not allow me to make tests though, as importing `import { EventBridge } from 'aws-sdk'` would just fail the test
  - consequently, I moved to an organization where lambda functions are super thin, only taking care of reading from Env and getting resources clients ready to pass to the actual implementation
  - this is also more aligned with what we are doing in `service` folder and I ended up liking the restructure
- Mini refactor to put resources (S3, Dynamo) helpers inside `helpers/resources.js` instead of `utils` for better clarity

resolves #13


stacks/carpark-stack.js
```diff
@@ -0,0 +1,123 @@
...
+
+  carparkBucket.addNotifications(stack, {
+    newCarPut: {
+      function: carparkConsumer,
+      events: ['object_created_put'],
+    }
+  })
+
+  return {
+    carparkBucket
```
vasco-santos, line 121: Using https://docs.sst.dev/advanced/cross-stack-references to get CAR reference to give permissions

api/carpark/backup-and-index.js
```diff
@@ -0,0 +1,163 @@
...
+ * Copy event content into destination bucket and index it
+ *
+ * @param {object} props
+ * @param {import('../utils/parse-sqs-event').EventRecord} props.record
+ * @param {S3Client} props.destinationBucket
+ * @param {S3Client} props.originBucket
+ * @param {string} props.destinationBucketCarName
+ * @param {string} props.destinationBucketSideIndexName
+ */
+ export const carBackupAndIndex = async ({
```
vasco-santos, line 25: This function is copied from https://github.com/web3-storage/w3up/pull/252

olizilla:
Yes. The factors that would motivate splitting a stack are really
- deployment speed – as i understand it SST + seed.run diff changes per stack and can skip a stack where it hasn't changed. Many smaller stacks means fewer checks
- code organisation - we can carve up stacks however makes the most sense.

So my suggestion was mostly based on an assumed mild deploy perf improvement if we group the carpark bucket + event bus, as that part will likely not have to change much, so could be skipped over on every re-deploy... but we are not feeling any pain around deploy speed currently so I've not strong feelings. The suggestion in this doc is to isolate stateful parts of the infra that wont change often: https://sst.dev/chapters/organizing-serverless-projects.html#a-practical-approach

But! I think we can tackle this later.

seed.yml
```diff
@@ -0,0 +1,4 @@
+
+before_compile:
+  - n 16.17.1
+  - npm install
```
olizilla, line 4:
```suggestion
  - npm ci
```

stacks/carpark-stack.js
```diff
@@ -0,0 +1,125 @@
...
+  // Elastic IPFS event for indexing
+  const indexerTopicQueue = new Queue(stack, 'indexer-topic-queue', {
+    cdk: {
+      queue: sqs.Queue.fromQueueArn(
+        stack,
+        'indexer-topic',
+        'arn:aws:sqs:us-west-2:505595374361:indexer-topic'
+      ),
+    },
+  })
```
olizilla, line 34: we should check in with @francardoso93 here... for pickup the request was to have it create an SNS topic that E-IPFS can subscribe to, rather than publishing directly to the existing queue. The preference is to have a topic per system, and have E-IPFS subscribe to them all.

olizilla: we can land it like this for now, but we might need to change it later.

stacks/carpark-stack.js
```diff
@@ -0,0 +1,125 @@
...
+
+  // carpark backup and index for Freeway
+  const carBackupAndIndexHandler = new Function(
+    stack,
+    'car-backup-and-index-handler',
+    {
+      environment: {
+        BACKUP_ACCOUNT_ID: process.env.BACKUP_ACCOUNT_ID || '',
```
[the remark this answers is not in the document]
olizilla: Maybe we rename it around being `freeway*`... the other parts are named around `eipfs*`

stacks/carpark-stack.js
vasco-santos: That would be cool! Let's not block this PR and create an issue to track it.

@francardoso93 is E-IPFS already ready to support this?

vasco-santos:
@alanshaw yes definitely. I will do in new PR for that https://github.com/web3-storage/upload-api/issues/46

api/functions/car-backup-and-index.js
```diff
@@ -0,0 +1,75 @@
...
+  } = getEnv()
+
+  const record = parseSqsEvent(event)
+  if (!record) {
+    throw new Error('Invalid CAR file format')
+  }
+
+  const destinationBucket = new S3Client({
+    region: 'auto',
+    endpoint: `https://${BACKUP_ACCOUNT_ID}.r2.cloudflarestorage.com`,
```
alanshaw, line 29: Would be nice if this wasn't specific to cloudflare - can the domain or account+domain be an env var?

stacks/carpark-stack.js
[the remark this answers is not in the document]
vasco-santos: We might have other things built on CF later on, like PTP transport. So, I would say we could call it `EDGE_*` instead? it fits well with edge gateway and all CF edge things we might do like freeway.

Thoughts?

api/functions/car-backup-and-index.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  return {
+    BACKUP_ACCOUNT_ID,
+    BACKUP_ACCESS_KEY_ID,
+    BACKUP_SECRET_ACCESS_KEY,
+    BACKUP_CAR_BUCKET_NAME,
+    BACKUP_INDEX_BUCKET_NAME,
+  }
+}
```
alanshaw, line 75: I usually create a `function mustGetEnv(name: string): string` that throws if the env variable is `undefined`, the advantage being that the error message can include the expected environment variable name.

```suggestion
/**
 * Get Env validating it is set.
 */
function getEnv() {
  return {
    BACKUP_ACCOUNT_ID: mustGetEnv('BACKUP_ACCOUNT_ID'),
    BACKUP_ACCESS_KEY_ID: mustGetEnv('BACKUP_ACCESS_KEY_ID'),
    BACKUP_SECRET_ACCESS_KEY: mustGetEnv('BACKUP_SECRET_ACCESS_KEY'),
    BACKUP_CAR_BUCKET_NAME: mustGetEnv('BACKUP_CAR_BUCKET_NAME'),
    BACKUP_INDEX_BUCKET_NAME: mustGetEnv('BACKUP_INDEX_BUCKET_NAME')
  }
}

function mustGetEnv (name) {
  if (!process.env[name]) throw new Error(`Missing env var: ${name}`)
  return process.env[name]
}
```

alanshaw (commented):
Looks great, I left just a couple of minor suggestions not blocking.

I agree that we should rename the "backup" stuff to "freeway".

olizilla:
An idea that @alanshaw had, that I want to document and refine here is "what id we split things up so that we have a lamda for creating indexes and a lambda for copying things to R2", and "what if create a separate bucket for indexes in aws... the aws SATNAV bucket"

Reading up on EventBridge, we can specify rules for which targets to invoke for any event, so we can have an R2 replicator lambda that knows how to map the aws bucket source to the right R2 bucket, but otherwise just does a bucket to bucket copy.

We can create a new aws SATNAV bucket, and send all events from the CARPARK and SATNAV buckets to the Event Bridge. when an index gets written to the SATNAV bucket, that gets copied over by the R2 Replicator, which is listening to all ObjectCreated events regardless of source bucket.

We have an CAR indexer lambda and configure that to only listen to ObjectCreated events from CARPARK. It writes the index to aws SATNAV... which in turn triggers the replicator via the event bridge.

w3-event-bus drawio

api/functions/car-backup-and-index.js
vasco-santos: endpoint is an ENV var instead :)

api/functions/carpark-event.js
```diff
@@ -0,0 +1,16 @@
+import { EventBridge } from 'aws-sdk'
```
olizilla, line 1: we're using the specific v3 package eveywhere else
```suggestion
import { EventBridge } from '@aws-sdk/client-eventbridge'
```

vasco-santos:
Merging in and iterating in new PR. Also considering https://github.com/web3-storage/upload-api/pull/36#pullrequestreview-1193235568

PR #42, opened 2022-11-23
chore: restructure tests to start container per test file

vasco-santos:
Restructure tests to start containers (DynamoDB + S3) per test file.

As discussed with @olizilla , each test creates its internal resources (tableName, bucketName) while keeping using same Docker containers:
- a simple `prepareResources` function is used to create the test specific resources
  - could be done in `beforeEach`, but this might create issues running tests in parallel as same references would be used to track tableName, bucketName
- names of the resources are propagated to `ucantoServer` creation to be used within the code (previously, they were in general context)
- creating access mock server in `beforeEach` still. Main reason is tests keeping track of counting calls which would be problematic. Given it does not spin any container keeping for now

### DIFF
--- a/api/test/service/store.test.js
+++ b/api/test/service/store.test.js
@@ -1,4 +1,5 @@
 import { testStore as test } from '../helpers/context.js'
+import { customAlphabet } from 'nanoid'
 import { CreateTableCommand, GetItemCommand } from '@aws-sdk/client-dynamodb'
 import { marshall, unmarshall } from '@aws-sdk/util-dynamodb'
 import { PutObjectCommand } from '@aws-sdk/client-s3'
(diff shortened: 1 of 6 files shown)


api/test/service/store.test.js
```diff
@@ -15,21 +16,28 @@ import { createS3, createBucket, createDynamodDb, createAccessServer } from '../
  * @typedef {import('../../service/types').ListResponse<StoreListResult>} ListResponse
  */
 
-test.beforeEach(async t => {
-  const region = 'us-west-2'
-  const tableName = 'store'
+const REGION = 'us-west-2'
```
olizilla, line 19: I think we can just set a default region for the tests and not worry about. We don't need to pass it around, minio don't care about it. I would remove all references of REGION from the tests.
```suggestion
```

api/test/service/store.test.js
```diff
@@ -63,10 +64,16 @@ test.afterEach(async t => {
 })
 
 test('store/add returns signed url for uploading', async (t) => {
+  const { tableName, bucketName } = await prepareResources(t.context.dynamoClient, t.context.s3Client)
+
   const uploadService = await Signer.generate()
   const alice = await Signer.generate()
   const { proof, spaceDid } = await createSpace(alice)
-  const connection = await getClientConnection(uploadService, t.context)
+  const connection = await getClientConnection(uploadService, t.context, {
```
olizilla, line 72: feels odd to have a different fn signature for this in tests.
```suggestion
  const connection = await getClientConnection(uploadService, { ...t.context, tableName, bucketName })
```

api/test/service/store.test.js
```diff
@@ -414,13 +469,32 @@ test('store/list can be paginated with custom size', async (t) => {
   }
 })
 
+/**
+ * @param {import("@aws-sdk/client-dynamodb").DynamoDBClient} dynamoClient
+ * @param {import("@aws-sdk/client-s3").S3Client} s3Client
+ */
+async function prepareResources (dynamoClient, s3Client) {
```
olizilla, line 476: Consider moving `prepareResources`, `createBucket` to `test/helpers/resources,js` and refactoring ``createDynamo*Table` to ``createDynamoTable` that takes a table definition... and extract the 2 table definitions to a single file that can be reused between the infra code and the test code ... we can use the documentDB to get to a closer match... can be a seperate PR tho. Would be good to ensure our infra definition of the table matches our test one

olizilla (commented):
see comments inline

olizilla (approved):
🚀

PR #43, opened 2022-11-23
feat: write mapping to dudewhere on upload add

### LINKED ISSUE #39: Write to Dudewhere on upload/add
We are still missing writing to DUDEWHERE bucket on `upload/add`

vasco-santos:
Write mapping (data CID -> car CID) into Dudewhere. It includes:
- bucket abstraction layer similar to what we have for carpark
- `upload/add` updated to write entries into Dudewhere by relying on bucket abstraction layer
  - doing same retry approach as we do in CF hosted API

Questions:
- wondering if we should just use dudewhere naming instead of long `data-cid-to-car-cid-store`. We will lose readability, but maybe we want same pattern as calling `carpark` everywhere. Thoughts?

Closes #39

### DIFF
--- a/api/buckets/data-cid-to-car-cid-store.js
+++ b/api/buckets/data-cid-to-car-cid-store.js
@@ -1,0 +1,34 @@
+import { S3Client, PutObjectCommand } from '@aws-sdk/client-s3'
+
+/**
+ * Abstraction layer with Factory to perform operations on bucket storing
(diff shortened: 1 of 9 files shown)


vasco-santos:
This has the first implementation per this week goals. If we decide to move to async per UCANLOG Table events, we need to move the code to the new trigger. Anyway, it is ready to review

vasco-santos:
We are dudewhere flavour now

alanshaw (approved):

PR #48, opened 2022-11-24
feat!: update table schema

### LINKED ISSUE #25: Decide on DB schema + implement
Currently, uploaderDID tracks the resource that an upload is invoked `with`. When we are using spaces, the `with` is the space DID. so UploaderDID is the space DID which feels wrong.

see: https://github.com/web3-storage/upload-api/blob/75f0b5327d3dc2cface553bf60f9e9ebc17598c9/api/test/service/store.test.js#L85-L86

olizilla:
Update the dynamoDB table schema to the following:

```js
/** @type TableProps */
export const storeTableProps = {
  fields: {
    space: 'string',        // `did:key:space`
    link: 'string',         // `bagy...1`
    size: 'number',         // `101`
    origin: 'string',       // `bagy...0` (prev CAR CID. optional)
    issuer: 'string',       // `did:key:agent` (issuer of ucan)
    invocation: 'string',   // `baf...ucan` (CID of invcation UCAN)
    insertedAt: 'string',   // `2022-12-24T...`
  },
  // space + link must be unique to satisfy index constraint
  primaryIndex: { partitionKey: 'space', sortKey: 'link' },
}

/** @type TableProps */
export const uploadTableProps = {
  fields: {
    space: 'string',        // `did:key:space`
    sk: 'string',           // `root#shard` + space must be unique for dynamo index constraint
    root: 'string',         // `baf...x`
    shard: 'string',        // `bagy...1
    issuer: 'string',       // `did:key:agent` (issuer of ucan)
    invocation: 'string',   // `baf...ucan` (CID of invcation UCAN)
    insertedAt: 'string',   // `2022-12-24T...`
  },
  // space + sk must be unique to satisfy index constraint
  primaryIndex: { partitionKey: 'space', sortKey: 'sk' },
}
```

- `uploaderDID` becomes `space`
- `payloadCID` becomes `link` to match the capability property
- `dataCID` becomes `root` to match the terminology in the upload-client.
- `carCID` becomes `shard` to match the terminology in the upload-client.
- adds `issuer` - the DID of the agent that issued the invocation
- adds `invocation` - the CID for the invocation ucan
- use `insertedAt` as all we are really capturing here is the time at which we inserted the doc into the db.
- align property names with the `upload-client`, our capability properties and the `access-api` data model
- ordered such the partitionKey and sortKey come first (not that it matters) and our accounting metadata last
- Fixed up types 🎉

...we can add more fields. Its a document db. The only fields we have to specify here are ones referenced in an index, but more fields are given here to give the flavour.

### References

- https://www.dynamodbguide.com/key-concepts
- w3 ucan caps https://github.com/web3-storage/w3protocol/blob/aef55e718eb8756ed95beb237780ef7125b3aace/packages/access/src/capabilities/upload.js#L37
- w3-access api db schema https://github.com/web3-storage/w3protocol/blob/cc0f4f44c06a5884c6af50cf739dad6164d8e468/packages/access-api/migrations/0000_create_spaces_table.sql
- upload-client naming https://github.com/web3-storage/w3protocol/blob/aef55e718eb8756ed95beb237780ef7125b3aace/packages/upload-client/README.md#uploadadd
- dynamodb data types https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes
- current w3up schema https://github.com/web3-storage/w3up/wiki/(current)-Ingestion-Database-Schema
- previous proposed schema update https://github.com/web3-storage/w3up/wiki/Desired-Ingestion-Schema

fixes #25 

License: MIT

### DIFF
--- a/api/service/types.ts
+++ b/api/service/types.ts
@@ -46,17 +46,17 @@ export interface CarStoreBucket {
 }
 
 export interface StoreTable {
-  exists: (uploaderDID: string, payloadCID: string) => Promise<boolean>
+  exists: (space: string, car: string) => Promise<boolean>
   insert: (item: StoreItemInput) => Promise<StoreItemOutput>
-  remove: (uploaderDID: string, payloadCID: string) => Promise<void>
-  list: (uploaderDID: string, options?: ListOptions) => Promise<ListResponse<StoreListResult>>
+  remove: (space: string, car: string) => Promise<void>
+  list: (space: string, options?: ListOptions) => Promise<ListResponse<StoreListResult>>
 }
 
 export interface UploadTable {
-  exists: (uploaderDID: string, dataCID: string) => Promise<boolean>
-  insert: (uploaderDID: string, item: UploadItemInput) => Promise<UploadItemOutput[]>
-  remove: (uploaderDID: string, dataCID: string) => Promise<void>
-  list: (uploaderDID: string, options?: ListOptions) => Promise<ListResponse<UploadItemOutput>>
+  exists: (space: string, root: string) => Promise<boolean>
+  insert: (item: UploadItemInput) => Promise<UploadItemOutput[]>
+  remove: (space: string, root: string) => Promise<void>
+  list: (space: string, options?: ListOptions) => Promise<ListResponse<UploadItemOutput>>
 }
 
 export interface Signer {
@@ -64,21 +64,16 @@ export interface Signer {
 }
 
 export interface StoreItemInput {
-  uploaderDID: string,
-  link: string,
-  origin?: string,
+  space: string,
+  car: string,
   size: number,
-  proof: string,
+  origin?: string,
+  agent: string,
+  ucan: string,
 }
 
-export interface StoreItemOutput {
-  uploaderDID: string,
-  payloadCID: string,
-  applicationDID: string,
-  origin: string,
-  size: number,
-  proof: string,
-  uploadedAt: string,
+export interface StoreItemOutput extends StoreItemInput {
+  insertedAt: string,
 }
 
 export interface StoreAddResult {
@@ -89,16 +84,16 @@ export interface StoreAddResult {
   headers?: Record<string, string>
 }
 
-export type ListOptions = {
+export interface ListOptions {
   size?: number,
   cursor?: string
 }
 
 export interface StoreListResult {
-  payloadCID: string
-  origin?: string
+  car: string
   size: number
-  uploadedAt: string
+  origin?: string
+  insertedAt: string
 }
 
 export interface ListResponse<R> {
@@ -107,16 +102,20 @@ export interface ListResponse<R> {
   results: R[]
 }
 
-export interface UploadItemInput {
-  dataCID: string,
-  carCIDs: string[]
+interface UploadItemBase {
+  space: string
+  root: string,
+  agent: string,
+  ucan: string,  
 }
 
-export interface UploadItemOutput {
-  uploaderDID: string,
-  dataCID: string,
-  carCID: string,
-  uploadedAt: string,
+export interface UploadItemInput extends UploadItemBase {
+  shards: string[]
+}
+
+export interface UploadItemOutput extends UploadItemBase {
+  shard: string,
+  insertedAt: string,
 }
 
 export interface AccessClient {
--- a/api/service/types.ts
+++ b/api/service/types.ts
@@ -1,8 +1,9 @@
 import type {
   Failure,
-  Link,
   Invocation,
   ServiceMethod,
+  UCANLink,
+  DID,
 } from '@ucanto/interface'
 import type { API } from '@ucanto/server'
 
@@ -14,6 +15,9 @@ import {
   UploadRemove,
   UploadList
 } from '@web3-storage/access/capabilities/types'
+
+/** CID v0 or CID v1 */
+export interface AnyLink extends API.Link<unknown, number, number, 0 | 1> {}
 
 export interface Service {
   store: {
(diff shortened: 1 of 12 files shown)


api/test/service/upload.test.js
```diff
@@ -492,9 +492,14 @@ async function prepareResources (dynamoClient, s3Client) {
...
       ':u': { S: spaceDid },
-    },  
-    KeyConditionExpression: 'uploaderDID = :u',
-    ProjectionExpression: 'dataCID, carCID, uploadedAt'
+    },
+    // gotta sidestep dynamo reserved words!?
+    ExpressionAttributeNames: {
+      '#space': 'space',
+      '#shard': 'shard'
+    },
```
olizilla, line 500: Minor bummer. both `space` and `shard` are reserved keywords in dyanmodb land, so you have to do this aliasing step in queries that want to use them!

api/service/types.ts
```diff
@@ -46,59 +50,61 @@ export interface CarStoreBucket {
...
-  size: number,
-  proof: string,
-  uploadedAt: string,
+  space: string
+  link: string
+  size: number
+  origin?: string
+  issuer: string
+  invocation: string
+  insertedAt: string,
```
alanshaw, line 87: Timestamp?

api/tables/store.js
```diff
@@ -26,15 +26,15 @@ export function createStoreTable (region, tableName, options = {}) {
...
+    exists: async (space, link) => {
       const cmd = new GetItemCommand({
         TableName: tableName,
         Key: marshall({
-          uploaderDID,
-          payloadCID,
+          space,
+          link: link.toString()
         }),
         AttributesToGet: ['uploaderDID'],
```
alanshaw, line 39: space?

api/tables/store.js
```diff
@@ -49,22 +49,26 @@ export function createStoreTable (region, tableName, options = {}) {
...
-        origin: origin || '',
+        space,
+        link: link.toString(),
         size,
-        proof,
-        uploadedAt: new Date().toISOString(),
+        origin: origin?.toString(),
+        issuer,
+        invocation: invocation.toString(),
+        insertedAt: new Date().toISOString(),
```
alanshaw, line 66: My mind is wanting single word since everything else is - inserted?

alanshaw (changes requested):

api/test/service/store.test.js
```diff
@@ -98,11 +98,9 @@ test('store/add returns signed url for uploading', async (t) => {
 
   const item = await getItemFromStoreTable(t.context.dynamoClient, tableName, spaceDid, link)
   t.truthy(item)
-  t.is(typeof item?.uploadedAt, 'string')
-  t.is(typeof item?.proof, 'string')
```
vasco-santos, line 102: why removing invocation?

api/service/types.ts
```diff
@@ -46,59 +50,61 @@ export interface CarStoreBucket {
...
 export interface UploadTable {
-  exists: (uploaderDID: string, dataCID: string) => Promise<boolean>
-  insert: (uploaderDID: string, item: UploadItemInput) => Promise<UploadItemOutput[]>
-  remove: (uploaderDID: string, dataCID: string) => Promise<void>
-  list: (uploaderDID: string, options?: ListOptions) => Promise<ListResponse<UploadItemOutput>>
+  exists: (space: DID, root: AnyLink) => Promise<boolean>
+  insert: (item: UploadItemInput) => Promise<UploadItemOutput[]>
+  remove: (space: DID, root: AnyLink) => Promise<void>
+  list: (space: DID, options?: ListOptions) => Promise<ListResponse<UploadItemOutput>>
 }
```
vasco-santos, line 64: So, my previous approach was having tables/* functions to have the types in the schema. Being caller responsibility to send strings instead of link/did/...

I think we should go with previous way of doing things to have DB layer clean and not responsible to know what formats UCAN have. Even more as we did not go with UCAN flavour schema

Finally, it would keep input and outputs the same in database layer, which with current commit does not happen

vasco-santos (changes requested):

api/service/types.ts
olizilla: note: We've ended up with a schema that is basically the ucan flavour! 

I was thinking that this table wrapper could take in the data as we have it, as DIDs and CIDs, and do the work of preparing it for the database. That we we get stronger type assertions on the method signatures, instead of `string` all the things` which would make it easier to mix up the args and not get a warning. There is more to do there as we still loose type info on things like the responses from dynamo, but I stand by this general pattern of having the wrapper map "the data as we have it" to the "the data as the db needs it".

I recall @alanshaw asked for this change on a previous PR as well!

api/service/types.ts
```diff
@@ -108,15 +114,21 @@ export interface ListResponse<R> {
...
 
+/** Formatted for inserting to the db */ 
 export interface UploadItemOutput {
-  uploaderDID: string,
-  dataCID: string,
-  carCID: string,
-  uploadedAt: string,
+  space: string
+  root: string
+  shard: string,
```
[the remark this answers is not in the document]
olizilla: `shard` is intended here. We wrap `shards[]` in to many individual `shard` rows.

api/tables/store.js
olizilla: This is an `exists: (DID, AnyLink) => boolean` function. We don't need any of the items properties. but yes, this is a left over that should change.

api/tables/store.js
olizilla: I vote we keep it as `insertedAt` to align with the `access-api` db schema


```sql
CREATE TABLE
    IF NOT EXISTS spaces (
        did TEXT NOT NULL PRIMARY KEY,
        product TEXT NOT NULL,
        email TEXT NOT NULL,
        agent TEXT NOT NULL,
        inserted_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
        updated_at TEXT NOT NULL DEFAULT (strftime ('%Y-%m-%dT%H:%M:%fZ', 'now')),
        UNIQUE (did)
    );
```


https://github.com/web3-storage/w3protocol/blob/77b999e2f4c40f1f0b5397db6b1f8ee99232b85e/packages/access-api/migrations/0000_create_spaces_table.sql#LL8

api/service/types.ts
vasco-santos: > note: We've ended up with a schema that is basically the ucan flavour!

In a way, I think we ended up in a between domain and UCAN based on initial proposal. 

I like to keep things separated and also consider it to be the best future proof solution. If we move to a world where CF Hosted API is `upload-api` implementing UCAN server, together with a Gateway with a Dynamo behind, we will get things written to Dynamo from a HTTP gateway that would not require data marshalling to move DID/CID around, instead of strings like the Data Model. If we move to a future where DB lives in different microservice, same applies where it should only know about its data model.

I don't remember @alanshaw calling this out before, or I did not understand that. Let's get Alan's input to get a decision

api/test/service/store.test.js
olizilla: Fixed and now tests that `invocation` is a CID, and that `insertedAt` is a recent parseable Date!

api/service/types.ts
alanshaw: I was referring to something slightly different - the response data from UCAN invocations can have CIDs because it's serialized using IPLD. It's just one less thing to serialize/deserialize (*serde*) for the wire.

IMO this abstraction should do the job of serde and we should work with rich types as much as possible. If this abstraction takes strings then we push serde code into multiple places in the code base or have a utility function to do the job and use that in multiple places - why not just have the abstraction do that? When do we want to use that utility for a purpose other than to prepare input for the abstraction?

Whenever we have to do serde for wire transfer we have to validate the serialized data the other end - we can't just pump it direct into dynamo. It usually involves parsing to create the rich data structure regardless.

api/service/types.ts
alanshaw: Just for output?

alanshaw (approved):

api/service/types.ts
vasco-santos: one route handler may call multiple of this abstractions. between tables and bucket abstractions, moving this direction means all abstractions need to handle transforming things themselves rather than the caller. 

I am ok moving forward this way, just raising my concerns with the decision

api/tables/store.js
```diff
@@ -26,15 +26,15 @@ export function createStoreTable (region, tableName, options = {}) {
...
+     * @param {import('../service/types').AnyLink} link
      */
-    exists: async (uploaderDID, payloadCID) => {
+    exists: async (space, link) => {
       const cmd = new GetItemCommand({
         TableName: tableName,
         Key: marshall({
-          uploaderDID,
-          payloadCID,
+          space,
```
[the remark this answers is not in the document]
olizilla: I understand the DID type to be a string type, so I think is fine here.

api/tables/store.js
```diff
@@ -49,22 +49,26 @@ export function createStoreTable (region, tableName, options = {}) {
...
+        origin: origin?.toString(),
+        issuer,
+        invocation: invocation.toString(),
+        insertedAt: new Date().toISOString(),
       }
 
       const cmd = new PutItemCommand({
         TableName: tableName,
-        Item: marshall(item),
+        Item: marshall(item, { removeUndefinedValues: true }),
```
olizilla, line 71: This lets us send in `shard: undefined` and `marshall` will remove that property for us.

PR #49, opened 2022-11-25
feat: satnav stack

### LINKED ISSUE #46: Provision a SATNAV bucket in AWS
SO that it is possible to run freeway in AWS

image

vasco-santos:
Adds satnav stack. In addition, `carpark` is refactored to not mention `index` given it is handled in `satnav` stack now. 

Note that:
- next PR will make `replicator` its own stack
- side index is still being created inside `replicator` (writing to CF, while new Stack just writes to AWS Bucket)
  - `replicator` refactor PR will address this. Given, replicator will not replicate only CARs from carpark bucket but also side index files
  - for this, SATNAV stack needs to wire its bucket with `EventBus` to send notifications for it (until refactor `replicator` this would be a circular stack dependency)
  - Alternatively, we can have multiple `EventBus`. We currently have `carparkEventBus` which triggers events from Carpark. These events are consumed by `eipfs-queue`, `replicator` and `satnav`. If we have a satnav specific bus `satnavEventBus`, it would mean that `replicator` would track events from both EventBus. I am tempted to this solution because `carparkEventBus` should be tied with events from carpark itself. But would appreciate thoughts @olizilla 

Closes #46


olizilla:
I think the idea of `event bridge` is to have one for your your system. I vote we keep just one for now and pull the creation up into it's own stack...

**event-stack**
- event-bridge

or

**data-stack**
- carpark-bucket
- satnav-bucket
- event-bridge



i'm kinda infavour of the latter, as it nudges us towards splitting out the stateful resources into their own project per https://sst.dev/chapters/organizing-serverless-projects.html#a-practical-approach and moves us away from thinking that the lambda that writes to a bucket somehow "owns" the bucket... the data wants to be free!

PR #55, opened 2022-11-29
feat: bus stack

### LINKED ISSUE #53: EventBus Stack
Based on https://github.com/web3-storage/upload-api/issues/50

vasco-santos:
Adds bus stack with global custom EventBridge (future will also include Websocket API).

As part of this refactor to use Bus Stack, I also looked into removing the need to have a lambda function on the Event Bridge exit just to transform S3Event into a SQS Topic. A new transformation target rules was created to achieve this. Initially, I intended to keep same structure as what E-IPFS indexer receives (`${message.region}/${message.bucketName}/${message.key}`), so that we could also remove that handler. This ended up not working in any way with  [AWS Events Rules](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events-readme.html#rule) + [InputTransformerProperty](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_events.CfnRule.InputTransformerProperty.html) + [RuleTargetInput](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-events.RuleTargetInput.html). Ended up just relying on a new JSON string with the expected properties, which also is nicer because we don't need to apply regexes to remove things from the string as before. We still need to handle E-IPFS indexing with a lambda, but that interaction will also change in the near future.

This was tested with deployed resources and everything is working as expected - CAR uploaded to CARPARK triggers lambda to copy, and `*.idx` file is written into SATNAV bucket

Other notes:
- renamed `event-bridge/index.js` file to `event-bridge/source.js` to better represent what that function does. It is the source event for the event bridge (so, lambda that forwards S3 Put Event to the Event bridge)
- hooks up writes to Satnav bucket to trigger Events - previously not done because of stacks circular dependency as described in https://github.com/web3-storage/upload-api/pull/49)

Closes #53


stacks/carpark-stack.js
```diff
@@ -79,41 +82,44 @@ export function CarparkStack({ stack }) {
...
-        SQS_REPLICATOR_QUEUE_URL: replicatorQueue.queueUrl,
+  /** @type {import('@serverless-stack/resources').EventBusQueueTargetProps} */
+  const targetReplicatorQueue = {
+    type: 'queue',
+    queue: replicatorQueue,
+    cdk: {
+      target: {
+        message: awsEvents.RuleTargetInput.fromObject({
+          region: awsEvents.EventField.fromPath('$.detail.region'),
+          bucket: awsEvents.EventField.fromPath('$.detail.bucketName'),
```
olizilla, line 97: Any reason not to match the original property name here? I am wary of renaming properties across events as it makes it harder to guess what a properties name will be when coding handlers for them.
```suggestion
          bucketName: awsEvents.EventField.fromPath('$.detail.bucketName'),
```

carpark/utils/parse-sqs-event.js
```diff
@@ -18,17 +18,15 @@
...
     return
   }
-
-  const [, bucketRegion, bucketName, key] = info
+  const { key, region, bucket } = JSON.parse(body)
 
   return {
-    bucketRegion,
-    bucketName,
+    bucketRegion: region,
```
olizilla, line 28: discussion: same here... `region` becomes `bucketRegion`... can we normalise to avoid the renames?

satnav/event-bus/source.js
```diff
@@ -0,0 +1,36 @@
...
+ * @typedef {{ detail: {key?: string, region: string, bucketName: string}}} EventBridgeEvent
+ */
+
+/**
+ * @param {import('aws-lambda').S3Event} event
+ * @param {import('@aws-sdk/client-eventbridge').EventBridge} eventBridge
+ * @param {string} eventBusName
+ */
+ export async function notifyBus(event, eventBridge, eventBusName) {
+  const entries = event.Records
```
olizilla, line 13:
```suggestion
  const s3Events = event.Records
```

satnav/event-bus/source.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (entries.length > 0) {
+    const feedbackEntries = entries.map((entry) => ({
+      EventBusName: eventBusName,
+      Source: SATNAV_EVENT_BRIDGE_SOURCE_EVENT,
+      DetailType: 'satnav_index_added',
+      Detail: JSON.stringify(entry),
+    }))
+    await eventBridge.putEvents({ Entries: feedbackEntries })
+  }
```
olizilla, line 29: minor: i think of this as mapping s3 events to bus events so suggestion is to name the lists to suggest that to the reader
```suggestion
  if (s3Events.length > 0) {
    const busEvents = s3Events.map((entry) => ({
      EventBusName: eventBusName,
      Source: SATNAV_EVENT_BRIDGE_SOURCE_EVENT,
      DetailType: 'satnav_index_added',
      Detail: JSON.stringify(entry),
    }))
    await eventBridge.putEvents({ Entries: busEvents })
  }
```

satnav/test/event-bus.test.js
```diff
@@ -0,0 +1,60 @@
...
+import { notifyBus } from '../event-bus/source.js'
+
+import {
+  s3PutInvalidRecords as fixtureS3PutInvalidRecords,
+  s3PutValidRecords as fixtureS3PutValidRecords
+} from './fixtures.js'
+
+const eventBusName = 'event-bus-arn'
+
+test('notifies event bus when new satnav bucket is written', async t => {
```
olizilla, line 12:
```suggestion
test('notifies event bus when an .idx file is added to the satnav bucket', async t => {
```

satnav/test/event-bus.test.js
```diff
@@ -0,0 +1,60 @@
...
+      Records: fixtureS3PutValidRecords
+    },
+    bus,
+    eventBusName
+  )
+  
+  t.is(response.statusCode, 200)
+})
+
+test('does not notify event bus when satnav bucket is written with non idx files', async t => {
```
olizilla, line 45:
```suggestion
test('does not notify event bus when a non .idx file is added to the satnav bucket', async t => {
```

satnav/utils/parse-sqs-event.js
```diff
@@ -18,17 +18,15 @@
     )
   }
 
-  const carId = sqsEvent.Records[0].body
-  const info = carId.match(/([^/]+)\/([^/]+)\/(.+)/)
-  if (!info) {
+  const body = sqsEvent.Records[0].body
```
olizilla, line 21: do we end up with a lib module at some point? I wish we didn't have to do this boilerplate at all!

stacks/satnav-stack.js
```diff
@@ -22,7 +23,9 @@ export function SatnavStack({ stack }) {
   const stackConfig = getConfig(stack.stage)
 
   // Get carpark reference
-  const { carparkBucket, carparkEventBus } = use(CarparkStack)
+  const { carparkBucket } = use(CarparkStack)
```
olizilla, line 26: I wonder if we can limit the permissions here so that the index writer can only _read_ files from the bucket.

olizilla (approved):
Nice bus 🚌✨

None of the suggestions and comments here are blocking... just flagging things as we go.

satnav/utils/parse-sqs-event.js
vasco-santos: Yes, I think we should. I will create issue

stacks/satnav-stack.js
vasco-santos: We should look into overall permissions yes. I will create issue

Issue #18, opened 2022-11-16
Integrate UCAN_LOG_TABLE


vasco-santos:
We might need an audit log for operations against the service.

@ice-breaker-tg context:
"The schema was going to change at some point to only hold the CID instead of the ucan itself and put all the ucans in a bucket.
"

This may be important even for billing/metrics based on https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1668532790877699?thread_ts=1668529078.032189&cid=C02BZPRS9HP

PR #56, opened 2022-11-29
feat!: return shards as array from `upload/add`

### LINKED ISSUE #52: Store upload/add `shards` as a Set.
DynamoDB supports Sets!
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes.SetTypes

We can insert 1 upload doc per space+root, instead of unwrapping the shards array:

```diff
{ 
  space: DID, 
  root: CID, 
+ shards: Set<CID>
- shard: CID
- sk: `root#shard`
}
```

with the key
```diff
{ 
  partitionKey: `space`,
+ sortKey: `root` 
- sortKey: `sk` 
}
```
so we make root unique per space...

If an agent calls `upload/add` multiple times with the same root, we can use [`UpdateItem`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html#API_UpdateItem_RequestSyntax) and specify that the `shards` set should be added to the existing item, preserving our existing "it's the union of all calls to `upload/add` semantics" while also avoiding opening us up to a single invocation expanding out to a large number of ops / records. 

We can use `UpdateItem` as an _upsert_ operation by specifying all the fields as:

> UpdateItem: Edits an existing item's attributes, or adds a new item to the table if it does not already exist.

if we choose to support user metadata from an allowed list of additional `nb` properties then we can similarly treat that as a metadata map, and merge the keys where we see multiple invocations for the same root.

This change would pair nicely with a tweak to the ouput for `upload/list` where we would return the records as we store them, giving the caller a single record per root, with a shards array, to mirror the shape that they send them to us, rather than sending repeated rows per root#shard combo.

olizilla:
Change `upload/add` to return items with a `shards` array instead of expanding them out to many root#shard rows.

Stores shards as a String Set in dynamo. Multiple invocations of `upload/add` with the same root store the union of all the shards from each invocation.

Fixes up type definitions and some tidying up from #48

Removes `issuer` and `invocation` from upload/add response, we're not using them, and there is an open question of how to store them...

## Question
- Do we want to store CIDs for all the invocations that change a given upload item _in the same document_? or seperate doc in the same table, or just not? We store a single invocation CID in the store table as we dont have any updates there, so this question hasn't come up yet. The answer connects to how we imaging the event/ucan log table working too see #18 

Fixes #52

License: MIT


api/tables/index.js
```diff
@@ -1,45 +1,10 @@
...
 /** @type TableProps */
 export const storeTableProps = {
   fields: {
     space: 'string',        // `did:key:space`
     link: 'string',         // `bagy...1`
-    size: 'number',         // `101`
-    origin: 'string',       // `bagy...0` (prev CAR CID. optional)
-    issuer: 'string',       // `did:key:agent` (issuer of ucan)
-    invocation: 'string',   // `baf...ucan` (CID of invcation UCAN)
-    insertedAt: 'string',   // `2022-12-24T...`
```
vasco-santos, line 42: nit: any reason to remove? I understand it is optional, but I actually would like to have a place where we can see the schema.

api/service/types.ts
```diff
@@ -56,23 +56,23 @@ export interface DudewhereBucket {
...
+  insert: (item: StoreAddInput) => Promise<StoreAddOutput>
   remove: (space: DID, link: AnyLink) => Promise<void>
-  list: (space: DID, options?: ListOptions) => Promise<ListResponse<StoreListResult>>
+  list: (space: DID, options?: ListOptions) => Promise<ListResponse<StoreListItem>>
 }
 
 export interface UploadTable {
   exists: (space: DID, root: AnyLink) => Promise<boolean>
-  insert: (item: UploadItemInput) => Promise<UploadItemOutput[]>
+  insert: (item: UploadAddInput) => Promise<UploadAddResult>
```
vasco-santos, line 66: Should we follow store pattern and use Output for return value?

api/test/service/upload.test.js
```diff
@@ -5,17 +5,17 @@ import { ListObjectsV2Command } from '@aws-sdk/client-s3'
...
-
-import { createS3, createBucket, createAccessServer, createDynamodDb } from '../helpers/resources.js'
+import { uploadTableProps } from '../../tables/index.js'
+import { createS3, createBucket, createAccessServer, createDynamodDb, dynamoDBTableConfig } from '../helpers/resources.js'
 import { randomCAR } from '../helpers/random.js'
 import { getClientConnection, createSpace } from '../helpers/ucanto.js'
-import { dynamoDBTableConfig, uploadTableProps } from '../../tables/index.js'
+
+// https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/classes/batchwriteitemcommand.html
+export const BATCH_MAX_SAFE_LIMIT = 25
```
vasco-santos, line 14:
```suggestion
const BATCH_MAX_SAFE_LIMIT = 25
```

api/test/service/upload.test.js
```diff
@@ -153,17 +154,132 @@ test('upload/add does not fail with no shards provided', async (t) => {
...
+  t.is(dbItems.length, 1)
+  const [item] = dbItems
+  t.falsy(item.shards)
 
   // Validate data CID -> car CID mapping
   const bucketItems = await getMappingItemsForUpload(t.context.s3Client, bucketName, root.toString())
   t.is(bucketItems.length, 0)
 })
 
+test('upload/add does can add shards to an existing item with no shards', async (t) => {
```
vasco-santos, line 173:
```suggestion
test('upload/add can add shards to an existing item with no shards', async (t) => {
```

vasco-santos (commented):
This is looking good! I am super happy with new Upload structure 🤩 

Tests are failing though

vasco-santos:
> Do we want to store CIDs for all the invocations that change a given upload item in the same document? or seperate doc in the same table, or just not? We store a single invocation CID in the store table as we dont have any updates there, so this question hasn't come up yet. 

I think maybe we don't need them if we are also storing the UCAN Log Table. But I might be missing out on something. Let's create an issue to get a broader group discussion on that without blocking this PR

olizilla:
oh my. the test is failing because it takes more than a second for the function under test to write to the db and return the value, so where i give 1000ms grace for an `insertedAt` value is not long enough!

api/service/types.ts
olizilla: we've not got a consistent pattern yet. The store table abstraction doesn't have enough info to create the `StoreAddResult` as the service includes other field like status and the signed url... so we go 

`StoreAddInput -> StoreAddOutput -> StoreAddResult`

whereas in for `upload/add` we already have all the info for the service level result. I'll split those types out to match if we need it.

Issue #50, opened 2022-11-28
Rename repo? Decouple repo? this is not upload-api


vasco-santos:
![writes-pipeline](https://user-images.githubusercontent.com/7295071/204264346-cf26a7f3-8c55-431a-be26-4d88cdf035d1.jpg)

gobengo:
wait this isn't the upload-api?

vasco-santos:
We sat together and decided to move this way

# Today

- web3-storage/w3infra
    - upload-api
    - upload-db
    - carpark
    - satnav
    - eventbus
    - replicator
    - ucan-stream

# Tomorrow

- w3protocol
    - upload-client  
    - upload-api
    - upload-api-cf

- w3infra
    - http-gateway
    - upload-db
    - carpark
    - satnav
    - eventbus
    - replicator
    - ucan-stream

PR #57, opened 2022-11-30
feat: replicator stack

### LINKED ISSUE #53: EventBus Stack

### LINKED ISSUE #54: Replicator Stack
Based on https://github.com/web3-storage/upload-api/issues/50

vasco-santos:
Adds replicator stack by moving its content out of CARPARK on the direction to #50 

Implementation details:
- replicator-stack has 2 queues, one to replicate CARs and other to replicate .idx files. At first I was attempting to do both in one single queue. Ended up moving to 2 queues because I would need to have handlers to identify origin bucket and have a hardcoded mapping where I would see if it was carpark/satnav and with that select the appropriate CF destination bucket.
  - while my decision for 2 queues was based on the above, it might actually be the best approach to not have queues throttling flow on copies from independent buckets.
  - also important noticing that despite 2 queues, there is only one handler function code. The lambda code is the same, but it is injected with destination BUCKET_NAME according to the queue it is handling.

Needs:
- [x] #55 

Closes #54


olizilla:
@vasco-santos can you say more about your thought here
>  it might actually be the best approach to not have queues throttling flow on copies from independent buckets.

replicator/index.js
```diff
@@ -0,0 +1,76 @@
...
+      })
+    )
+  } catch {
+    // Not in destinationBucket, so read from origin bucket and write to destination bucket
+    const getCmd = new GetObjectCommand({
+      Bucket: record.bucketName,
+      Key: key,
+    })
+
+    // TODO: md5
```
olizilla, line 43: We should DO this TODO!

I think the aws sdk does md5 validation automatically, but this wasn't always the case, so we should consider doing our own validation as we do in the dotStorage workers.

We could require that all files are written to buckets with the `sha256` as a metadata property, so we can reuse it here regardless of key structure or which bucket it's written to.

olizilla (commented):
Some renaming left to do. Otherwise looks good!

Please can you add all the new env vars to the README... or maybe pull them out to an ENVIRONMENT.md file.

vasco-santos:
> @vasco-santos can you say more about your thought here

@olizilla 

Sharing a queue between copy CAR files and copy idx files to R2 means less parallelisation work. 

Main goals with Queues are:
- fault tolerance
- back pressure and flow control between systems - specially important to handle usage peaks

Let's consider CAR copy happens from AWS carpark to CF carpark and idx copy happens from AWS stanav to CF satnav. Focusing on the second goal in this context, they are different blocks and we do not have any benefit on using same queue.

replicator/index.js
vasco-santos: Yes, we definitely will do this before launch. Created an issue https://github.com/web3-storage/upload-api/issues/58

vasco-santos:
> Please can you add all the new env vars to the README... or maybe pull them out to an ENVIRONMENT.md file.

Will throw a new PR asap with them.

PR #59, opened 2022-12-01
feat: add sentry

### LINKED ISSUE #20: Wire up Sentry

vasco-santos:
Adds Sentry to project based on https://blog.sentry.io/2022/02/03/guest-post-how-to-monitor-lambda-functions-in-your-sst-application/

Needs:
- [x] https://github.com/web3-storage/upload-api/pull/57 + add to new lambda

Closes #20

### DIFF
--- a/satnav/functions/satnav-writer.js
+++ b/satnav/functions/satnav-writer.js
@@ -1,17 +1,21 @@
-import {
-  S3Client
-} from '@aws-sdk/client-s3'
+import { S3Client } from '@aws-sdk/client-s3'
+import * as Sentry from '@sentry/serverless'
 
 import parseSqsEvent from '../utils/parse-sqs-event.js'
 
 import { writeSatnavIndex } from '../index.js'
+
+Sentry.AWSLambda.init({
+  dsn: process.env.SENTRY_DSN,
+  tracesSampleRate: 1.0,
+})
 
 /**
  * Get EventRecord from the SQS Event triggering the handler
  *
  * @param {import('aws-lambda').SQSEvent} event
  */
-export function handler (event) {
+function writerHandler (event) {
   const {
     SATNAV_BUCKET_NAME,
   } = getEnv()
@@ -29,6 +33,8 @@ export function handler (event) {
     satnavBucketName: SATNAV_BUCKET_NAME
   })
 }
+
+export const handler = Sentry.AWSLambda.wrapHandler(writerHandler)
 
 /**
  * Get Env validating it is set.
--- a/stacks/api-stack.js
+++ b/stacks/api-stack.js
@@ -79,7 +79,7 @@ export function ApiStack({ stack, app }) {
     routes: {
       'POST /':        'functions/ucan-invocation-router.handler',
        'GET /':        'functions/get.home',
-       'GET /error': 'functions/get.error',
+       'GET /error':   'functions/get.error',
        'GET /version': 'functions/get.version'
     },
   })
--- a/stacks/api-stack.js
+++ b/stacks/api-stack.js
@@ -22,10 +22,8 @@ export function ApiStack({ stack, app }) {
   // @ts-expect-error "prod" | "dev" | "staging" only allowed for stage
   const stackConfig = getConfig(stack.stage)
 
-  // Setup Sentry when not in local
-  if (!app.local) {
-    setupSentry(stack)
-  }
+  // Setup app monitoring with Sentry
+  setupSentry(app, stack)
 
   // Get carpark reference
   const { carparkBucket } = use(CarparkStack)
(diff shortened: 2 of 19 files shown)


stacks/config.js
```diff
@@ -100,3 +101,44 @@ export function getGitInfo () {
     branch: git.branch('.')
   }
 }
+
+/**
+ * @param {import('@serverless-stack/resources').Stack} stack
+ */
+export function setupSentry (stack) {
```
vasco-santos, line 108: maybe we should check stage here instead? thoughts?

olizilla (commented):
Use `app.local` to determine if we want Sentry. There isn't really a `dev` env. When i deploy locally the scope is `olizilla`.

`app.local` tells us if the deploy is in dev mode from running `sst start`, rather than a deploy. I think it'd be fine to put that check once in the config file as that is how we do other per stage config.

From the blog post it looks like you chose different options, can you explain a little about `NODE_OPTIONS: "-r @sentry/serverless/dist/awslambda-auto",` ? what does this do? Why not use it? Does it auto init with the sentry env vars?

stacks/config.js
olizilla: yes, we're using the config file set config depending on the stage already, so let's do the `if (!app.local)` check once here rather than in every stack

vasco-santos: done ✨

vasco-santos:
It is weird that that they recommend both things in their [blogpost](https://blog.sentry.io/2022/02/03/guest-post-how-to-monitor-lambda-functions-in-your-sst-application/) and then only one is needed.

Did you setup this for pickup/linkdex? I saw `w3up` also uses both things:
- https://github.com/web3-storage/w3up/blob/main/stacks/IngestionStack.js#L25
- https://github.com/web3-storage/w3up/blob/main/src/feedback/events/eipfs.js#L30

I can definitely try without it first and report back if it works. For this, I can add a simple route handler in our API that throws an Error

olizilla:
The blogpost doesn't say to install `@sentry/serverless`... it skips straight to "now import this dep"... and i believe the dep will magically be available becuase that's what layers are used for.

- see here where the sentry layer is created and the sentry deps are baked in https://github.com/getsentry/sentry-javascript/blob/b4ce2cd47f215109961cf4e0d3914d61b071fd51/packages/serverless/scripts/buildLambdaLayer.ts#L16-L48
- https://docs.aws.amazon.com/lambda/latest/dg/nodejs-package.html

olizilla:
> I can add a simple route handler in our API that throws an Error

Yes please. This has been really helpful for testing Sentry in the web3.storage api

vasco-santos:
@olizilla great instinct!

Looks like Lambda layer is not really required. We only need to do:

```
stack.addDefaultFunctionEnv({
    SENTRY_DSN,
  })
```

within the stack, so that functions are able to get this. Otherwise, they do not send nothing to sentry.

See https://sentry.io/organizations/protocol-labs-it/issues/3783223576/?project=4504253926539264&query=is%3Aunresolved&referrer=issue-stream

vasco-santos:
Also, will keep /error to test in staging and make sure everything is also good there

stacks/api-stack.js
```diff
@@ -76,6 +79,7 @@ export function ApiStack({ stack }) {
     routes: {
       'POST /':        'functions/ucan-invocation-router.handler',
        'GET /':        'functions/get.home',
+       'GET /error': 'functions/get.error',
```
olizilla, line 82:
```suggestion
       'GET /error':   'functions/get.error',
```

satnav/functions/satnav-writer.js
```diff
@@ -1,17 +1,21 @@
...
+import { S3Client } from '@aws-sdk/client-s3'
+import * as Sentry from '@sentry/serverless'
 
 import parseSqsEvent from '../utils/parse-sqs-event.js'
 
 import { writeSatnavIndex } from '../index.js'
 
+Sentry.AWSLambda.init({
+  dsn: process.env.SENTRY_DSN,
+  tracesSampleRate: 1.0,
```
olizilla, line 10: Note that this is gonna send telemetry to sentry from **every** invocation... this opts us in to Sentry perf monitoring where it tracks the response time of every request rather than just errors. The sample rate of 1 means "send to sentry for every request, no filtering"

olizilla (approved):
Note that `SENTRY_DSN` was already copied over from w3up when we set up `upload-api` in seed.run, so that value needs to be replaced with a new one that is specifically for `upload-api`

satnav/functions/satnav-writer.js
vasco-santos: yes, I guess we can tweak or disable when launching or after we understand better our performance profile

PR #60, opened 2022-12-01
fix: env variables

vasco-santos:
Adds env variables to the docs. Meanwhile discovered two issues with env variables:
- last replicator PR start using new pattern with `REPLICATOR_ENDPOINT` instead of `REPLICATOR_ACCOUNT_ID`. It is used already in lambda invoked https://github.com/web3-storage/upload-api/blob/main/replicator/functions/replicator.js#L15
- SATNAV_BUCKET_NAME was being injected in ENV Vars, but we don't need it anymore with latest refactors. We can just pass bucketName from reference.

TODO:
- [ ] Post Merge, remove OLD ENV VARs from build

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,27 +51,27 @@ DID of the w3access service.
 
 URL of the w3access service.
 
-### `REPLICATOR_ACCESS_KEY_ID`
+### `R2_ACCESS_KEY_ID`
 
 Access key for S3 like cloud object storage to replicate content into.
 
-### `REPLICATOR_SECRET_ACCESS_KEY`
+### `R2_SECRET_ACCESS_KEY`
 
 Secret access key for S3 like cloud object storage to replicate content into.
 
-### `REPLICATOR_ENDPOINT`
+### `R2_ENDPOINT`
 
 Endpoint for S3 like cloud object storage to replicate content into.
 
-### `REPLICATOR_CAR_BUCKET_NAME`
+### `R2_CAR_BUCKET_NAME`
 
 Bucket name to replicate written CAR files.
 
-### `REPLICATOR_INDEX_BUCKET_NAME`
+### `R2_SATNAV_BUCKET_NAME`
 
 Bucket name to replicate written .idx files.
 
-### `REPLICATOR_DUDEWHERE_BUCKET_NAME`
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -51,6 +51,34 @@ DID of the w3access service.
...
+
+### `REPLICATOR_CAR_BUCKET_NAME`
+
+Bucket name to replicate written CAR files.
+
+### `REPLICATOR_INDEX_BUCKET_NAME`
+
+Bucket name to replicate written .idx files.
+
+### `REPLICATOR_DUDEWHERE_BUCKET_NAME`
```
olizilla, line 74: documenting them helps me see what they should be! I think these names can be env vars that are not tied to the replicator!

```suggestion
### `CARPARK_BUCKET_NAME`

Bucket name to replicate written CAR files.

### `SATNAV_BUCKET_NAME`

Bucket name to replicate written .idx files.

### `DUDEWHERE_BUCKET_NAME`
```

vasco-santos: IMO we might have the need in other places to have these names. Another stack can interact with SATNAV, CAR_BUCKET within AWS and need name for that.

Thoughts? I think replicator makes it clear these are from REPLICATOR context and not internal AWS ones. This was also a recommendation from Alan in previous review

olizilla: I'm imagining we'd set env vars in our SST build called `R2_CARPARK_BUCKET_NAME` and `S3_CARPARK_BUCKET_NAME` etc.... and the in the replicator we'd bind them like:

```js
Function(stack, 'carpark_replicator  { environment: {
  TARGET_BUCKET_NAME: R2_CARPARK_BUCKET_NAME
}})

Function(stack, 'satnav_replicator  { environment: {
  TARGET_BUCKET_NAME: R2_SATNAV_BUCKET_NAME
}})
```

vasco-santos: I also think R2 is better naming, specially as replicator can/will have other things. This was decided earlier, but I will revert then

vasco-santos: I renamed out level to be `R2_*` as suggested

olizilla (approved):
`R2_CAR_BUCKET_NAME` would be `R2_CARPARK_BUCKET_NAME` to match the pattern of DUDEWHERE and SATNAV

PR #65, opened 2022-12-05
fix: import capabilities from @web3-storage/capabilities 

gobengo:
(instead of @web3-storage/access/capabilities)

Motivation:
* https://github.com/web3-storage/upload-api/issues/64


api/package.json
```diff
@@ -17,6 +17,7 @@
     "@ucanto/transport": "^3.0.2",
     "@web-std/fetch": "^4.1.0",
     "@web3-storage/access": "^6.0.1",
```
vasco-santos, line 19: This is not needed anymore right?

vasco-santos (approved):
LGTM
Just needs access client dependency to be removed

PR #66, opened 2022-12-05
refactor: move s3 url signing to car-store

### LINKED ISSUE #38: Include content-length in signed headers for signed url
See https://github.com/web3-storage/w3up/pull/266

olizilla:
- Add `createUploadUrl` to `car-store.js` as creating presigned s3 urls is a job for the bucket abstraction and having a thing called `signer` that wasn't a keypair for a DID for ucanto was confusing.
- Switch to the `@aws-sdk/s3-request-presigner` for creation of signed s3 urls to simplify credential management. The existing s3 client is reused.
- Adds `content-length` to the list of headers that form part of the url signature. 

fixes #38 

### Notes

When presigning a URL for s3, some headers are "hoisted" to the URLSearchParams; they appear in the query string so you can hand the url off as a string, and the params needed for auth are baked right in there. S3 support unpacking those params from either the query _or_ from http request headers. The query params are part of the signature, so they can't be tampered with. 

You can also "hoist" and sign any other params you want... it is tempting to move `x-amz-checksum-sha256` to the query so a user wouldn't need to provide it separately. However it appears that aws does not support pulling that value out of the query, only from a request header.

~~Meanwhile, it also appears that minio does not validate the PUT payload matches the `x-amz-checksum-sha256` header at all. I include a test here that fails to show minio accepting a presigned PUT, but sending data that doesn't match the checksum.~~

It is unknown how s3 reacts to sending a content-length that doesn't match the payload size, but there are tests for it in https://github.com/web3-storage/sigv4/pull/7

### TODO:
- [x] add in content-length assertion for presigned url
- [x] check if we can remove the amz header from store/add response. ~~The checksum must be enforced via the URL, not from an additional header that the client may(not) send.~~ I think the signing of headers bakes in an assertion that those headers will be provided when using the presigned url.

License: MIT


api/buckets/car-store.js
```diff
@@ -7,28 +9,53 @@ import { S3Client, HeadObjectCommand } from '@aws-sdk/client-s3'
...
+      }) 
       try {
-        headObjectResponse = await s3client.send(
-          new HeadObjectCommand({
-            Bucket: bucketName,
-            Key: normalizedKey
-          })
-        )
-      } catch (error) {
-        console.log(`failed head command for ${normalizedKey} with: ` + error)
```
olizilla, line 28: not all errors mean the object does not exist

olizilla:
these are the cool headers i get on the signed url

```js
URLSearchParams {
  'X-Amz-Algorithm' => 'AWS4-HMAC-SHA256',
  'X-Amz-Content-Sha256' => 'UNSIGNED-PAYLOAD',
  'X-Amz-Credential' => 'minioadmin/20221205/us-west-2/s3/aws4_request',
  'X-Amz-Date' => '20221205T231949Z',
  'X-Amz-Expires' => '86400',
  'X-Amz-Signature' => '4bf836f67afea8349a5441c5d798c9b1fb07e233e5da24c309ac6e2df2a1d5d8',
  'X-Amz-SignedHeaders' => 'content-length;host',
  'x-amz-checksum-sha256' => 'go8RUmHWc07TANIf//M+HV8oQaKr93qElP0pc4qkHTA=',
  'x-id' => 'PutObject' }
  ```

**Comment by olizilla** [comment]
docs
- https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
- https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html

TODO: figure out what s3 does with content length. It appears to be using it as one of the signed headers, so an upload should fail if the client sends a different content length header than the one used to create the url, as the signature wouldn't match.

**Comment by vasco-santos** [comment]
@olizilla tsc is complaining 😅

File: api/test/service/store.test.js [hunk 2]
```diff
@@ -63,7 +63,7 @@ test.afterEach(async t => {
   t.context.access.httpServer.close()
 })
 
-test('store/add returns signed url for uploading', async (t) => {
+test.only('store/add returns signed url for uploading', async (t) => {
```
vasco-santos, line 66:
```suggestion
test('store/add returns signed url for uploading', async (t) => {
```

api/test/service/store.test.js
```diff
@@ -97,10 +98,49 @@ test('store/add returns signed url for uploading', async (t) => {
...
+  })
+  t.is(goodPut.status, 200, await goodPut.text())
+
+  const unexpectedLength = await fetch(url, {
+    method: 'PUT',
+    mode: 'cors',
+    body: new Uint8Array([11, 22, 34, 44, 55, 1]), // one byte more
+    headers: {
+      'content-length': '6',
+      'x-amz-checksum-sha256': storeAdd.headers['x-amz-checksum-sha256']
```
vasco-santos, line 130:
```suggestion
      'x-amz-checksum-sha256': storeAdd.headers?['x-amz-checksum-sha256']
```

api/test/service/store.test.js
```diff
@@ -97,10 +98,49 @@ test('store/add returns signed url for uploading', async (t) => {
...
+    }
+  })
+  // expect signature to not match
+  t.is(unexpectedLength.status, 403, await unexpectedLength.text())
+
+  // TODO: why is minio not respecting the checksum header?
```
vasco-santos, line 136: Can we crease an issue there?

vasco-santos (commented):
This looks good to me! Ci is failing because of lint, we need to confirm all the tests are running.

Using @aws-sdk/s3-request-presigner seems quite straightforward. Would love to know the reasons that we wrote sigv4 though. Is there something that we are missing?

olizilla:
@hugomrdias wrote it for using in things like cloudflare workers where we don't want to pull in a bunch of aws SDK and don't already have an S3 client set up.

api/test/service/store.test.js
olizilla: https://github.com/web3-storage/upload-api/issues/72

olizilla:
@vasco-santos fixed merge conflicts, appeased tsc, and made tests actually work given they depend on global fetch... now, as I am about to drop a PR that upgrades us to use node 18 here, I've chosen to just get it working by enabling the node >=16.15 support for global fetch with the `--experimental-fetch' flag, just for the tests. I can remove that when we are all node 18

olizilla:
Holy cow. Consider #72 closed! I just got that test passing in CI using the latest minio from docker hub. When i removed my local 4 month old image and let it pull the latest the test started passing locally too!

api/test/service/store.test.js
```diff
@@ -97,10 +98,40 @@ test('store/add returns signed url for uploading', async (t) => {
...
+    method: 'PUT',
+    mode: 'cors',
+    body: new Uint8Array([11, 22, 34, 44, 55, 1]), // one byte more
+    headers: {
+      ...storeAdd.headers,
+      'content-length': '6'
+    }
+  })
+  // expect signature to not match
+  t.is(unexpectedLength.status, 403, await unexpectedLength.text())
```
vasco-santos, line 134: Test description does not match this flow. Can we change it or decouple in 2 tests?

olizilla: strictly this *is* a "signature not match" failure as the content-length header that was used to sign the url has the value 5, and when using it we send 6, so it fails when validating the url sig.

olizilla: i've split the tests anyway

vasco-santos (approved):

PR #68, opened 2022-12-06
feat: add ucan log table

### LINKED ISSUE #18: Integrate UCAN_LOG_TABLE

vasco-santos:
Adds ucan log table.

Please read https://www.notion.so/UCAN-LOG-TABLE-0f3870fc4b404f5cbf646bf16b463365 for more details about proposal. This is an initial step for the proposed solution on having a log of all the operations within the ucan service we run in this repo.

There is a difference on the initial discussed proposal that is implemented on this PR based on our sync discussion earlier today (will update the proposal if we really go this way!). We are only storing in Dynamo Table the raw bytes together with the raw invocation CID and the insertedTs. In other words, not storing the decoded invocation encoded as JSON. There are two main reasons for this:
1. Encode invocation in JSON adds some challenges that we did not anticipate. An invocation is super flexible and upgradable further down the line. This way, it can have a lot of different instances (DIDs, CIDs, ...), and some of them in deep levels of an object (within nb). Being able to properly parse all the invocation object into strings is quite complex and error prone. Trying to use `dag-json` to encode it in JSON is not an option given undefined properties might exist, like `nb`, `nbf`, `nnc`, ... which is not possible with `dag-json` and therefore will not create a structure that we can out of the box reconstruct into its original binary representation.
2. If we take into account the reasoning behind using a DynamoDB table here is to perform transactions as fast as possible and leaving async computations to do what is needed with a more proper DB as backend for business intelligence. It actually makes sense to not spend execution cycles encoding and decoding stuff to put into a DB that is not going to be used for querying. It feels way more correct to perform the needed encodings/decodings on the post processing lambdas from dynamoDB stream before writing into other DB.

Implementation details:
- Dynamo instead of S3 - we had decided this before in the proposal, however given current implementation where we do not keep the JSON encoded invocation I thought about this again. After some reading, DynamoDB performs faster for writing small pieces of data than S3. On the other hand, S3 storage costs are cheaper. Reading a bit more about these use cases, recommendations usually are to use DynamoDB anyway because despite the higher storage costs it has, the operation costs are significantly cheaper. So, considering a high number of UCAN invocations it will be cheaper and faster to rely on DynamoDB.
- Followed same pattern by creating abstraction factory for new Table

Needs:
- [ ] decision on better naming for ucan log table

Closes #18

### DIFF
--- a/api/functions/ucan-invocation-router.js
+++ b/api/functions/ucan-invocation-router.js
@@ -7,8 +7,9 @@ import { createCarStore } from '../buckets/car-store.js'
 import { createDudewhereStore } from '../buckets/dudewhere-store.js'
 import { createStoreTable } from '../tables/store.js'
 import { createUploadTable } from '../tables/upload.js'
+import { createUcanLogTable } from '../tables/ucan-log.js'
 import { getServiceSigner } from '../config.js'
-import { createUcantoServer } from '../service/index.js'
+import { createUcantoServer, createUcanLogger } from '../service/index.js'
 
 Sentry.AWSLambda.init({
   dsn: process.env.SENTRY_DSN,
@@ -41,6 +42,7 @@ async function ucanInvocationRouter (request) {
     STORE_TABLE_NAME: storeTableName = '',
     STORE_BUCKET_NAME: storeBucketName = '',
     UPLOAD_TABLE_NAME: uploadTableName = '',
+    UCAN_LOG_TABLE_NAME: ucanLogTableName = '',
     // set for testing
     DYNAMO_DB_ENDPOINT: dbEndpoint,
     ACCESS_SERVICE_DID: accessServiceDID = '',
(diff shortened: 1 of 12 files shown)


vasco-santos:
A note on the encoding that needs to be performed to get a JSON encoded content in:

```js
import * as DAGJSON from '@ipld/dag-json'
import * as CBOR from '@ipld/dag-cbor'


const car = await CAR.codec.decode(bytes) // Needed anyway to get CAR root CID
const cbor = CBOR.decode(car.roots[0].bytes)
const jsonEncoded = DAGJSON.encode(cbor)
const jsonDecoded = DAGJSON.decode(jsonEncoded)

// write jsonDecoded to DB
```

This means two decode operations, and one encode operation which won't be cheap for execution time. So, IMHO we should go forward with what we talked in daily today and only store raw bytes in Dynamo. Post processing should handle this encoding to store on DB we want to perform queries on

vasco-santos:
This will be partially done in https://github.com/web3-storage/upload-api/pull/74 being the DB related code added as follow up

vasco-santos:
This is no longer relevant with kinesis

PR #73, opened 2022-12-08
fix: make carpark cors again

olizilla:
> I get an upload URL! ...but
> `Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://carpark-staging-0.s3.us-east-2.amazonaws.com/`

Add CORS config to carpark bucket to allow uploads via signed URLs.

License: MIT

### DIFF
--- a/stacks/carpark-stack.js
+++ b/stacks/carpark-stack.js
@@ -28,6 +28,7 @@ export function CarparkStack({ stack, app }) {
   const { eventBus } = use(BusStack)
 
   const carparkBucket = new Bucket(stack, 'car-store', {
+    cors: true,
     ...stackConfig.carparkBucketConfig,
   })
 


stacks/carpark-stack.js
```diff
@@ -28,6 +28,7 @@ export function CarparkStack({ stack, app }) {
   const { eventBus } = use(BusStack)
 
   const carparkBucket = new Bucket(stack, 'car-store', {
+    cors: true,
```
olizilla, line 31: see: https://docs.sst.dev/constructs/Bucket#cors

olizilla: `cors: true` means allow everything from everywhere (only in CORS land tho, access still requires auth)
https://github.com/serverless-stack/sst/blob/9854bbf09e82bb05dae24b76162c08a44540f67e/packages/resources/src/Bucket.ts#L643-L654

alanshaw (approved):
Thanks!

vasco-santos (approved):

PR #74, opened 2022-12-08
feat: add kinesis log stream

vasco-santos:
Adds kinesis ucan log stream to ucanto service. Once a UCAN invocation is handled by the service, it is sent to Amazon Kinesis data streams for post processing (JSON with invocation CID, invocation bytes, and decoded invocation).

Kinesis log stream has its own stack named `UcanStreamStack` which will include resources needed for post processing of ucan stream ops.  `ApiStack` depends on `UcanStreamStack` given it will use its stream, as well as its data further down the line to get content like user facing stats

Per https://www.notion.so/UCAN-LOG-0f3870fc4b404f5cbf646bf16b463365

Other notes:
- SST Kinesis guide: https://sst.dev/examples/how-to-use-kinesis-data-streams-in-your-serverless-app.html
  - we are configuring 365 days for now
- aws related deps updated to use same everywhere
- we are currently doing redundant encodings/decodings that could be avoided if we do https://github.com/web3-storage/ucanto/issues/169
- Follow up PRs will be created with consumers:
  - data aggregation lambda for user facing stats
  - dynamoDB
  - ...
  - NOTE: they can start from before as we talked https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-lambda-event-sources.KinesisEventSourceProps.html#properties
- we need to define a partition key, but we do not need to commit now to one. Only by the time we want more shards for scaling up


stacks/ucan-stream-stack.js
```diff
@@ -0,0 +1,34 @@
...
+export function UcanStreamStack({ stack, app }) {
+  stack.setDefaultFunctionProps({
+    srcPath: 'ucan-stream'
+  })
+
+  // create a kinesis stream
+  const stream = new KinesisStream(stack, 'ucan-stream', {
+    cdk: {
+      stream: {
+        retentionPeriod: Duration.days(365)
```
vasco-santos, line 20: https://docs.aws.amazon.com/cdk/api/v1/docs/aws-kinesis-readme.html#streams

vasco-santos:
@alanshaw FYI we need some changes by our discussion last night. So, removed review until I sort it out

vasco-santos:
SST is having a bad time. Need to close to wipe out cloudformation...

Issue #78, opened 2022-12-08
S3 Presigned URL `checksum-sha256` vs `content-sha256`


olizilla:
The docs for creating S3 presigned URL state:

> You don't include a payload hash in the Canonical Request, because when you create a presigned URL, you don't know the payload content because the URL is used to upload an arbitrary payload. Instead, you use a constant string UNSIGNED-PAYLOAD.

which shows that aws hasn't considered the power of putting content-addresses everywhere... In our case we do know the hash of the expected payload in advance. The existing implementations of aws URL signers assume that we MUST include `X-Amz-Content-Sha256: UNSIGNED_PAYLOAD`... as all the docs assume guide you to do that... but I think that is just because of that original assumption.  It may be possible that we can actually set `X-Amz-Content-Sha256` to be the hash of the payload we expect, instead of that magic string.

That would be rad as it is one of the headers that aws still honours if hoisted into the query string, so we could bake all the expectations into the URL, and the recipient of the presigned URL wouldn't need to provide the additional `x-amz-checksum-sha256` header that we currently use. That one cannot be hoisted to the query, and must be provided as a header to be honoured by S3.

It's just odd that our presigned URLS currently include both `x-amz-content-sha256: UNSIGNED_PAYLOAD` **and** `x-amz-checksum-sha256: ${base64pad(sha256(bytes))}`.

**a note on encoding**

- The value for `x-amz-content-sha256` must be the **base16** encoded raw digest (no multibase or multihash prefixes)
- The value for `x-amz-checksum-sha256` must be the **base64pad** encoded raw digest (no multibase or multihash prefixes)

PR #79, opened 2022-12-08
docs: add notes on s3 presigned url generation

olizilla:
I've been digging into presigned URLs for a couple of days, and wanted to capture what I know, so i can read it when i forget next week.

I've stuck to the facts here and added my conjecture in #78 

License: MIT

### DIFF
--- a/docs/s3-presigned-url.md
+++ b/docs/s3-presigned-url.md
@@ -6,7 +6,7 @@ The URL has authentication params in the query string, so anyone with the URL ca
 
 This might sound bad, but the data used to sign the URL includes object key they can write to, the hash of the payload they can send, and the length of the payload, so if anyone has those exact bytes, they are welcome to provide them!
 
-When S3 receives a request for that URL it verifies the signature sent in the `X-Amz-Signature` search param. When we construct the url we add the `x-amz-checksum-sha256` and the `content-length` headers to the `X-Amz-SignedHeaders` param, The values for those headers are included in the signature, and must be present in the request headers for the request to be accepted. If they are missing or changed, the request will fail early with a SignatureInvalid error. The http verb (`PUT`) and the URI with the desired object key are also part of the signature.
+When S3 receives a request for that URL it verifies the signature sent in the `X-Amz-Signature` search param. When we construct the URL we add the `x-amz-checksum-sha256` and the `content-length` headers to the `X-Amz-SignedHeaders` param, The values for those headers are included in the signature, and must be present in the request headers for the request to be accepted. If they are missing or changed, the request will fail early with a `SignatureInvalid` error. The HTTP verb (`PUT`) and the URI with the desired object key are also part of the signature.
 
 This means:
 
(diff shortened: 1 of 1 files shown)


docs/s3-presigned-url.md
```diff
@@ -0,0 +1,30 @@
+# S3 Presigned URLs
+
+When a user invokes `store/add`, we create a "presigned" URL allowing them to make a PUT request directly to S3 with their upload.
+
+The URL has authentication params in the query string, so anyone with the URL can use it to write to our bucket! 
+
+This might sound bad, but the data used to sign the URL includes object key they can write to, the hash of the payload they can send, and the length of the payload, so if anyone has those exact bytes, they are welcome to provide them!
+
+When S3 recieves a request for that URL it verifies the signature sent in the `X-Amz-Signature` search param. When we construct the url we add the `x-amz-checksum-sha256` and the `content-length` headers to the `X-Amz-SignedHeaders` param, The values for those headers are included in the signature, and must be present in the request headers for the request to be accepted. If they are missing or changed, the request will fail early with a SignatureInvalid error. The http verb (`PUT`) and the URI with the desired object key are also part of the signature.
```
vasco-santos, line 9:
```suggestion
When S3 receives a request for that URL it verifies the signature sent in the `X-Amz-Signature` search param. When we construct the url we add the `x-amz-checksum-sha256` and the `content-length` headers to the `X-Amz-SignedHeaders` param, The values for those headers are included in the signature, and must be present in the request headers for the request to be accepted. If they are missing or changed, the request will fail early with a SignatureInvalid error. The http verb (`PUT`) and the URI with the desired object key are also part of the signature.
```

docs/s3-presigned-url.md
```diff
@@ -0,0 +1,30 @@
...
+
+This means:
+
+- It can only PUT to the object key `/${carCID}/${carCID}.car` and the user cannot change it.
+- The `x-amz-checksum-sha256` header is included in the signature; we assert the hash of the CAR we expect and the user must provide this header and the expected value when making the request.
+  - if they send a different checksum or omit the header the PUT will be rejected early due to the request not matching the URL signature.
+  - if they send the expected checksum header but different payload bytes the upload will be rejected once it is recieved when the payload checksum is verified.
```
vasco-santos, line 16:
```suggestion
  - if they send the expected checksum header but different payload bytes the upload will be rejected once it is receives when the payload checksum is verified.
```

docs/s3-presigned-url.md
```diff
@@ -0,0 +1,30 @@
...
+- the `content-length` header is also included; the user has to tell us the size in bytes of the CAR they will send.
+  - if they make a request with a different `content-length` header it will be rejected early due to the request not matching the URL signature.
+  - (TBC) if they send the expected `content-length` header but the payload is larger aws will truncate the upload after the number of bytes asserted in the `content-length` header. For the upload to succeed, this truncated blob would also need to mathc the checksum given. In the unlikely event that the checksum and the content-length both match then the upload would be accepted, as, well, the chunk we have is exactly, byte-for-byte, what they promised to send us. Wild! More likely, this truncated upload would fail to match the checksum.
```
vasco-santos, line 19:
```suggestion
  - (TBC) if they send the expected `content-length` header but the payload is larger aws will truncate the upload after the number of bytes asserted in the `content-length` header. For the upload to succeed, this truncated blob would also need to match the checksum given. In the unlikely event that the checksum and the content-length both match then the upload would be accepted, as, well, the chunk we have is exactly, byte-for-byte, what they promised to send us. Wild! More likely, this truncated upload would fail to match the checksum.
```

vasco-santos (approved):

alanshaw (approved):

PR #83, opened 2022-12-09
feat: persist ucan invocation car to s3 and replicate to r2

vasco-santos:
Per last night call, persisting CAR files of UCAN invocations handled by the service into S3 with replication to R2.

Implementation details:
- An abstraction store for the bucket was created with same pattern as we use for Dudewhere + Carpark
- persistence happens when ucan route handler was successfully executed
- create ucan-invocation stack
  - it includes S3 UCAN Bucket and its function to send notifications of written files to EventBus, so that Replicator can kick in
  - future: will include the Kinesis stream, as well as its consumer functions
- replicator stack updated with tracking events for UCAN bucket and replicating them into R2 bucket with new bucket

TODO:
- [x] create R2 buckets
- [x] set env variables in seed.run


stacks/ucan-invocation-stack.js
```diff
@@ -0,0 +1,50 @@
...
+  use
+} from '@serverless-stack/resources'
+
+import { BusStack } from './bus-stack.js'
+import { getConfig, setupSentry } from './config.js'
+
+/**
+ * @param {import('@serverless-stack/resources').StackContext} properties
+ */
+export function UcanInvocationStack({ stack, app }) {
```
olizilla, line 13: minor: perhaps this is the ucan-archive-stack ?
```suggestion
export function UcanArchiveStack({ stack, app }) {
```

ucan-invocation/test/event-bus.test.js
```diff
@@ -0,0 +1,60 @@
...
+import { notifyBus } from '../event-bus/source.js'
+
+import {
+  s3PutInvalidRecords as fixtureS3PutInvalidRecords,
+  s3PutValidRecords as fixtureS3PutValidRecords
+} from './fixtures.js'
+
+const eventBusName = 'event-bus-arn'
+
+test('notifies event bus when an .idx file is added to the satnav bucket', async t => {
```
olizilla, line 12: copy/pasta test names

upload-api/test/service/ucan-invocation.test.js
```diff
@@ -0,0 +1,118 @@
...
+        nb
+      }],
+      proofs: [proof],
+    }
+  ])
+
+  // @ts-expect-error different type interface in AWS expected request
+  await persistUcanInvocation(request, ucanStore)
+
+  const requestCar = await CAR.codec.decode(request.body)
```
olizilla, line 98: do we need to test we can round trip the ucan by fetching it from the bucket and parsing it? there is a base64 decode step that gives me pause

upload-api/buckets/ucan-store.js
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Put UCAN invocation CAR file into bucket
+     *
+     * @param {string} carCid
+     * @param {Uint8Array} bytes
+     */
+    put: async (carCid, bytes) => {
+      const putCmd = new PutObjectCommand({
+        Bucket: bucketName,
+        Key: carCid,
```
olizilla, line 28: maybe we should match the carpark keys here and `/${carCid}/${carCid}`

olizilla (commented):
comments inline

stacks/ucan-invocation-stack.js
vasco-santos: it is not archive in reality. We do other kind of post processing on it. I like `invocation` because of being a spec term https://github.com/ucan-wg/spec#29-invocation that is what we are doing.

upload-api/test/service/ucan-invocation.test.js
vasco-santos: yeah adding. That is the aws base64 thing, so adding tests as I agree we should validate things

vasco-santos:
https://github.com/web3-storage/w3infra/pull/86#pullrequestreview-1213372176 handled in https://github.com/web3-storage/w3infra/pull/83/commits/0d55985ea1a36176f742a3bb83ac402d9dee82a4

olizilla:
The latest discussion landed on "store the invocation CAR as is and store keys for the block cids that point to the car they are in like `/blockCID/carCID`. The values can be empty like we do in DUDEWHERE. This means we store the exact invocation bytes and can also find specific proof blocks quickly if we need to.

@alanshaw @vasco-santos what do you think?

olizilla:
This also sounds like something we could add in later if storing the invocations as cars is useful anyway

upload-api/test/service/ucan-invocation.test.js
```diff
@@ -0,0 +1,138 @@
...
+
+  const cmd = new GetObjectCommand({
+    Key: `${requestCarRootCid}/${requestCarRootCid}.car`,
+    Bucket: bucketName,
+  })
+  const s3Response = await t.context.s3Client.send(cmd)
+  t.is(s3Response.$metadata.httpStatusCode, 200)
+
+  // @ts-expect-error AWS types with readable stream
+  const bytes = (await s3Response.Body.toArray())[0]
```
olizilla, line 109: What is this wild magic? Why would a toArray method stick all the bytes in the first element of an array

olizilla (approved):

upload-api/test/service/ucan-invocation.test.js
vasco-santos: https://nodejs.org/api/stream.html#readabletoarrayoptions

It is kind of the easier way of obtaining the contents of a stream. And for a test purpose seems good

PR #85, opened 2022-12-09
fix: rename repo and rescope stacks

vasco-santos:
See #50 for decisions.

Notes:
- [x] seed.run project associated with new repo path and renamed

Closes https://github.com/web3-storage/upload-api/issues/50

PR #87, opened 2022-12-12
chore: tweak clean script

vasco-santos:
Per https://github.com/web3-storage/w3infra/pull/85#discussion_r1045600886


olizilla (approved):
Note that the change to package-lock.json is unexpected here

vasco-santos:
I will remove it

Issue #90, opened 2022-12-12
upload-api ucanto server has `id.did()` corresponding to `env.DID` when present


gobengo:
Motivation
* https://github.com/web3-storage/w3protocol/issues/183

Requirements
* if `env.DID` is unset or empty, the ucanto server `id.did()` should be the `did:key` corresponding to `env.PRIVATE_KEY`

gobengo:
#status pr open here https://github.com/web3-storage/w3infra/pull/91

PR #91, opened 2022-12-12
feat: upload-api ucanto server id uses did from env.DID

gobengo:
Motivation:
* https://github.com/web3-storage/w3infra/issues/90

What
* w3infra upload-api will use env.UPLOAD_API_DID for its ucanto server id
* UPLOAD_API_DID gets set to did:web:{customDomain} on new sst stack creation, where customDomain comes from seed.run env.HOSTED_ZONE

Questions:
* I think I did the `Config.Parameter` right using sst for the upload-api `UPLOAD_API_DID` parameter? but not sure since other configs just use environment variables without explicitly using `Config.Parameter`. How do we decide when to use `Config.Parameter` and when not to?


stacks/upload-api-stack.js
```diff
@@ -28,6 +28,7 @@ export function UploadApiStack({ stack, app }) {
   const pkg = getApiPackageJson()
   const git = getGitInfo()
   const privateKey = new Config.Secret(stack, 'PRIVATE_KEY')
+  const uploadApiDid = new Config.Parameter(stack, 'UPLOAD_API_DID', { value: customDomain ? `did:web:${customDomain.domainName}` : '' })
```
Gozala, line 31: Perhaps add a comment what does this do ? I don't know much about SST and perhaps it would have been obvious if I knew. My guess is that this will either read from env variable or fallback to a thing with `customDomain` which I'm not sure what that one would be.

Does this imply we'll always have `uploadApiDid` ? Do we always want `did:web` would it make sense to stay with `did:key` if env var isn't set like you chose to do in access API ?

Gozala (commented):
Looks good to me, but I'd be more comfortable someone with more knowledge of the code here to approve this

stacks/upload-api-stack.js
vasco-santos, line 31: Can you also add to README?

gobengo: Yep, done :)

gobengo: @gozala this basically created a new parameter/environment variable 'UPLOAD_API_DID', and gives permissions to the upload-api lambdas to read it. The first time the stack and parameter are created, it'll set its initial value using `customDomain` if possible, otherwise it'll be empty string.

>  Do we always want `did:web` would it make sense to stay with `did:key` if env var isn't set like you chose to do in access API ?

If UPLOAD_API_DID gets set to empty string, I coded the rest so the ucanto server id signer would have a `did:key` as its did

gobengo: https://github.com/web3-storage/w3infra/pull/91/commits/a52ff18f1b2ec03fd80af116b958a821e0f4fa2c

vasco-santos (commented):
@gobengo your suggestion here with Config.Parameter is quite interesting. We are only using Config for a Secret at the moment. Mind raising an issue to consider moving all ENV VARs into that? Probably good to make it that way and be consistent with CF way of doing things. But would like to get feedback from the team.

In the meantime, let's use environment variables here like the others

In the meantime

upload-api/test/config.test.js
```diff
@@ -0,0 +1,48 @@
...
+     * Public key encoded as a did:key
+     */
+    did: 'did:key:z6MkqBzPG7oNu7At8fktasQuS7QR7Tj7CujaijPMAgzdmAxD',
+  },
+}
+
+test('upload-api/config getServiceSigner creates a signer using config.{UPLOAD_API_KEY,PRIVATE_KEY}', async (t) => {
+  const config = {
+    PRIVATE_KEY: testKeypair.private.multiformats,
+    UPLOAD_API_DID: 'did:web:exampe.com',
```
vasco-santos, line 24: does this still make sense here?

vasco-santos (approved):
LGTM

Just nits on tests that seem to still use config?

upload-api/test/config.test.js
gobengo: Yeah, it's still testing how `getServiceSigner` behaves with combinations of `PRIVATE_KEY` and `UPLOAD_API_DID`.

gobengo:
> Just nits on tests that seem to still use config?

those tests use an object called `config` but notably don't use SST's `Config` obj

PR #96, opened 2022-12-13
chore: use node 18

olizilla:
node 18 is supported by aws, but we are waiting on sst to adopt it

see: https://github.com/serverless-stack/sst/pull/2206

License: MIT


travis:
I think they support it now, but I think we need to upgrade to v2 - mostly done here: https://github.com/web3-storage/w3infra/pull/152

vasco-santos:
replaced by https://github.com/web3-storage/w3infra/pull/335

PR #100, opened 2022-12-13
fix: upload-api doesn't use a did:web as the serviceSigner in accessClient

gobengo:
Motivation:
* @alanshaw informed me via slack that upload-api signs invocations sent to access-api. These signatures wouldn't be verifiable if their signer did was a non-key-did.

What
* This PR makes it so the service signer for most of upload-api knows nothing about `env.UPLOAD_API_DID`.
* `env.UPLOAD_API_DID` only is used for the Principal passed to the ucanto server id (to verify incoming invocations 'aud')
* similar to this I did for w3protocol https://github.com/web3-storage/w3protocol/pull/303
* /version api response object has new property `aud`, which is the expected `aud` value of any UCAN invocations sent to the ucanto server (when env.UPLOAD_API_DID is set, it'll be that did)

### DIFF
--- a/upload-api/config.js
+++ b/upload-api/config.js
@@ -1,10 +1,3 @@
-/**
- * This file uses SSTs magic Config handler.
- * If you depend on it in a test then you need to use the `sst bind` CLI to setup the config object.
- *
- * see: https://docs.sst.dev/config
- * see: https://docs.sst.dev/advanced/testing#how-sst-bind-works
- */
 import * as ed25519 from '@ucanto/principal/ed25519'
 import * as DID from '@ipld/dag-ucan/did'
 
@@ -25,7 +18,7 @@ import * as DID from '@ipld/dag-ucan/did'
  * @param {{ UPLOAD_API_DID: string } | { PRIVATE_KEY: string }} config
  * @returns {import('@ucanto/interface').Principal}
  */
-export function getServerPrincipal(config) {
+export function getServicePrincipal(config) {
   if ('UPLOAD_API_DID' in config) {
     return DID.parse(config.UPLOAD_API_DID)
   }
--- a/upload-api/config.js
+++ b/upload-api/config.js
@@ -25,7 +25,7 @@ import * as DID from '@ipld/dag-ucan/did'
  * @param {{ UPLOAD_API_DID: string } | { PRIVATE_KEY: string }} config
  * @returns {import('@ucanto/interface').Principal}
  */
-export function configureUcantoServerId(config) {
+export function getServerPrincipal(config) {
   if ('UPLOAD_API_DID' in config) {
     return DID.parse(config.UPLOAD_API_DID)
   }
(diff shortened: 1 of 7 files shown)


upload-api/functions/ucan-invocation-router.js
```diff
@@ -59,10 +59,11 @@ async function ucanInvocationRouter (request) {
 
   const { UPLOAD_API_DID } = process.env;
   const { PRIVATE_KEY } = Config
-  const serviceSigner = getServiceSigner({ UPLOAD_API_DID, PRIVATE_KEY })
+  const servicePrincipal = getServerPrincipal({ UPLOAD_API_DID, PRIVATE_KEY })
+  const serviceSigner = getServiceSigner({ PRIVATE_KEY })
   const ucanStoreBucket = createUcanStore(AWS_REGION, ucanBucketName)
 
-  const server = await createUcantoServer(serviceSigner, {
+  const server = await createUcantoServer(servicePrincipal, {
```
gobengo, line 66: Notably, the `serviceSigner` (no env.did) is still used with `createAccessClient` below

vasco-santos (approved):

upload-api/config.js
```diff
@@ -6,20 +6,28 @@
...
+  return signer
+}
+
+/**
+ * Given a config, return a ucanto principal
+ *
+ * @param {{ UPLOAD_API_DID: string } | { PRIVATE_KEY: string }} config
+ * @returns {import('@ucanto/interface').Principal}
+ */
+export function getServerPrincipal(config) {
```
olizilla, line 28: We've got `service` and `server` and i'm not clear if there is a difference.

should this be
```suggestion
export function getServicePrincipal(config) {
```

upload-api/functions/ucan-invocation-router.js
```diff
@@ -59,10 +59,11 @@ async function ucanInvocationRouter (request) {
 
   const { UPLOAD_API_DID } = process.env;
   const { PRIVATE_KEY } = Config
-  const serviceSigner = getServiceSigner({ UPLOAD_API_DID, PRIVATE_KEY })
+  const servicePrincipal = getServerPrincipal({ UPLOAD_API_DID, PRIVATE_KEY })
```
olizilla, line 62: notable here that the return value of `getServerPrincipal` is stored as `servicePrincipal`... suggesting that the function may really be `getServicePrincipal`
```suggestion
  const servicePrincipal = getServerPrincipal({ UPLOAD_API_DID, PRIVATE_KEY })
```

olizilla (approved):
👍 nice! just a minor note on naming consistency.

olizilla:
Going to adopt this and tweak the fn name and get it merged! TO GLORIOUS PROD!

upload-api/config.js
gobengo: great question. It seems to me that 'service' gets used frequently for the object with a few methods, one per ucan capability name. But then you pass that 'service' along with some other options (like `id: Principal`) to the ucanto `Server` constructor to make a 'ucanto server' that exposes the (js obj) 'service'. That's my thinking.

upload-api/functions/ucan-invocation-router.js
gobengo: ha you're right too. I did start drifting on terminology a bit. good catch :)

PR #106, opened 2022-12-14
feat: deploy replicator stack

olizilla:
the replicator stack should be deployed! We need it to replicate things like car indexes to R2.

License: MIT


alanshaw (approved):
🙏 thank you thank you

vasco-santos (approved):
😱

PR #108, opened 2022-12-14
fix: correct repo url

olizilla:
Minor as the old repo url would redirect here, _but we must keep up with the times._

License: MIT


vasco-santos (approved):
LGTM

PR #109, opened 2022-12-15
feat: enforce max single put size

### LINKED ISSUE #97: Enforce a maximum upload size
store/add should error (for now) if size is greater than 32GiB (and possibly slightly less if we want to fit it in a deal.)

olizilla:
S3 max put size is 5GB, so we enforce that in store/add to avoid giving folks a PUT url they can't use.

see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/upload-objects.html

fixes #97

License: MIT


upload-api/test/service/store.test.js
```diff
@@ -351,6 +353,35 @@ test('store/add disallowed if invocation fails access verification', async (t) =
   t.is(service.space.info.callCount, 1)
 })
 
+test.only('store/add fails when size to large to PUT', async (t) => {
```
vasco-santos, line 356:
```suggestion
test('store/add fails when size to large to PUT', async (t) => {
```

vasco-santos (approved):
LGTM
Just needs linting fixed by suggestion :)

alanshaw (approved):

Issue #105, opened 2022-12-14
Now that we use did:web it's not always obvious which did:key's are used, we should print out both.


Gozala:

gobengo:
I'm going to implement this in a similar way as https://github.com/web3-storage/w3protocol/pull/317/files

gobengo:
it's on prod now https://up.web3.storage/version (`publicKey` prop)

PR #112, opened 2022-12-15
feat(upload-api): versionGet endpoint has publicKey prop which is always a did:key

gobengo:
* also add to homeGet

Motivation:
* https://github.com/web3-storage/w3infra/issues/105


alanshaw (approved):
LGTM

Issue #58, opened 2022-12-01
md5 / sha256 on replicator


vasco-santos:
aws sdk does md5 validation automatically, but this wasn't always the case, so we should consider doing our own validation as we do in the dotStorage workers.

We could require that all files are written to buckets with the sha256 as a metadata property, so we can reuse it here regardless of key structure or which bucket it's written to.

vasco-santos:
We have `sha256` as an attribute in our bucket entries. This could enable us to use S3's `GetObjectAttributesCommand` to get the sha256 of this. However, we can't follow same pattern as `dotStorage workers` because CF Workers have an internal API for R2 that supports sha256, while S3 API is still [not](https://developers.cloudflare.com/r2/data-access/s3-api/api/) compatible.

Moved towards trying to integrate md5, with the assumption of ETag header having md5 https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html in it and that we can use S3 Client for R2 with   ✅ Content-MD5

While trying it I got `InvalidDigest: The Content-Md5 you specified is not valid.`. In theory, we should have md5 there because we are not doing Multipart upload. While attempting this locally, I am getting `The Content-MD5 you specified is not valid.` and the `md5` I can generate with https://www.npmjs.com/package/md5 is also different value.

Reaching out to R2 team about this. Will report back once I get more updated

PR #113, opened 2022-12-16
feat: data integrity using md5

vasco-santos:
We have `sha256` as an attribute in our bucket entries. This could enable us to use S3's `GetObjectAttributesCommand` to get the sha256 of this. However, we can't follow same pattern as `dotStorage workers` because CF Workers have an internal API for R2 that supports sha256, while S3 API is still [not](https://developers.cloudflare.com/r2/data-access/s3-api/api/) compatible.

Moved towards trying to integrate md5, with the assumption of ETag header having md5 https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html in it and that we can use S3 Client for R2 with   ✅ Content-MD5

We get md5 ETag in base16 (thanks docs for not mentioning this anywhere 😘) wrapped with quotation marks and S3 client for PutObjectCommand expects the base64-encoded 128-bit MD5 digest of the message (without the headers) according to [RFC 1864](https://datatracker.ietf.org/doc/html/rfc1864)

Closes https://github.com/web3-storage/w3infra/issues/58

### DIFF
--- a/replicator/index.js
+++ b/replicator/index.js
@@ -3,6 +3,8 @@ import {
   HeadObjectCommand,
   PutObjectCommand,
 } from '@aws-sdk/client-s3'
+import { toString } from 'uint8arrays/to-string'
+import { fromString } from 'uint8arrays/from-string'
 
 /**
  * @typedef {import('@aws-sdk/client-s3').S3Client} S3Client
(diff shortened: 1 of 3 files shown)


vasco-santos:
See:

1. wrote random CAR using w3up-client to S3 https://s3.console.aws.amazon.com/s3/object/carpark-pr113-0?region=us-east-2&prefix=bagbaiera6k23fndacmzy3zxxk7ql3diylrczsomo2hchyg7ox4ucjrkeshba/bagbaiera6k23fndacmzy3zxxk7ql3diylrczsomo2hchyg7ox4ucjrkeshba.car
2. Lambda replicator ran https://us-east-2.console.aws.amazon.com/lambda/home?region=us-east-2#/functions/pr113-w3infra-ReplicatorS-ucanreplicatorhandler5EE-xIFhExB74pNT?tab=monitoring
3. made it to R2 https://dash.cloudflare.com/fffa4b4363a7e5250af8357087263b3a/r2/overview/buckets/carpark-staging-0/objects/bagbaiera6k23fndacmzy3zxxk7ql3diylrczsomo2hchyg7ox4ucjrkeshba%2Fbagbaiera6k23fndacmzy3zxxk7ql3diylrczsomo2hchyg7ox4ucjrkeshba.car

olizilla (approved):
it's always the string encoding surprise. They should make hash digests be self describing.

PR #118, opened 2022-12-22
feat: ucan stream consumer count uploads in space

vasco-santos:
Adds a consumer to UCAN stream kinesis where number of uploads is tracked per space. To achieve this tracking, we simply consume UCAN stream, filter `upload/add` operations, group these operations by spaces they affect and create a DynamoDB TransactWriteItemsCommand to update the counter of each space.

This PR adds only one metric consumer. Follow up PRs for remaining user facing metrics will follow same pattern as what we decide here.

It includes:
- DynamoDB Table to track space count of uploads
- Default config for consumers
  - still each consumer can of course be tweaked based on particular needs
  - `maxBatchingWindow: Duration.minutes(2)` - the maximum amount of time to gather records before invoking the function - this number should be a balance between the duration we are ok to have as delay and number of executed lambdas we are billed. It can go up to 5 minutes. Starting with 2 minutes seems appropriate, but maybe we can consider going a bit further later on.
  - `startingPosition: StartingPosition.TRIM_HORIZON` - Where to begin consuming the stream - we will typically want to process everything from day 0
  - `bisectBatchOnError: true` - If the function returns an error, split the batch in two and retry
  - `batchSize: 10` - batch size per lambda executed - **This is the most important property to get right!** Dynamo Transactions allow up to 100 writes per transactions. However, our protocol allows multiple capabilities to be executed within same invocation. Therefore, we need to define a cap on how many we are comfortable to execute. If we allow 10 capabilities executed per request, we can have up to 100 considering that each one would change a given space. Consequently, the "safe" number to use is batchSize of 10.
    - Taking into account that we filter out non `store/add` ops, we "merge" `store/add` operations for same Spaces and we perform `bisectBatchOnError` in case of going over limit of writes error, I consider that it makes sense to increase `batchSize` optimistically. If we decide on `batchSize: 30` I think that the probability of hitting maximum will be super low. I guess that most executed capabilities in same invocation will be from same space (they will merge in single update), and we still perform `bisectBatchOnError`, which in 2 retries would resolve for the edge-case that every single operation would be `store/add` in a different space. We throw error synchronously in case transaction items are more than we can support, so it should be quite fast to fail and retry anyway. We can monitor how this goes and tweak over time.
    - Alternatively, we can just perform multiple transactions within our Lambda for these edge cases. We would need to chunk the update items in multiple transactions as needed. I think we can consider this solution later on in case we see issues. Main reason to avoid this is that we need to take into account that no replicates happen on error handling, and therefore we need way more complex logic to track and update the kinesis consumer "cursor" as we process things.
- Dead letter queue
  - starting with simple SQS with no consumers. We can monitor issues in https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html#how-to-monitor-log-dead-letter-queues but we need to research how to automate alerts for this
  - As we evolve, we might actually consider SNS queue with a topic per consumer.
  - I suggest opening an issue to better study this theme and also discuss with E-IPFS team on what they would recommend for this.
- Integration tests to check metrics are updated running the UCAN Stream consumer

See DynamoDB Tables content:
- store table: https://us-east-2.console.aws.amazon.com/dynamodbv2/home?region=us-east-2#item-explorer?initialTagKey=&table=pr118-w3infra-store
- space-size-accumulated-store: https://us-east-2.console.aws.amazon.com/dynamodbv2/home?region=us-east-2#item-explorer?table=pr118-w3infra-space-upload-count


stacks/upload-db-stack.js
```diff
@@ -23,8 +29,14 @@ export function UploadDbStack({ stack, app }) {
    */
    const uploadTable = new Table(stack, 'upload', uploadTableProps)
 
+   /**
+    * This table tracks count of uploads per space.
+    */
+   const spaceUploadCountTable = new Table(stack, 'space-upload-count', spaceUploadCountTableProps)
```
alanshaw, line 35: Do you think we should have a `upload-metrics` table instead rather than having a table per metric we want to collect?

"upload" because the metrics relate to the `upload/*` invocations.

So then you'd have a key as "space DID + metric name" and you can fetch all metrics for a given space with one query.

alanshaw: ...and then the metric name might be `add-count` - as it's the count of the number of `upload/add` invocations.

As with the other PR, we should collect `remove-count` also.

ucan-invocation/functions/space-upload-count.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @param {import('../types').UcanInvocation[]} ucanInvocations
+ * @param {import('../types').UploadCountCtx} ctx
+ */
+export async function updateUploadCount (ucanInvocations, ctx) {
+  const invocationsWithUploadAdd = ucanInvocations.filter(
+    inv => inv.value.att.find(a => a.can === UPLOAD_ADD)
+  ).flatMap(inv => inv.value.att)
+
+  await ctx.uploadCountTable.increment(invocationsWithUploadAdd)
```
alanshaw, line 49: `upload/add` can mutate existing items - we should be sure this is a new one before incrementing the counter.

stacks/upload-db-stack.js
alanshaw: ...although thinking about it, why separate it, why not just have a generic "metrics by space" table with key as "space DID + metric name" and metric names as `upload-add-count` and `upload-remove-count`.

alanshaw (changes requested):

stacks/upload-db-stack.js
vasco-santos: @alanshaw I think your suggestion makes sense!

At this point, considering if that should leave in a `metrics-db-stack` then, instead of within upload stack. What do you think?

vasco-santos:
@alanshaw based on what we talked today, this PR is ready for review. We are tracking all `upload/add` until receipts

ucan-invocation/functions/space-metrics-upload-add-total.js
```diff
@@ -0,0 +1,52 @@
...
+  const ucanInvocations = parseKinesisEvent(event)
+
+  const {
+    TABLE_NAME: tableName = '',
+    // set for testing
+    DYNAMO_DB_ENDPOINT: dbEndpoint,
+  } = process.env
+
+  await updateUploadCount(ucanInvocations, {
+    metricsBySpaceTable: createMetricsBySpaceTable(AWS_REGION, tableName, {
```
alanshaw, line 34: I would just call this `spaceMetricsTable` for consistency with the table name.

alanshaw (approved):

Issue #95, opened 2022-12-13
How to find the space when a user PUTs a CAR


olizilla:
We will want to be able to show when a store/add is pending vs we have actually got the CAR.

The store/add dynamo table is currently partitioned by space DID, and the car CID is the sort key, so we enforce a unique car per space. This makes it cheap to find all cars for a space and terrible for finding a space by car CID.

We need some way to connect the ObjectCreated event we get from S3 back to the store/add invocation that enabled it.

olizilla:
The proposal that @Gozala re-shared is to write the `store/add` invocation to a bucket prefixed with the car cid like `/${carCid}/${invocationCid}`... so when we get an S3 ObjectCreated event we can go and list all the keys that start with the carCID to find the 1 or more invocations that got it there.

Alternatives are 
- another dynamo table keyed by carCid (same idea but state in a db)
- a redis (same idea but state in a fast store)
- encode the space or invocation CID into metadata in the query string in the presigned PUT URL.
  - the later is possible, but would need to prove that we get the metadata in the S3 event to deal with the case where the same shard is PUT from two different store/add invocations.

olizilla:
or we could add global secondary index on the the store/add table to let us find spaces by car.

```js
{ partitionKey: 'link', sortKey: 'space' }
```

PR #121, opened 2023-01-04
feat: add secondary index to store table

olizilla:
adds an index to the store table using link as the partition key as a possible fix for #95 to let us find all the records for a given CAR CID.

adds support for `gloablIndexes` to our dynamo test util, to convert the sst definition to a dynamo definition.

see: https://docs.sst.dev/constructs/Table#globalindexes
see: https://www.dynamodbguide.com/global-secondary-indexes

TODO
- [ ] add method to store so we can query by link
- [ ] add tests

License: MIT

### DIFF
--- a/upload-api/tables/index.js
+++ b/upload-api/tables/index.js
@@ -13,6 +13,10 @@ export const storeTableProps = {
   },
   // space + link must be unique to satisfy index constraint
   primaryIndex: { partitionKey: 'space', sortKey: 'link' },
+  globalIndexes: {
+    // @ts-expect-error sst got an typo in its types for TableGlobalIndexProps.partitionKey
+    linkSpace: { partitionKey: 'link', sortKey: 'space' }
+  }
 }
 
 /** @type TableProps */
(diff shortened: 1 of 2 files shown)


olizilla:
@vasco-santos this is is a WIP that shows how we could add a secondary index to the store table to let us query by CAR CID but wanted to share it to get your thoughts before I continue.

upload-api/tables/index.js
```diff
@@ -13,6 +13,10 @@ export const storeTableProps = {
   },
   // space + link must be unique to satisfy index constraint
   primaryIndex: { partitionKey: 'space', sortKey: 'link' },
+  globalIndexes: {
+    // @ts-expect-error sst got an typo in its types for TableGlobalIndexProps.partitionKey
```
vasco-santos, line 17: Interesting, github actions do not agree

vasco-santos (commented):
@olizilla yes, I think this is the way to go. 

We need to make sure this does can be added to dynamoDB table without needing to recreate the table though

travis:
I'm going to close this since it's superceded by https://github.com/web3-storage/w3infra/pull/221 and is currently using AWS IAM roles that are apparently quite scarce

PR #122, opened 2023-01-04
test: integration tests setup

vasco-santos:
This PR adds the basic setup for integration tests. It relies on https://seed.run/docs/adding-a-post-deploy-phase.html to run the integration tests once seed.run finishes its deploy. You can see output in: https://console.seed.run/dag-house/w3infra/activity/stages/pr122/builds/16/post-deploy In case post deploy [fails](https://console.seed.run/dag-house/w3infra/activity/stages/pr122/builds/15/post-deploy), it is marked as failing in github CI too 👍🏼 

After this approach is validated, we should in follow up PRs to the actual interesting integration tests. Such as, using w3up client to interact with upload API, and checking its dynamoDB, buckets, R2, etc to guarantee everything happened as expect.

By default, not running `integration tests` with normal unit tests script. Main reason for this is the way longer time needed with `sst deploy` to run. We can have this in CI only, and be able to easily replicate locally when we see errors in CI.

Alternatively, I tried to get tests to run in github action first. For this, attempted to use https://github.com/marketplace/actions/wait-for-github-deployment (as well as other similar actions), but looks like we can't get these to awake with seed.run releases (even though they in theory both use Github Check API 🤷🏼‍♂️ ). After not being able to get this to work, I found out the seed run post deploy trick. From one side, I would love to get these to run in Github Actions, but I also feel this is ok while we have deployment step also running in Seed.run. When (and if) we consider moving this to github as well, we will be able to quite easily make the integration tests also run in actions.

After merging:
- [ ] enable post deploy stage for all stages (except prod)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ dist
 
 # local env files
 .env*.local
+.test-env.json
--- a/.env.tpl
+++ b/.env.tpl
@@ -3,3 +3,6 @@
 # uncomment to try out deploying the api under a custom domain.
 # the value should match a hosted zone configured in route53 that your aws account has access to.
 # HOSTED_ZONE=up.dag.haus
+
+# uncomment to set SENTRY_DSN
+# SENTRY_DSN = ''
--- a/README.md
+++ b/README.md
@@ -42,6 +42,19 @@ Deployment is managed by [seed.run]. PR's are deployed automatically to `https:/
 
 The `main` branch is deployed to https://staging.up.web3.storage and staging builds are promoted to prod manually via the UI at https://console.seed.run
 
+## Integration tests
+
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -11,4 +11,6 @@ dist
 .vscode
 
 # local env files
-.env*.local
\ No newline at end of file
+.env
+.env*.local
+.test-env.json
```
vasco-santos, line 15: Added for local integration tests outputs from `sst deploy`

vasco-santos:
@hugomrdias we pretty much aligned on this in our call last week. Going to merge to unblock this for now, until we further down the line get release + testing in Github actions instead of relying directly in seed.run like we do.

PR #132, opened 2023-01-18
fix: eipfs indexer with environment variable url

### LINKED ISSUE #130: EIPFS_INDEXER_SQS_ARN - document good value to use locally
I wanted to do a test deploy of this to test against access-api, but when I run `npm start` I see `Error: Missing env var: EIPFS_INDEXER_SQS_ARN`

I see that env var requirements was added here https://github.com/web3-storage/w3infra/pull/129

Can we:
* make this environment variable not required when running locally?
* or
  * include a good default for this var in `/.env.tpl`
  * document in the README how to product this variable

vasco-santos:
SQS Client for E-IPFS needs to have env per environment too

Closes #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,10 @@ Data source name for Sentry application monitoring service.
 
 AWS ARN for Elastic IPFS SQS indexer used to request Elastic IPFS to index given CAR files.
 
+### `EIPFS_INDEXER_SQS_URL`
+
+AWS URL for Elastic IPFS SQS indexer used to request Elastic IPFS to index given CAR files.
+
 ### Secrets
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -111,6 +111,10 @@ Data source name for Sentry application monitoring service.
 
 AWS ARN for Elastic IPFS SQS indexer used to request Elastic IPFS to index given CAR files.
 
+### `EIPFS_INDEXER_SQS_URL`
```
alanshaw, line 114: ####!

vasco-santos: https://github.com/web3-storage/w3infra/pull/133

PR #138, opened 2023-02-07
feat: replicator event

vasco-santos:
> Part of https://github.com/web3-storage/w3filecoin/issues/12

Adds event to Event Bridge on replicated file. This replicate event is generic to not be tied with single requirements of w3filecoin (i.e. event for replicated CARs). This way, everything that is replicated will trigger an event with the file key replicated and the URL where it lives. Event consumers can filter the events they care about using filters like key suffix `.car`

Needed ENV vars were set to have knowledge of public URLs of each bucket used for replication purposes


vasco-santos:
This is outdated and now lives on https://github.com/web3-storage/w3infra/pull/199

PR #142, opened 2023-02-16
proposal: test helpers in upload and store service tests

travis:
We repeat the same pattern again and again in the upload and store service tests - creating the `issuer`, `audience`, `with`, `proofs`, and `connection`, eg:

```js
  const { tableName, bucketName } = await prepareResources(t.context.dynamoClient, t.context.s3Client)

  const uploadService = await Signer.generate()
  const alice = await Signer.generate()
  const { proof, spaceDid } = await createSpace(alice)
  const connection = await getClientConnection(uploadService, {
    ...t.context,
    tableName,
    bucketName
  })
```

and then constructing/invoking/executing capabilities using them, eg:

```js
  const uploadList = await UploadCapabilities.list.invoke({
    issuer: alice,
    audience: uploadService,
    with: spaceDid,
    proofs: [proof],
    nb: {}
  }).execute(connection)
```

we often do this last bit more than once in a test, which gets pretty verbose and, as a new person in the tests, makes it a bit hard to figure out what is signal and what is noise.

 After chatting with @vasco-santos last night I decided to throw together a quick prototype of one way we might abstract this, by adding a helper that initializes the variables and creates a couple helper functions, one for each capability, eg for the upload capability:

```js
async function uploadCapabilityHelpers(t) {
  const { tableName, bucketName } = await prepareResources(t.context.dynamoClient, t.context.s3Client)

  const uploadService = await Signer.generate()
  const alice = await Signer.generate()
  const { proof, spaceDid } = await createSpace(alice)
  const proofs = [proof]
  const connection = await getClientConnection(uploadService, {
    ...t.context,
    tableName,
    bucketName
  })

  async function add(car) {
    return UploadCapabilities.add.invoke({
      issuer: alice,
      audience: uploadService,
      with: spaceDid,
      nb: { root: car.roots[0], shards: [car.cid] },
      proofs
    }).execute(connection)
  }

  async function list(size, cursor, pre = false) {
    return UploadCapabilities.list.invoke({
      issuer: alice,
      audience: uploadService,
      with: spaceDid,
      proofs,
      nb: {
        size,
        cursor,
        pre
      }
    }).execute(connection)
  }

  return {
    tableName, bucketName, audience: uploadService, issuer: alice, with: spaceDid, proofs: [proof],
    add, list
  }
}
```

This does make tests much more readable, imho. The downside of this is that these tests don't work quite as well as usage examples of the capability code, which @vasco-santos said @olizilla had previously advocated for. I'm down to try to find some better middle ground here, so this PR is intended to be a starting point for a conversation.

@vasco-santos @olizilla what do you think?

### DIFF
--- a/upload-api/test/service/upload.test.js
+++ b/upload-api/test/service/upload.test.js
@@ -595,32 +595,59 @@ test('upload/list returns entries previously uploaded by the user', async (t) =>
   }
 })
 
-test('upload/list can be paginated with custom size', async (t) => {
+async function uploadCapabilityHelpers(t) {
   const { tableName, bucketName } = await prepareResources(t.context.dynamoClient, t.context.s3Client)
 
   const uploadService = await Signer.generate()
   const alice = await Signer.generate()
   const { proof, spaceDid } = await createSpace(alice)
+  const proofs = [proof]
   const connection = await getClientConnection(uploadService, {
     ...t.context,
     tableName,
     bucketName
   })
 
-  // invoke multiple upload/add with proof
+  async function add(car) {
+    return UploadCapabilities.add.invoke({
+      issuer: alice,
+      audience: uploadService,
+      with: spaceDid,
+      nb: { root: car.roots[0], shards: [car.cid] },
+      proofs
+    }).execute(connection)
+  }
+
+  async function list(size, cursor, pre = false) {
+    return UploadCapabilities.list.invoke({
+      issuer: alice,
+      audience: uploadService,
+      with: spaceDid,
+      proofs,
+      nb: {
+        size,
+        cursor,
+        pre
+      }
+    }).execute(connection)
+  }
+
+  return {
+    tableName, bucketName, audience: uploadService, issuer: alice, with: spaceDid, proofs: [proof],
+    add, list
+  }
+}
+
+test('upload/list can be paginated with custom size', async (t) => {
+  const { add, list } = await uploadCapabilityHelpers(t)
+
   const cars = [
     await randomCAR(128),
     await randomCAR(128)
   ]
 
   for (const car of cars) {
-    await UploadCapabilities.add.invoke({
-      issuer: alice,
-      audience: uploadService,
-      with: spaceDid,
-      nb: { root: car.roots[0], shards: [car.cid] },
-      proofs: [proof]
-    }).execute(connection)
+    await add(car)
   }
 
   // Get list with page size 1 (two pages)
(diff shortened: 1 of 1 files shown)


olizilla:
It looks a bit like an upload-client! I'd prefer to either find a way to use the actual upload-client so our tests test the things we use, or keep them as simple copy/pasta/edit friendly boilerplate.

If you want to explore custom test abstractions further, I'd nudge to split some concerns out.. that example function is dealing with buckets, tables **and** capabilities. Perhaps set up the connection outside of it and pass it in?

vasco-santos:
@travis thanks for writing this up! I totally agree on simplifying these tests and I have pitched for that when we did them.

Your solution makes definitely everything cleaner, but makes more difficult to use params from `UploadCapabilities.add`, `UploadCapabilities.list` for instance. Of course we could wrap `add` and `list` returned functions with a function to add params there. With that in mind, and going towards a more expected client usage, I was more keen to keep capabilities invocations within the test callback function. In other words, the middle ground I would like to see is the abstraction of the following:

```js
async function uploadCapabilityHelpers(t) {
  const { tableName, bucketName } = await prepareResources(t.context.dynamoClient, t.context.s3Client)

  const uploadService = await Signer.generate()
  const alice = await Signer.generate()
  const { proof, spaceDid } = await createSpace(alice)
  const proofs = [proof]
  const connection = await getClientConnection(uploadService, {
    ...t.context,
    tableName,
    bucketName
  })

  return {
    tableName, bucketName, audience: uploadService, issuer: alice, with: spaceDid, proofs: [proof],
  }
}
```

These are always the same, and I would also expect clients using this API to have these abstracted and part of an application state.

vasco-santos:
> It looks a bit like an upload-client! I'd prefer to either find a way to use the actual upload-client so our tests test the things we use, or keep them as simple copy/pasta/edit friendly boilerplate.

@olizilla not sure this is the best path forward. Client is dependent on API, introducing it as a `devDependency` for low level unit tests is not good because we need to ship things in API before shipping client. Consequently, this circular dependency will be a pain to manage and will result in not having a standard for doing tests because there will happen cases like client does not yet support X, and we need to do differently

travis:
cool - this all makes sense to me, I think the `uploadCapabilityHelpers` function @vasco-santos proposes makes sense and will make things DRYer without overabstracting - I'll take it for a spin and drop a new PR with it this week!

closing for now as we won't use this directly.

PR #144, opened 2023-02-17
fix: ensure that paging forward and backward yields the same first page

travis:
My implementation of paging has a seriously confusing bug - if you paged forward and then paged backward, the list you got after paging backward would look like the original list but have its start and end cursors reversed. It would also be in the reverse order. We knew about the reverse ordering but didn't catch its implications for cursors.

This happens because when you passed `pre`, Dynamo would evaluate and list keys in reverse order.

This all brought me around to the opinion that we should always list keys in the same order, and make sure we swap endCursor and startCursor when `pre` is passed.

This PR adds tests to ensure that paging forward once and then backward does indeed give the same list of keys, in the same order, with the same startCursor and endCursor.

The power of demos!

### DIFF
--- a/upload-api/tables/store.js
+++ b/upload-api/tables/store.js
@@ -126,19 +126,19 @@ export function createStoreTable (region, tableName, options = {}) {
       const response = await dynamoDb.send(cmd)
 
       const results = response.Items?.map(i => toStoreListResult(unmarshall(i))) ?? []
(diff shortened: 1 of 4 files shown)


alanshaw (approved):

travis:
Verified this works in the seed-deploy env, thanks for the quick review @alanshaw! Merging.

Issue #117, opened 2022-12-19
UCAN Log stream consumers


vasco-santos:
> Goal of this issue is to track all metrics in our requirements and to discuss general strategy to provide them

## Requirements

* User facing aggregations 
	* [x] Space upload/add counter - https://github.com/web3-storage/w3infra/pull/118
	* [x] Space store/add counter - https://github.com/web3-storage/w3infra/pull/160
	* [x] Space store/remove counter - https://github.com/web3-storage/w3infra/pull/160
	* [x] Space store/add size sum - https://github.com/web3-storage/w3infra/pull/164
	* [x] Space store/remove size sum - https://github.com/web3-storage/w3infra/pull/164
	* [x] Space upload/remove counter - https://github.com/web3-storage/w3infra/pull/163
* System wide metrics
	* [x] Total storage committed => `store/add size` https://github.com/web3-storage/w3infra/pull/146
	* [x] Total number of CARs stored  => `store/add count` https://github.com/web3-storage/w3infra/pull/151
	* [x] Total number of CARs removed  => `store/remove count` https://github.com/web3-storage/w3infra/pull/151
	* [x] Total number of uploads removed => `upload/remove count` https://github.com/web3-storage/w3infra/pull/157
	* [x] Total number of uploads => `upload/add count` https://github.com/web3-storage/w3infra/pull/162
	* [x] Total storage uncommited => `store/remove + receipt from system` ? https://github.com/web3-storage/w3infra/pull/158
	* [ ] BLOCKED  ~~Total number of CARs actually stored => `s3 events + receipts`~~
		* both able to track uniqueness and non uploaded
* Metrics visualization
	* [x] HTTP Endpoint  https://github.com/web3-storage/w3infra/pull/156
	* [ ] Prometheus https://github.com/web3-storage/telemetry-grafana-agent/pull/3
	* [x] Grafana dashboard

## Proposed solution

- [x] https://github.com/web3-storage/w3infra/pull/118
- [x] https://github.com/web3-storage/w3infra/pull/146

## References

* https://www.notion.so/2022-12-06-Metrics-billing-requirements-f43205faa042439ea41e4bb40c911ff8
* https://www.notion.so/Metrics-c0e154acb6e548efaf9e10e5849f830b

vasco-santos:
This is now done. New scope for other metrics will be created

PR #146, opened 2023-02-22
feat: ucan stream consumer w3 accum store/add size

### LINKED ISSUE #145: Add w3 wide metrics
> Part of https://github.com/web3-storage/w3infra/issues/117 to track user facing aggregations

Now that we have UCAN STREAM log in action with Kinesis stream, in order to easily keep track of w3 wide metrics exposing an HTTP API.

vasco-santos:
Adds a consumer to UCAN stream kinesis where total size of store/add operations is tracked in the system. To achieve this tracking, we simply consume UCAN stream, filter `store/add` operations and update a DynamoDB table row incrementing previous value in batches.

This PR adds only one metric consumer. Follow up PRs for remaining systsem metrics will follow same pattern as what we decide here.

It includes:
- DynamoDB Table w3metrics, with one row per each system metric
- Default config for consumers - more details about those in https://github.com/web3-storage/w3infra/pull/118
- Dead letter queue for failures
- Integration tests to check metrics are updated running the UCAN Stream consumer

Note that:
> [Numbers can be positive, negative, or zero. Numbers can have up to 38 digits of precision.](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes)

See DynamoDB Tables content after integration tests ran:
- store table: https://us-east-2.console.aws.amazon.com/dynamodbv2/home?region=us-east-2#item-explorer?table=pr146-w3infra-store
- w3 metrics table: https://us-east-2.console.aws.amazon.com/dynamodbv2/home?region=us-east-2#item-explorer?table=pr146-w3infra-w3-metrics

Closes #145


stacks/upload-db-stack.js
```diff
@@ -21,10 +24,16 @@ export function UploadDbStack({ stack, app }) {
    * This table maps stored CAR files (shards) to an upload root cid.
    * Used by the upload/* capabilities.
    */
-   const uploadTable = new Table(stack, 'upload', uploadTableProps)
+  const uploadTable = new Table(stack, 'upload', uploadTableProps)
+
+  /**
+   * This table tracks w3 wider metrics.
+   */
+  const w3MetricsTable = new Table(stack, 'w3-metrics', w3MetricsTableProps)
```
alanshaw, line 32: Why does the table have `w3-` prefix? The others do not...

test/upload-api.test.js
```diff
@@ -33,8 +43,12 @@ test('GET /version', async t => {
 })
 
 test('POST / client can upload a file and list it', async t => {
-  const apiEndpoint = getApiEndpoint()
-  const client = await getClient(apiEndpoint)
+  const client = await getClient(t.context.apiEndpoint)
+
+  // Get metrics before upload
+  const beforeOperationW3Metrics = await getW3Metrics(t)
+  const beforeW3AccumulatedSize = beforeOperationW3Metrics.find(row => row.name === W3_METRICS_NAMES.STORE_ADD_ACCUM_SIZE)
```
alanshaw, line 50: Please can we have a separate `test('POST / updates metrics')` for example?

ucan-invocation/functions/metrics-accumulated-size.js
```diff
@@ -0,0 +1,46 @@
...
+  })
+}
+
+/**
+ * @param {import('../types').UcanInvocation[]} ucanInvocations
+ * @param {import('../types').W3AccumulatedSizeCtx} ctx
+ */
+export async function updateAccumulatedSize (ucanInvocations, ctx) {
+  const invocationsWithStoreAdd = ucanInvocations.filter(
+    inv => inv.value.att.find(a => a.can === STORE_ADD)
```
alanshaw, line 40: Should we consider also `store/remove` into this calculation? Or, perhaps better - collect a separate metric that for `store/remove-accumulated-size` (followup PR)?

test/upload-api.test.js
```diff
@@ -137,4 +151,24 @@ test('POST / client can upload a file and list it', async t => {
   }, {
     interval: 100,
   })
+
+  // Check metrics were updated
+  beforeW3AccumulatedSize && await pWaitFor(async () => {
+    const afterOperationW3Metrics = await getW3Metrics(t)
+    const afterW3AccumulatedSize = afterOperationW3Metrics.find(row => row.name === W3_METRICS_NAMES.STORE_ADD_ACCUM_SIZE)
+
+    return afterW3AccumulatedSize?.value > beforeW3AccumulatedSize.value
```
alanshaw, line 160: Please can we test the real value after storing _multiple_ times?

ucan-invocation/constants.js
```diff
@@ -0,0 +1,7 @@
+// UCAN protocol
+export const STORE_ADD = 'store/add'
+
+// Metrics
+export const W3_METRICS_NAMES = {
+  STORE_ADD_ACCUM_SIZE: `${STORE_ADD}-accumulated-size`
```
alanshaw, line 6: Meaning is basically the same but I would have gone for `total` over `accumulated` for succinctness - just wondering if there was a different reason?

ucan-invocation/tables/index.js
```diff
@@ -0,0 +1,11 @@
+/** @typedef {import('@serverless-stack/resources').TableProps} TableProps */
+
+/** @type TableProps */
+export const w3MetricsTableProps = {
+  fields: {
+    name: 'string',        // `total-size`
```
alanshaw, line 6: ha `total-size`

ucan-invocation/tables/w3-metrics.js
```diff
@@ -0,0 +1,58 @@
...
+    endpoint: options.endpoint
+  })
+
+  return {
+    /**
+     * Increment accumulated value from new given operations.
+     *
+     * @param {Capabilities} operationsInv
+     */
+    incrementAccumulatedSize: async (operationsInv) => {
```
alanshaw, line 39: This should either be specific to store add i.e. `incrementStoreAddAccumulatedSize` or take the key as a param.

alanshaw (changes requested):

test/upload-api.test.js
vasco-santos: I wanted to do what you suggested and should have documented decision in PR description!

This is the integration tests and therefore we write real CAR files (a new one per PR/commit) that will live in S3/R2. Adding more individual tests will lead into more and more files that we will need to pay for storage + egress just for testing. 

Probably I can suggest renaming this file to `integration.test.js` and make test names more appropriate to include everything. Thoughts?

ucan-invocation/functions/metrics-accumulated-size.js
vasco-santos: yes, follow up PR with your suggestion. The goal of this PR is to get a "this is the way we should do things". Then follow up PRs with consumers for each metrics will be just follow same pattern

test/upload-api.test.js
vasco-santos: Goal here was to create a balance to get to a point where we do not have flaky tests. When this integration tests run in the PR scope is actually fine. However, they also run in staging. Making such assumption in staging can be problematic because of parallel engineer testing something in staging or when we hook up other tests in staging that write stuff there (for example, for w3infra)...

We can make a conditional more explicit. If is staging, just accept `>` and if it is PR expect real value different.

What do you think?

test/upload-api.test.js
alanshaw: Ohhh ok, yeah renaming will be fine!

stacks/upload-db-stack.js
alanshaw: Should this be `admin-metrics` instead to signify it's not user facing?

ucan-invocation/functions/metrics-accumulated-size.js
vasco-santos: Added in tracking https://github.com/web3-storage/w3infra/issues/117

ucan-invocation/constants.js
vasco-santos: changed to total everywhere

test/upload-api.test.js
vasco-santos: Did this in commit

stacks/upload-db-stack.js
```diff
@@ -21,10 +24,16 @@ export function UploadDbStack({ stack, app }) {
    * This table maps stored CAR files (shards) to an upload root cid.
    * Used by the upload/* capabilities.
    */
-   const uploadTable = new Table(stack, 'upload', uploadTableProps)
+  const uploadTable = new Table(stack, 'upload', uploadTableProps)
+
+  /**
+   * This table tracks w3 wider metrics.
+   */
+  const adminMetricsTable = new Table(stack, 'metrics', adminMetricsTableProps)
```
alanshaw, line 32:
```suggestion
  const adminMetricsTable = new Table(stack, 'admin-metrics', adminMetricsTableProps)
```

test/integration.test.js
```diff
@@ -32,9 +42,14 @@ test('GET /version', async t => {
...
-test('POST / client can upload a file and list it', async t => {
-  const apiEndpoint = getApiEndpoint()
-  const client = await getClient(apiEndpoint)
+// Integration test for all flow from uploading a file to Kinesis events consumers and replicator
+test('w3infra integration flow', async t => {
+  const client = await getClient(t.context.apiEndpoint)
+
+  // Get metrics before upload
+  const beforeOperationMetrics = await getMetrics(t)
+  const beforeAccumulatedSize = beforeOperationMetrics.find(row => row.name === METRICS_NAMES.STORE_ADD_ACCUM_SIZE)
```
alanshaw, line 51:
```suggestion
  const beforeAccumulatedSize = beforeOperationMetrics.find(row => row.name === METRICS_NAMES.STORE_ADD_TOTAL_SIZE)
```

test/integration.test.js
```diff
@@ -137,4 +154,29 @@ test('POST / client can upload a file and list it', async t => {
...
   })
+
+  // Check metrics were updated
+  beforeAccumulatedSize && await pWaitFor(async () => {
+    const afterOperationMetrics = await getMetrics(t)
+    const afterAccumulatedSize = afterOperationMetrics.find(row => row.name === METRICS_NAMES.STORE_ADD_ACCUM_SIZE)
+
+    // If staging accept more broad condition given multiple parallel tests can happen there
+    if (stage === 'staging') {
+      return afterAccumulatedSize?.value > beforeAccumulatedSize.value
```
alanshaw, line 165:
```suggestion
      return afterAccumulatedSize?.value >= beforeAccumulatedSize.value + carSize
```

test/integration.test.js
```diff
@@ -137,4 +154,29 @@ test('POST / client can upload a file and list it', async t => {
   }, {
     interval: 100,
   })
+
+  // Check metrics were updated
+  beforeAccumulatedSize && await pWaitFor(async () => {
+    const afterOperationMetrics = await getMetrics(t)
+    const afterAccumulatedSize = afterOperationMetrics.find(row => row.name === METRICS_NAMES.STORE_ADD_ACCUM_SIZE)
```
alanshaw, line 161:
```suggestion
    const afterTotalSize = afterOperationMetrics.find(row => row.name === METRICS_NAMES.STORE_ADD_ACCUM_SIZE)
```

alanshaw (changes requested):

vasco-santos:
@alanshaw addresses last comments, also renaming lambda consumer to include `store-add` to better separate from future ones. Also, I named things putting `total` in the end to be coherent with Prometheus best practises so that once we expose these metrics we can keep everything consistent

alanshaw (approved):

Issue #84, opened 2022-12-09
Future repo structuring


vasco-santos:
- w3protocol
    - upload-client  
    - upload-api
    - upload-api-cf

- w3infra
    - http-gateway
    - upload-db
    - carpark
    - satnav
    - eventbus
    - replicator
    - ucan-stream

Context: https://github.com/web3-storage/upload-api/issues/50

PR #152, opened 2023-03-07
fix: update sst to v2

vasco-santos:
Fixes CI issue from https://github.com/serverless-stack/sst/issues/2587 by updating to new sst and update aws sdk deps to be compatible. See [migration guide](https://docs.sst.dev/upgrade-guide) for details.

In addition:
- srcPath was sadly removed...
- node version updated in CI to match what we use in deployments https://github.com/web3-storage/w3infra/blob/main/seed.yml#L3

TODO:
- [ ] https://docs.sst.dev/upgrade-guide#secrets


travis:
@vasco-santos for when you're back - should we get this resolved and merged?

vasco-santos:
@travis it would be cool to land this as soon as possible. But I will need to get progress on w3filecoin side of things first

travis:
ok cool - I'm down to help with updating it and getting it merged - no rush, but maybe worth chatting about later this week!

vasco-santos:
Please remember to revert 

https://github.com/web3-storage/w3infra/pull/233#pullrequestreview-1635649566

olizilla:
We could break the problem up a little, and extract services like roundabout to their own repo, and migrate it to SST v2 while we are at it. related to this and #84

vasco-santos:
> We could break the problem up a little, and extract services like roundabout to their own repo, and migrate it to SST v2 while we are at it. related to this and https://github.com/web3-storage/w3infra/issues/84

I think we will need to sync as a team on this type one. Maybe a topic for next Eng team monthly?

We previously decided to put all the things here, and that is why Roundabout came here. For instance, I also have feeling more and more that the kinesis thing should be in its own repo. But would like us to be consistent and go with an approach, instead of keeping changing between "All in same repo" and "all in different repo".

vasco-santos:
replaced by https://github.com/web3-storage/w3infra/pull/335

PR #153, opened 2023-03-07
fix: temporary fix for access log format to allow http gateway deploy

vasco-santos:
Temporary fix for CI issue from https://github.com/serverless-stack/sst/issues/2587 as suggested in https://github.com/serverless-stack/sst/issues/2587#issuecomment-1455875437

The fix simply tweaks the `accessLog` format to not include the `cognitoIdentityId` like fix from sst https://github.com/serverless-stack/sst/pull/2602/files

In the meantime, hopefully sst team provides a fix for v1, or we need to work on landing our update https://github.com/web3-storage/w3infra/pull/152 . I started by trying to update, but after spending over 2 hours debugging random CI SegFaults and other deployment issues decided to punt on this for now. v2 is quite recent and maybe we should stay in v1 some more time given these SegFault errors are likely not related to our code...

### DIFF
--- a/stacks/upload-api-stack.js
+++ b/stacks/upload-api-stack.js
@@ -66,6 +66,9 @@ export function UploadApiStack({ stack, app }) {
        'GET /error':   'functions/get.error',
        'GET /version': 'functions/get.version'
     },
+    accessLog: {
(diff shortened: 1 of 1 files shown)


olizilla (approved):
👍 

If it doesn't take more than a couple of minutes it would be good to add a link to the sst or aws source where the default version of this log line is configured so reviewers can see the diff.

vasco-santos:
> If it doesn't take more than a couple of minutes it would be good to add a link to the sst or aws source where the default version of this log line is configured so reviewers can see the diff.

I can't find it in v1.x to add a comment in code 😅  but PR description points to PR that fixes in v2 and includes all the properties we also add here. Anyone in the code can find out why this by Git Blaming

PR #154, opened 2023-03-08
feat!: refactor upload-api

Gozala:
This PR depends on https://github.com/web3-storage/w3protocol/pull/504, which will require release and update here. With local symlink  it works as expected.


Gozala:
I have no idea why do test runner here says

```
npm ERR! Missing: @ucanto/interface@4.2.3 from lock file
npm ERR! Missing: nanoid@4.0.0 from lock file
```

or how to fix it. I'm not sure why does it even expects `@ucanto/interface@4.2.3` in the lock file 🤷‍♂️

package.json
```diff
@@ -21,14 +21,14 @@
...
-    "@ucanto/transport": "^4.2.3",
+    "@ucanto/client": "^5.1.0",
+    "@ucanto/core": "^5.1.0",
+    "@ucanto/principal": "^5.0.0",
+    "@ucanto/interface": "^6.0.0",
+    "@ucanto/transport": "^5.1.0",
     "@web-std/blob": "^3.0.4",
     "@web-std/fetch": "^4.1.0",
-    "@web3-storage/access": "^9.2.0",
-    "@web3-storage/w3up-client": "^4.1.0",
```
vasco-santos, line 31: This is needed for integration tests

Gozala: But they aren't imported anywhere, how are they used ?

gobengo: https://github.com/web3-storage/w3infra/blob/cd1db3d627c9f294d59bd22f459398115fc98b8a/test/helpers/up-client.js#L9

PR #159, opened 2023-03-14
fix: api prometheus bytes naming

vasco-santos:
Per naming conventions


vasco-santos:
Merging this in as we already discussed yesterday and other CI jobs are failing because we ran out of resources to use in AWS...

PR #160, opened 2023-03-14
feat: ucan stream consumer space store add and remove count

vasco-santos:
Part of metrics work https://github.com/web3-storage/w3infra/issues/117

Adds space counters for `store/add` and `store/remove`.

### DIFF
--- a/ucan-invocation/functions/space-metrics-store-remove-total.js
+++ b/ucan-invocation/functions/space-metrics-store-remove-total.js
@@ -1,0 +1,52 @@
+import * as Sentry from '@sentry/serverless'
+
+import { createSpaceMetricsTable } from '../tables/space-metrics.js'
+import { parseKinesisEvent } from '../utils/parse-kinesis-event.js'
+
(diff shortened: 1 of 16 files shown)


ucan-invocation/functions/space-metrics-store-remove-total.js
```diff
@@ -0,0 +1,52 @@
...
+import { createSpaceMetricsTable } from '../tables/space-metrics.js'
+import { parseKinesisEvent } from '../utils/parse-kinesis-event.js'
+
+Sentry.AWSLambda.init({
+  environment: process.env.SST_STAGE,
+  dsn: process.env.SENTRY_DSN,
+  tracesSampleRate: 1.0,
+})
+
+const STORE_REMOVE = 'store/remove'
```
alanshaw, line 12: These should come from `@web3-storage/capabilities` so that if they change we'll know here.

alanshaw (approved):

PR #166, opened 2023-03-15
fix: metrics http api handler accepting trailing slash

vasco-santos:
AWS API Gateway does not know trailing slash... and Grafana Agent puts trailing slash

See:
- https://repost.aws/questions/QUuoQUNhpDRxWufwWJKIU6Uw/can-you-invoke-api-gateway-with-just-a-hostname-and-without-a-trailing-slash
- https://forum.serverless.com/t/switching-to-http-api-and-trailing-slash-issue/12717

### DIFF
--- a/stacks/upload-api-stack.js
+++ b/stacks/upload-api-stack.js
@@ -67,6 +67,8 @@ export function UploadApiStack({ stack, app }) {
        'GET /error':   'functions/get.error',
        'GET /version': 'functions/get.version',
        'GET /metrics': 'functions/metrics.handler',
+       // AWS API Gateway does not know trailing slash... and Grafana Agent puts trailing slash
+       'GET /metrics/{proxy+}': 'functions/metrics.handler',
     },
     accessLog: {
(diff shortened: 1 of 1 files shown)


stacks/upload-api-stack.js
```diff
@@ -67,6 +67,8 @@ export function UploadApiStack({ stack, app }) {
        'GET /error':   'functions/get.error',
        'GET /version': 'functions/get.version',
        'GET /metrics': 'functions/metrics.handler',
+       // AWS API Gateway does not know trailing slash... and Grafana Agent puts trailing slash
+       'GET /metrics/{proxy+}': 'functions/metrics.handler',
```
olizilla, line 71: > A greedy path variable catches all child resources of a route. To create a greedy path variable, add + to the variable name—for example, {proxy+}. The greedy path variable must be at the end of the resource path.
– https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-routes.html

Do we need a variable here? could we just do 
```suggestion
       'GET /metrics/': 'functions/metrics.handler',
```

vasco-santos: yes, it fails to deploy with error of invalid route name ending in `/`. also tried `/+` which also fails...

vasco-santos:
@olizilla merging this in to unblock grafana agent. Can do a follow up PR if there is any other feedback :)

PR #167, opened 2023-03-16
feat: github action to fetch metrics per space

### LINKED ISSUE #116: Add user facing aggregations
> Part of https://github.com/web3-storage/w3infra/issues/117 to track user facing aggregations

Now that we have UCAN STREAM log in action with Kinesis stream, in order to easily keep track of user facing aggregations.

vasco-santos:
Adds github action to query space metrics for given space

```
npm run fetch-metrics-for-space

> w3infra@0.0.0 fetch-metrics-for-space
> npm run fetch-metrics-for-space -w tools


> tools@1.0.0 fetch-metrics-for-space
> node cli.js fetch-metrics-for-space

Metrics found for provided space DID: 1
upload/add-total: 7
```

TODO
- [x] Add AWS secrets

Closes #116

### DIFF
--- a/tools/fetch-metrics-for-space.js
+++ b/tools/fetch-metrics-for-space.js
@@ -1,0 +1,102 @@
+import {
+  DynamoDBClient,
+  QueryCommand
+} from '@aws-sdk/client-dynamodb'
+import { unmarshall } from '@aws-sdk/util-dynamodb'
+
+export async function fetchMetricsForSpaceCmd () {
+  const {
+    ENV,
+    SPACE_DID,
+    TABLE_NAME,
+  } = getEnv()
+
+  const { client, tableName } = getDynamoDb(
+    TABLE_NAME,
+    ENV,
+    getRegion(ENV)
+  )
+
+  const rows = await getAllTableRows(client, tableName, SPACE_DID)
+  console.log(`Metrics found for provided space DID: ${rows.length}`)
+  for (const row of rows) {
+    console.log(`${row.name}: ${row.value}`)
+  }
+}
+
+/**
+ * @param {import('@aws-sdk/client-dynamodb').DynamoDBClient} dynamo
+ * @param {string} tableName
+ * @param {string} space
+ * @param {object} [options]
+ * @param {number} [options.limit]
+ */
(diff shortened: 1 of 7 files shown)


tools/fetch-metrics-for-space.js
```diff
@@ -0,0 +1,102 @@
...
+ * 
+ * @param {string} name 
+ * @returns {string}
+ */
+function mustGetEnv (name) {
+  if (!process.env[name]) {
+    throw new Error(`Missing env var: ${name}`)
+  }
+
+  // @ts-expect-error there will always be a string there, but typescript does not believe
```
alanshaw, line 71: I think if you assign to a variable and then do the check it'll not complain...

tools/fetch-metrics-for-space.js
```diff
@@ -0,0 +1,102 @@
...
+
+  return 'us-west-2'
+}
+
+/**
+ * @param {string} tableName
+ * @param {string} env
+ * @param {string} region
+ */
+function getDynamoDb (tableName, env, region) {
```
alanshaw, line 91: How does it authenticate?

vasco-santos: Do you mean authenticate to run the action? This is only for admin usage at the moment. We will integrate this with API later on (with `space/info`, or other API) where users will be able to interact with the system. Github triggers can only be run by collaborators with run permissions in repo.

Or do you mean dynamoDB? Yeah, I guess we need to put AWS credentials in... In my env I have them configured, and seed.run (where integration tests run with similar setup) also should have them set. I will add them

alanshaw: yes the latter!

vasco-santos:
Now node18 is failing to download 😡

vasco-santos:
@alanshaw looks like node installer is back :)

tools/fetch-metrics-for-space.js
alanshaw: No like:

```js
function mustGetEnv (name) {
  const value = process.env[name]
  if (!value) throw new Error(`Missing env var: ${name}`)
  return value
}
```

alanshaw (approved):

PR #171, opened 2023-03-20
feat: ucan invocation post handler

vasco-santos:
Adds handler for reporting UCAN invocations from other services.

This includes a `POST /ucan` route in `up.web3.storage`  protected by a HTTP Authentication token.

Per @Gozala 's feedback, also rearranges handling of invocations to support multiple invocations within a CAR file

TODO:
- [x] Add HTTP Authentication token


upload-api/ucan-invocation.js
```diff
@@ -57,6 +55,80 @@ export async function parseUcanInvocationRequest(request) {
...
+ * @param {import('aws-lambda').APIGatewayProxyEventV2} request
+ * @param {UcanInvocationCtx} ctx
+ */
+export async function processUcanInvocation(request, ctx) {
+  const token = getTokenFromRequest(request)
+  if (token !== ctx.basicAuth) {
+    throw new NoValidTokenError('invalid Authorization credentials provided')
+  }
+
+  const ucanInvocation = await parseUcanInvocationRequest(request)
```
Gozala, line 97: It looks like `parseUcanInvocationRequest` has an invalid assumption that CAR will contain a single invocation, while in practice that seems to be the case (mostly because how clients are written) all of the ucanto was written such that you could have bunch of the invocations send in one request (and I hope clients will start utilizing it sooner than later).

It would be very unfortunate if we bake single invocation per request assumption in this layer because we'll make it really hard to support multiple in the future. Furthermore while our clients do not send multiple invocations in one request there is no guarantee that third party clients won't so in which case we'll not persist them correctly.

Gozala: I think relevant code paths need to be updated to methods like `processUcanInvocations` and then deal with multiples in many cases concurrently.

upload-api/ucan-invocation.js
```diff
@@ -57,6 +55,80 @@ export async function parseUcanInvocationRequest(request) {
...
+  )
+}
+
+/**
+ * @param {import('aws-lambda').APIGatewayProxyEventV2} request
+ */
+function getTokenFromRequest (request) {
+  const authHeader = request.headers.Authorization || ''
+  if (!authHeader) {
+    throw new NoTokenError('no Authorization header provided')
```
Gozala, line 114: I wonder if introducing authorization token is even worth it. We allow anyone to send us requests anyway and we only reject requests if UCAN audience isn't our service (we may even lift that in the future).

In other words we'll be forwarding CAR requests here anyway so it is unclear to me what kind of abuse would auth header here prevent or why would any actor choose to send them this way as opposed to main API endpoint which would forward them here anyhow.

Only abuse I can think of is populating CAR with some garbage, but probably auth token is not going to help with that anyway. Maybe better guard here would be to limit CAR size and check that audience is our service instead ?

To be clear I'm not opposing having auth header, it's just in this instance I fail to recognize a merit.

Gozala (changes requested):
It looks good and pretty straightforward to use! My only concern is the baked assumption that there will be one invocation per CAR which isn't valid and I would prefer not to reinforce it.

upload-api/ucan-invocation.js
vasco-santos: Thanks for raising this @Gozala 

I just addressed all code paths to take this into account. Ended up renaming some types and functions to better reflect distinction between Invocation and Car with multiple Invocations

upload-api/ucan-invocation.js
vasco-santos: > We allow anyone to send us requests anyway and we only reject requests if UCAN audience isn't our service (we may even lift that in the future).

This is true. But if we receive such requests we have guarantee that they were really processed. Having this open has the opposite effect where attacker could send us Invocations that were not handled. For instance, `store/remove` to decrease used size...

Services we want to enable signing something would solve this issue. But I would avoid complexity here given this is a temporary solution until we get to work on invocation spec and make this transparent for Providers from ucanto.

upload-api/test/service/ucan-invocation.test.js
```diff
@@ -1,210 +0,0 @@
-import { s3 as test } from '../helpers/context.js'
```
vasco-santos, line 1: All this file content was moved into top level test folder given they are not really within ucanto service.

upload-api/test/ucan-invocation.test.js
```diff
@@ -0,0 +1,406 @@
...
+  t.is(ucan.iss.did(), alice.did())
+  t.is(ucan.aud.did(), uploadService.did())
+  t.deepEqual(ucan.prf, [proof.root.cid])
+  t.is(ucan.att.length, 1)
+  t.like(ucan.att[0], {
+    nb,
+    can,
+    with: spaceDid,
+  })
+})
```
vasco-santos, line 146: All code moved from previous test...

upload-api/test/ucan-invocation.test.js
```diff
@@ -0,0 +1,406 @@
...
+  // @ts-expect-error Property '/' does not exist on type 'Link<Partial<Model>
+  t.is(att[1].nb.root['/'], root.toString())
+
+  // Array with Link
+  t.deepEqual(
+    // @ts-expect-error Property '/' does not exist on type 'Link<Partial<Model>
+    att[1].nb.shards?.map((s) => s['/']),
+    shards.map((s) => s.toString())
+  )
+})
```
vasco-santos, line 372: All code moved from previous test...

alanshaw (approved):
LGTM, left a few minor suggestions.

upload-api/ucan-invocation.js
vasco-santos: @Gozala aligned with Alan to move this way for now in sync A. This is a temporary solution anyway given we want to make this more integrated with ucanto

PR #172, opened 2023-03-22
feat: ucan stream endpoint and upload api with invocations and receipts

vasco-santos:
This PR includes two significant changes:
- `upload-api` ucanto server now follows same pattern as `access-api` proposal from @Gozala's https://github.com/web3-storage/w3protocol/pull/592
  - saves CAR file with UCAN invocations before executing them
  - sends one receipt per invocation when they are executed
- modifies `POST /ucan` endpoint to accept both receipts and CAR invocations as discussed sync yesterday
  - it identifies by ContentType wether is a CAR of invocations or a Receipt (`processUcanLogRequest`)
  - for CAR of invocations, everything is quite similar as before:
    - parse CAR from bytes, persist CAR and send a view of it to Kinesis
    - more than persisting the CARs, we also persist a map of invocation CIDs to CARs where they live in same fashion as we do with DUDEWHERE
    - kinesis now also receives a type property, which can either be Invocation, or receipt like the actual endpoint
    - NOTE: removed `proofs` as they can get quite big and we already talked about getting rid of it before
  - for receipts:
    - we parse the receipt CBOR, verify if given receipt is from an Invocation that we previously got. If so, we can persist the Receipt
    - Kinesis stream receives quite similar content to the invocation. Main differences are that a receipt type is used, and we also include the `inovcationCid`. In other words, we provide value from invocation so that Kinesis consumers can know about `invocation` and `nb` from invocation to filter out and account metrics

Other notes:
- `upload-api` hooks in the same functions for integrating with UCAN Stream as the `/ucan` endpoint, avoiding the HTTP calls
- Storing receipts and invocations Bucket keys were decided yesterday. 
- Like https://github.com/web3-storage/w3protocol/pull/592, we aim to get to ucanto part of the code that we currently inline to create the receipts
- Will do a follow up PR for metric consumers to update to new world


upload-api/buckets/ucan-store.js
```diff
@@ -19,23 +24,81 @@ export function createUcanStore(region, bucketName, options = {}) {
...
+    /**
+     * Put mapping for where each invocation lives in a CAR file.
+     *
+     * @param {string} invocationCid
+     * @param {string} carCid
+     */
+    putInvocation: async (invocationCid, carCid) => {
+      const putCmd = new PutObjectCommand({
+        Bucket: bucketName,
+        Key: `${invocationCid}/${carCid}.invocation`,
```
alanshaw, line 54: 🙏 Please can you document the bucket structure on a README somewhere?

alanshaw: So if I understand this correctly, we get a single CAR file (an invocation), and each root is the CID of a _task_

I think we've been calling tasks => invocations and invocations => requests, but we should probably orient more around invocation+task.

I think we just need to rename `invocationCid` to `taskCid` and `carCid` to `invocationCid`...right?

Our mapping would be:

```sh
# CAR file of invocation, roots are tasks
/invocation-cid/invocation-cid.car

# empty file, pointer back to invocation
/task-cid/invocation-cid.invocation

# CBOR bytes of receipt
/task-cid/receipt-cid.receipt
```

alanshaw: cc @Gozala

vasco-santos: Yes, I will add docs for that! 

I was under the impression that a CAR file had multiple invocations, that is a batch of invocations. So `carCid` is CID of a batch, each invocationCid is the taskCid within the batch.

But yes, would love @Gozala thoughts to converge to UCAN flavour naming

Gozala: > I think we've been calling tasks => invocations and invocations => requests, but we should probably orient more around invocation+task.

Lets not overload invocation plz because of the invocation spec according to which invocation is essentially a singed task. What I was calling a request would be a set or a batch of invocations, or a query in GraphQL terms.

Gozala: Couple of things:

1. Right now we do not have notion of a task because we have not implemented invocation spec. What we have is a lot more like invocation than task.
2. In the future when invocation spec is implemented we would want to lookup receipts by task CID, which is why I kept mentioning task CID 🫣
3. Above means that we may receive multiple invocations for the same task and have multiple corresponding receipt for each one.

With that said I think what we can do currently is either:

1. Derive task (& it's CID) from the delegations we receive (should be as simple as deriving receipts was in the pr) and treat delegation as if it was an Invocation.
   - We could derive invocation as well, but that is probably not a good idea because how would we look things up then ?
1. Tread received delegations as both task CID and invocation CID.

Whatever we decide above I think we should identify three components:

1. Task (basically with, can, nb fields)
2. Invocation (task + proofs + signature)
3. Workflow (CAR containing one or more invocations)

With the above I think following would make most sense to me:

```sh
# Requests blobs likely in CAR but cid codec should tell if it's something else
/${workflow.cid}/${workflow.cid}.workflow

# Empty file telling you the workflow CID
/${task.cid}/${invocation.cid}/${workflow.cid}.workflow
# Receipt block of the receipt for a specific task invocation
/${task.cid}/${invocation.cid}.receipt
```

If we don't have symlinks we should do symlink instead of empty file.

Few notes:

1. One task may have many receipts (one per invocation) all with the same result.
2. We could (in theory) receive same invocation in multiple CARs, which is why we have dir of workflows
3. While we currently use CARs and CBOR I think we may in the future support other formats so I suggest we infer format from the CID.
    - Makes me realize we should probably make endpoint aware of that

vasco-santos: @Gozala thanks for going through all the namings and ideas here.

1. Workflow (CAR containing one or more invocations) keyed as `/${workflow.cid}/${workflow.cid}.workflow` makes sense to me. Would probably keep the `.car` but that is a detail ✅ 
2. Invocation (task + proofs + signature) makes sense to me as well. Would love a bit more detail on the suggested key. `/${task.cid}/${invocation.cid}/${workflow.cid}.workflow`: `invocation.cid` will be the the root of the block within the Workflow, `workflow.cid` is also clear. The `task.cid` looks to be either of your suggestions, encode as CBOR the `with, can, nb fields` and get CID for it, or just use both the same for now?  
3. considering format proposed here, it is not clear on how receipt would look like ⚠️ Looking at [`access-api#592`](https://github.com/web3-storage/w3protocol/blob/feat/ucan-log/packages/access-api/src/routes/root.js#L87), a receipt will only include the `invocationCid` on its log. This way, it won't be possible for us to find out which `task.cid` does this relate to. Unless we just use both same naming for the time being.
4. From above, looks like there is no reason for us to put `receipt.cid` in receipt name as this PR has. But rather just `/${task.cid}/${invocation.cid}.receipt` ✅ 

In short, looks like moving forward with `Tread received delegations as both task CID and invocation CID` would work. Would like to know a bit more about other solution.

> While we currently use CARs and CBOR I think we may in the future support other formats so I suggest we infer format from the CID.
> Makes me realize we should probably make endpoint aware of that

Wondering if we should just suffix as "extension" for now?

Gozala: > 1. Workflow (CAR containing one or more invocations) keyed as `/${workflow.cid}/${workflow.cid}.workflow` makes sense to me. Would probably keep the `.car` but that is a detail

The reason I proposed not to use `.car` extension is specifically to handle the future case where we might use different packet format.

> Invocation (task + proofs + signature) makes sense to me as well. Would love a bit more detail on the suggested key. /${task.cid}/${invocation.cid}/${workflow.cid}.workflow: invocation.cid will be the the root of the block within the Workflow, workflow.cid is also clear.

To be precise Invocation has [following IPLD schema](https://github.com/ucan-wg/invocation/#5-invocation) for now we probably could just use delegation CID, that is root in CAR CID untill we implement support for invocations.

Another option would be for us to take received delegation and issue invocation that matches format in the spec, but that does not seem like a good idea to me as it would be hard to map it back to the car.

> The task.cid looks to be either of your suggestions, encode as CBOR the with, can, nb fields and get CID for it, or just use both the same for now?

Here is the [link to the Task spec](https://github.com/ucan-wg/invocation/#3-task), in essence it is CBOR of those fields but with different names.

I think probably we should just use delegation CID for now here as well.

> considering format proposed here, it is not clear on how receipt would look like ⚠️ Looking at [access-api#592](https://github.com/web3-storage/w3protocol/blob/feat/ucan-log/packages/access-api/src/routes/root.js#L87), a receipt will only include the invocationCid on its log. 

That is correct receipt links to the invocation.

> This way, it won't be possible for us to find out which task.cid does this relate to. Unless we just use both same naming for the time being.

That is true for now and in the future, because receipt links to the invocation and not the task. However Invocation (in spec) links to the Task via `run` field. In other words `Task` cid could be derived from the `Invocation` block which we could do now via CBOR encoding `with, can, nb` fields and in the future reading `.run` field.

We also could considering just providing task CID to the endpoint because we'll know it when we receive the workflow.

Alternatively one could derive `Task` by finding an `invoaction.cid`  in the paths, because invocation can only run a a single task.

Gozala: > From above, looks like there is no reason for us to put receipt.cid in receipt name as this PR has. But rather just /${task.cid}/${invocation.cid}.receipt

There might be a reason to capture receipt cid in the path e.g. to signal encoding so we could do following instead

```
/${workflow.cid}/${workflow.cid}.workflow
/${task.cid}/${invocation.cid}/${workflow.cid}.workflow
/${task.cid}/${invocation.cid}/${receipt.cid}.receipt
```

Gozala: > Wondering if we should just suffix as "extension" for now?

I mean we already have a tag in CID for this so why not just embrace it

Gozala: Me and @alanshaw had a chat and came up with a following plan

1. This is the layout that makes most sense in the future and should map well to what we have now:

```sh
# Request blobs likely in CAR but cid codec should tell if it's something else
/workflow/${workflow.cid}

# Pseudo symlink to `/workflow/${workflow.cid}`
/invocation/${invocation.cid}/${workflow.cid}.workflow

# Receipt block issued for a specific task invocation.
/invocation/${invocation.cid}.receipt

# Block containing `out` field of the receipt. So that when we get an invocation
# with the same task we can read the result and issue receipt without rerunning
# a task. Could be written on first receipt.
/task/${task.cid}.result

# Pseudo symlinks to `/invocation/${invocation.cid}/`
# So we can lookup invocations & receipts by task
/task/${task.cid}/${invocation.cid}.invocation
```

2. For now we'll use delegation CID as both invocation and task CID. Delegation CIDs are the roots in the received CAR and CID in the `.ran` field of the received receipt.

3. When we implement invocation spec we'll change endpoint such that receipts will be send in CARs that will also contain corresponding tasks. That way both invocation and task CID will be known without doing additional lookups.

alanshaw: @Gozala we decided to put workflow, invocation and task items in separate buckets to avoid S3 rate limits on a common prefix and still retain the grouping.

vasco-santos: Thanks for syncing and reporting back last night. As talked with Alan today;s morning, I like the format but that brings rate limit issues in buckets. Therefore, moved forward with:

```
# Request blobs likely in CAR but cid codec should tell if it's something else
${workflow.cid}/${workflow.cid}

# Pseudo symlink to `/workflow/${workflow.cid}`
${invocation.cid}/${workflow.cid}.workflow

# Receipt block issued for a specific task invocation.
${invocation.cid}/${invocation.cid}.receipt

# Block containing `out` field of the receipt. So that when we get an invocation
# with the same task we can read the result and issue receipt without rerunning
# a task. Could be written on first receipt.
${task.cid}/${task.cid}.result

# Pseudo symlinks to `/invocation/${invocation.cid}/`
# So we can lookup invocations & receipts by task
${task.cid}/${invocation.cid}.invocation

```

Note that we could also do:

```
# Pseudo symlink to `/workflow/${workflow.cid}`
${invocation.cid}/${invocation.cid}/${workflow.cid}.workflow

# Pseudo symlinks to `/invocation/${invocation.cid}/`
# So we can lookup invocations & receipts by task
${task.cid}/${task.cid}/${invocation.cid}.invocation
```

but I feel that it is not needed? Thoughts?

stacks/ucan-invocation-stack.js
```diff
@@ -28,29 +27,33 @@ export function UcanInvocationStack({ stack, app }) {
...
-      function: ucanPutEventConsumer,
-      events: ['object_created_put'],
+  const taskBucket = new Bucket(stack, 'task-store', {
+    cors: true,
+    cdk: {
+      bucket: getBucketConfig('task-store', app.stage)
+    }
+  })
+
+  // TODO: keep for historical content that we might want to process
```
alanshaw, line 52: Does this definitely keep it around? The act of instantiating a class registers it?

upload-api/buckets/invocation-store.js
```diff
@@ -0,0 +1,79 @@
...
+    putReceipt: async (invocationCid, bytes) => {
+      const putCmd = new PutObjectCommand({
+        Bucket: bucketName,
+        Key: `${invocationCid}/${invocationCid}.receipt`,
+        Body: bytes,
+      })
+      await pRetry(() => s3client.send(putCmd))
+    },
+    /**
+     * Get CAR bytes for a given invocation.
```
alanshaw, line 59:
```suggestion
     * Get the workflow CID for an invocation.
```

upload-api/buckets/task-store.js
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @param {S3Client} s3client
+ * @param {string} bucketName
+ * @returns {import('../types').TaskBucket}
+ */
+export const useTaskStore = (s3client, bucketName) => {
+  return {
+    /**
+     * Put block containing `out` filed of the receipt. So that when we get an invocation
```
alanshaw, line 31:
```suggestion
     * Put block containing `out` field of the receipt, so that when we get an invocation
```

upload-api/buckets/task-store.js
```diff
@@ -0,0 +1,61 @@
...
+     * with the same task we can read the result and issue receipt without rerunning
+     * a task. Could be written on first receipt.
+     *
+     * @param {string} cid
+     * @param {Uint8Array} bytes
+     */
+    putResult: async (cid, bytes) => {
+      const putCmd = new PutObjectCommand({
+        Bucket: bucketName,
+        Key: `${cid}/${cid}.result`,
```
alanshaw, line 41: I'd explicitly name this so it is not confusing what this CID is referring to.

```suggestion
     * @param {string} taskCid
     * @param {Uint8Array} bytes
     */
    putResult: async (taskCid, bytes) => {
      const putCmd = new PutObjectCommand({
        Bucket: bucketName,
        Key: `${taskCid}/${taskCid}.result`,
```

upload-api/buckets/task-store.js
```diff
@@ -0,0 +1,61 @@
...
+     * Put mapping for where each task lives in an invocation file.
+     * Enables lookup invocations & receipts by task.
+     *
+     * @param {string} cid 
+     * @param {string} invocationCid 
+     */
+    putIndex: async (cid, invocationCid) => {
+      const putCmd = new PutObjectCommand({
+        Bucket: bucketName,
+        Key: `${cid}/${invocationCid}.invocation`,
```
alanshaw, line 56:
```suggestion
     * @param {string} taskCid 
     * @param {string} invocationCid 
     */
    putIndex: async (taskCid, invocationCid) => {
      const putCmd = new PutObjectCommand({
        Bucket: bucketName,
        Key: `${taskCid}/${invocationCid}.invocation`,
```

upload-api/buckets/invocation-store.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+export const useInvocationStore = (s3client, bucketName) => {
+  return {
+    /**
+     * Put mapping for where each invocation lives in a Workflow file.
+     *
+     * @param {string} invocationCid
+     * @param {string} workflowCid
+     */
+    putIndex: async (invocationCid, workflowCid) => {
```
alanshaw, line 37:
```suggestion
    putWorkflowIndex: async (invocationCid, workflowCid) => {
```

...might be better to name this "link" rather than "index"

upload-api/ucan-invocation.js
```diff
@@ -1,46 +1,145 @@
...
+  if (!request.body) {
+    throw new BadBodyError('service requests are required to have body')
+  }
+  const bytes = Buffer.from(request.body, 'base64')
+
+  const contentType = request.headers['Content-Type'] || ''
+  if (contentType === CONTENT_TYPE.WORKFLOW) {
+    return await processWorkflow(bytes, ctx)
+  } else if (contentType === CONTENT_TYPE.RECEIPT) {
+    return await processTaskReceipt(bytes, ctx)
```
alanshaw, line 49: This is an invocation receipt.

```suggestion
    return await processInvocationReceipt(bytes, ctx)
```

alanshaw (changes requested):

stacks/ucan-invocation-stack.js
vasco-santos: Yes, it keeps in Cloudformation. in reality we could even remove it... In prod/staging it won't get deleted per https://github.com/web3-storage/w3infra/blob/main/stacks/config.js#L38

But better to keep it around for now to hook up consumers if we want to still read from there. Otherwise, we will need to hook it up with ARN id

alanshaw (approved):

PR #174, opened 2023-03-23
feat: ucan stream handles receipts

vasco-santos:
This PR hooks kinesis consumers with receipts, instead of previous world where it would hook up with workflows. For this, it simply filters out Kinesis events that are not receipts as they should not be account for the metrics we want to track at the moment.

As talked in today`s morning sync, setting up a new Kinesis stream so that:
- we reset clock history of 1 year
- we still keep old stream around so that we can persist what is there to S3 (or even re-process it and pipe to the new stream with receipts and needed files in correct buckets!)

### DIFF
--- a/stacks/ucan-invocation-stack.js
+++ b/stacks/ucan-invocation-stack.js
@@ -189,8 +189,17 @@ export function UcanInvocationStack({ stack, app }) {
     deadLetterQueue: spaceMetricsDLQ.cdk.queue,
   })
 
+  // TODO: keep for historical content that we might want to process
+  new KinesisStream(stack, 'ucan-stream', {
+    cdk: {
+      stream: {
+        retentionPeriod: Duration.days(365)
+      }
+    },
+  })
+
   // create a kinesis stream
-  const ucanStream = new KinesisStream(stack, 'ucan-stream', {
+  const ucanStream = new KinesisStream(stack, 'ucan-stream-v2', {
     cdk: {
       stream: {
         retentionPeriod: Duration.days(365)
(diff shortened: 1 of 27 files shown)


stacks/ucan-invocation-stack.js
```diff
@@ -189,8 +189,17 @@ export function UcanInvocationStack({ stack, app }) {
     deadLetterQueue: spaceMetricsDLQ.cdk.queue,
   })
 
+  // TODO: keep for historical content that we might want to process
```
vasco-santos, line 192: See 2 streams there https://us-east-2.console.aws.amazon.com/kinesis/home?region=us-east-2#/streams/list

alanshaw (approved):

ucan-invocation/constants.js
```diff
@@ -7,6 +7,11 @@ import {
   remove as uploadRemove
 } from '@web3-storage/capabilities/upload'
 
+export const CONTENT_TYPE = {
+  WORKFLOW: 'application/invocations+car',
+  RECEIPT: 'application/receipt+dag-cbor'
```
olizilla, line 12: - the + suffix denotes an existing media type which are registered as `vnd.ipld.car` and `vnd.ipld.dag-cbor`
- the prefix is a tree, and we can add things to the vnd tree without much fuss like:
```
application/vnd.w3.ucan+vnd.ipld.car
application/vnd.w3.ucan.receipt+vnd.ipld.dag-cbor
```
of if we really don't care then
```
application/x.ucan+vnd.ipld.car
application/x.ucan.receipt+vnd.ipld.dag-cbor
```
see: https://en.wikipedia.org/wiki/Media_type#Naming

olizilla: https://github.com/ipfs/in-web-browsers/issues/201
https://github.com/ipfs/in-web-browsers/issues/202

Issue #170, opened 2023-03-17
UCAN Stream Consumers for access-api invocations


vasco-santos:
TBD

PR #182, opened 2023-03-30
feat: ucan stream consumer provider add count

vasco-santos:
Adds UCAN Stream consumer to track provider/add count success receipts as part of https://github.com/web3-storage/w3infra/issues/170. The count of this receipts will mean number of registered spaces with a provider.

Question: should we track spaces count in provider add by given provider? This would require us to have a different table, or change the current schema  of admin table, given we would need to columns as keys.


travis:
@alanshaw I'm going to close this for now to try to delete the attached dev env, open to re-opening if we want to get this in!

Issue #186, opened 2023-04-10
Update ucanlog to ucanto@7.0.0


Gozala:
ucanto is introducing major breaking changes in 7.0 in an attempt to avoid breaking changes in the future, specifically it's wire protocol was updated to send versioned messages between (client/server) agents, which look as follows (for more details see https://github.com/web3-storage/ucanto/pull/274)

```ts
/**
 * Describes messages send across ucanto agents.
 */
export type AgentMessageModel<T> = Variant<{
  'ucanto/message@7.0.0': AgentMessageData<T>
}>

/**
 * Describes ucanto@7.0.0 message format send between (client/server) agents.
 *
 * @template T - Phantom type capturing types of the payload for the inference.
 */
export interface AgentMessageData<T> extends Phantom<T> {
  /**
   * Set of (invocation) delegation links to be executed by the agent.
   */
  execute?: Tuple<Link<UCAN.UCAN<[Capability]>>>

  /**
   * Map of receipts keyed by the (invocation) delegation.
   */
  report?: Record<ToString<UCANLink>, Link<ReceiptModel>>
}
```

TLDR; CARs sent will contain single root pointing to the `AgentMessageModel` which is variant (a.k.a [keyed union](https://ipld.io/docs/schemas/features/representation-strategies/#union-keyed-representation)). At the moment variant has a single branch corresponding to ucanto 7.0 schema, but in the future we'll add branches as we implement different versions of invocation spec (so we can infer which version of wire protocol used).

Current version tagged as `ucanto/message@7.0.0` has links to:

1. Set of invocation delegations (which used to be the CAR roots in the requests up until ucanto 7.0).
1. Set of receipt links

While right now all requests will only contain only invocations and all responses receipts, in the future we may start dispatching invocations back to client agents and consequently also get receipts for those from the, which is why `AgentMessageData` has both fields. Both feilds are optional and are omitted if corresponding set is empty, although this is only enforced at the type level.

---

Deployed UCAN log server now receives bytes on on `/ucan/` endpoint and based on the content-type header either treats it as invocations (if `'application/invocations+car'`)  or receipts (if `'application/receipt+dag-cbor'`) https://github.com/web3-storage/w3infra/blob/abe2410fd43e7183c6d4a333277e2bd605726606/upload-api/ucan-invocation.js#L39-L57

We should update the `/ucan/` endpoint to instead to receive CARs containing `AgentMessageModel` in a root and index invocations & receipts from it instead.

[...]

PR #187, opened 2023-04-11
feat: ucan endpoint supports agent message

vasco-santos:
As agreed, first iteration adds support for `POST /ucan` into the API. `upload-api` is still using old world, therefore we need to keep both `ucanto` dependencies for this first iteration. Code changes aim to just drop most of the functions in place in favour of the added ones once in next PR we update ucanto in `upload-api`.

Decisions:
- new bucket structures based on https://github.com/web3-storage/w3infra/issues/186#issuecomment-1502309162
- we install `"@ucanto/core-next": "npm:@ucanto/core@^7.0.1"` + `"@ucanto/transport-legacy": "npm:@ucanto/transport@^5.1.1"`
  - initial goal here was to install new ones and keep olds with suffix `-legacy` to easily drop them in next PR. Sadly, given other `ucanto` packages depend on the 5th line of core, I could not work around having `"@ucanto/core-legacy": "npm:@ucanto/core@^5.1.0"` because dependencies would struggle to find correct one...
  - path forward is still relatively easy given `ucanto/core` is barely used
- tried to avoid naming workflow everywhere, moving to agentMessage to uniform with ucanto

Closes https://github.com/web3-storage/w3infra/issues/186


upload-api/ucan-invocation.js
```diff
@@ -56,6 +71,176 @@ export async function processUcanLogRequest (request, ctx) {
...
+      throw new NoCarFoundForGivenReceiptError()
+    }
+
+    const agentMessageBytes = await ctx.workflowBucket.get(agentMessageWithInvocationCid)
+    if (!agentMessageBytes) {
+      throw new NoCarFoundForGivenReceiptError()
+    }
+
+    // @ts-expect-error headers are not needed
+    const agentMessage = await CAR.request.decode({ body: agentMessageBytes })
```
Gozala, line 151: Nit: you can just pass `headers: {}` instead.

upload-api/ucan-invocation.js
```diff
@@ -1,5 +1,6 @@
 import * as CAR from '@ucanto/transport/car'
-import * as CBOR from '@ucanto/transport/cbor'
+import * as CAR_LEGACY from '@ucanto/transport-legacy/car'
+import * as CBOR_LEGACY from '@ucanto/transport-legacy/cbor'
```
Gozala, line 3: BTW `@ucanto/core` exports `CBOR` codec now.

Gozala (approved):
Awesome! Thanks for getting it all done so quickly! It all looks good although I made some minor suggestions here and there.

I do however find naming misleading however, because through out `agentMessage` is used to refer to the CAR file containing an `AgentMessage` block in a root. I would suggest update naming to call it `agentArchive` or `agentMessageArchive` instead to disambiguate CAR containing the message from the message block itself.

docs/ucan-invocation-stream.md
```diff
@@ -29,11 +30,11 @@ Note that at the time of writing Event Archival flow is still to be implemented.
 
 UCAN Invocation Stack contains 3 buckets so that it can keep an audit of the entire system, while allowing this information to be queried in multiple fashions.
 
-Firstly, the **`workflow-store` bucket** stores the entire encoded file containing one or more invocations to be executed. It is stored as received from UCAN services interacting with UCAN Invocation Stream. It is keyed as `${workflow.cid}/${workflow.cid}` and its value is likely in CAR format. However, CID codec should tell if it is something else.
+Firstly, the **`workflow-store` bucket** stores the entire encoded agent message files containing invocations to be executed, and/or created receipts for ran invocations. It is stored as received from UCAN services interacting with UCAN Invocation Stream. It is keyed as `${agentMessage.cid}/${agentMessage.cid}` and its value is likely in CAR format. However, CID codec should tell if it is something else.
```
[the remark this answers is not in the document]
vasco-santos: Per suggestion in review text, I changed this into `agentMessageArchive.cid`. Let me know if it is more clear this way

vasco-santos:
@Gozala addressed all your review, looks like we might need to still align on 1 point:
- https://github.com/web3-storage/w3infra/pull/187#discussion_r1163770550

Given PR is approved, I will move forward and merge this in, so that we can start working on `access-api` and `upload-api` upgrade. Happy to do any needed change in upcoming PR based on your answer there

Issue #190, opened 2023-04-13
Real time usage metrics by logs from UCAN Stream


vasco-santos:
## Context

Within w3up infrastructure, through the UCAN stream we track system wide metrics and space metrics. These metrics allow us to know overall system metrics and w3up users to know about their total usage. However, we have no visibility on real time volume of usage by each space. Based on the operation of older APIs, we see value on knowing usage volume, so that we can proactively avoid abused and get to know patterns.

## Requirements

- Ability to visualise real time usage for the spaces with most usage (ideally in Grafana)
  - ability to filter by capability executed, and if receipt or workflow

vasco-santos:
## Available Options

1. Kinesis Data Analytics for Apache Flink + Amazon timestream
2. Kinesis Data Firehose + S3 + Athena

## 1. Kinesis Data Analytics for Apache Flink + Amazon timestream

- Kinesis Data Analytics for Apache Flink
  - One consumer for the data stream that preprocesses and ingests data into Timestream.
  - Needs Apache flink application with tables

- Amazon Timestream
  - scalable serverless time series database for operational events ingestion and querying
  - allows you to configure different retention periods for your tables to optimize storage costs
  - There is a [grafana plugin](https://grafana.com/grafana/plugins/grafana-timestream-datasource/)

References
- https://docs.aws.amazon.com/timestream/latest/developerguide/ApacheFlink.html
- https://github.com/aws-samples/amazon-kinesis-timestream-grafanahttps://docs.aws.amazon.com/kinesis/index.html#amazon-kinesis-data-analytics
- https://aws.amazon.com/blogs/database/near-real-time-processing-with-amazon-kinesis-amazon-timestream-and-grafana/

## 2. Kinesis Data Firehose + S3 Data Lake + Athena

- Kinesis Data Firehose
    - fully managed service for delivering real-time streaming data to destinations such as Amazon Simple Storage Service (Amazon S3)
    - no need to write applications or manage resources. You configure your data producers to send data to Kinesis Data Firehose, and it automatically delivers the data to the destination that you specified
    - also possible to configure Kinesis Data Firehose to transform data before delivering it
- Amazon Athena
    - Provides a simplified, flexible way to analyze petabytes of data where it lives. Analyze data or build applications from an Amazon Simple Storage Service (S3) data lake
    - Can be integrated with Grafana
        - https://grafana.com/grafana/plugins/grafana-athena-datasource/
[...]

PR #191, opened 2023-04-14
feat: Kinesis delivery stream and Athena query infrastructure to enable queries over the UCAN logs

vasco-santos:
This PR has an implementation of option 2 from https://github.com/web3-storage/w3infra/issues/190#issuecomment-1506618126

This encompasses a fair amount of functionality - partitioning the UCAN logs into S3 buckets, configuring a Glue database and tables, adding example queries to Athena and more. A partial list of functionality follow:

- implement UCAN log partitioning in S3
  - first partition by `type` - everything in "workflows" shows up in "receipts" so this reduces the amount of data scanned by ~50%
  - next partition by `op` to allow us to create tables that only query a specific operation (ie, `store/add` or `provider/add`) - this lets us add operation-specific Glue table schemas with much less clutter in result types than we'd need if we tried to defined all possible inputs and outputs in a single table
  - finally partition by date to allow queries to only load recent data

- use these partitions to implement standalone tables for receipts in general and the `store/add`, `upload/add` and `provider/add` UCANs specifically, 
- add queries that demonstrate the use of all of these tables
- add dynamo connector so we can join the UCAN logs to our Dynamo tables in queries
- add queries that demonstrate using the Dynamo and Glue tables together


vasco-santos (commented):
This looks great to me! Thanks for moving this forward @travis ❤️

travis (approved):

PR #192, opened 2023-04-20
feat: implement DelegationsStorage on top of DynamoDB

travis:
This is a first step toward https://github.com/web3-storage/w3up/issues/711 as documented in https://github.com/web3-storage/w3up/pull/744/files?short_path=4e012d9#diff-4e012d97ece45b455806fabbe1f93ce11af56213b9137463a427e1c5d4d8bda6

This is an implementation spike of storing delegations in DynamoDB rather than D1.

Because we already use DynamoDB in the upload-api, it made sense to reuse the configuration and testing infrastructure we've already built there. I copied a number of interfaces from https://github.com/web3-storage/w3up/tree/main/packages/access-api to get this done, and will migrate back to them soon.

Next steps:

- [] finish implementing provisions table
- [] port the delegations-related service endpoints (but *not* the service provider implementations!) in `w3up/access-api` over to this repository and get them running in an AWS Lambda
- [] decide whether we want to use R2 or S3 for delegations, point the S3 code in this PR at R2 if we want to use that

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Next, ensure the `AWS_REGION`, `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` e
 not need to be set to real values - the following works in `bash`-like shells:
 
 ```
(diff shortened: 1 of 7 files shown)


travis:
subsumed by #194

PR #193, opened 2023-05-05
feat!: upgrade to ucanto@7

Gozala:
Updates to ucanto@7 and removes support for legacy formats. We can release w3up first and then deploy the w3infra to avoid loosing any data


upload-api/test/access-verifier.js
```diff
@@ -60,8 +62,9 @@ export const create = ({ id }) => {
   const client = Client.connect({
     id,
     channel: server,
-    encoder: CAR,
-    decoder: CBOR,
+    // encoder: CAR,
+    // decoder: CBOR,
```
Gozala, line 66:
```suggestion
```

upload-api/test/helpers/resources.js
```diff
@@ -4,7 +4,8 @@ import { S3Client, CreateBucketCommand } from '@aws-sdk/client-s3'
 import { DynamoDBClient, CreateTableCommand } from '@aws-sdk/client-dynamodb'
 import * as Signer from '@ucanto/principal/ed25519'
 import * as Server from '@ucanto/server'
-import { CAR, CBOR } from '@ucanto/transport-legacy'
+import * as Legacy from '@ucanto/transport/legacy'
+// import { CAR, CBOR } from '@ucanto/transport-legacy'
```
Gozala, line 8:
```suggestion
```

upload-api/functions/ucan-invocation-router.js
```diff
@@ -108,90 +108,41 @@ async function ucanInvocationRouter(request) {
...
-    out.push(receipt.data.out.error || receipt.data.out.ok)
-    forks.push(processInvocationReceipt(receipt.bytes, processingCtx))
-  }
+  // Execute invocations
+  const outgoing = await Server.execute(incoming, server)
+  const response = CAR.response.encode(outgoing)
 
-  await Promise.all(forks)
-  const response = await CBOR.encode(out)
+  processAgentMessageArchive(response, processingCtx)
```
vasco-santos, line 131: we were awaiting on this before. Should we keep until we wire up aa dead letter queue?

Gozala: Ah good catch, I didn't intend to change this

```suggestion
  await processAgentMessageArchive(response, processingCtx)
```

Gozala:
Ok finally fixed the segfault, it was caused by aws minor version updates that went into the lockfile after package updates. End up resolving by getting rid of those updates

upload-api/functions/ucan-invocation-router.js
vasco-santos: @Gozala this still did not get changed. Maybe commit got lost

vasco-santos (approved):
LGTM just needs the promise to be awaited in the UCAN handler

PR #194, opened 2023-05-05
feat!: fold `access-api` into `upload-api`

travis:
Get `@web3-storage/access-api` handlers running in the `upload-api` service. 

This introduces two new DynamoDB tables to store "delegations" and "provisions". NB: I have a related PR in https://github.com/web3-storage/w3infra/pull/200 that uses "consumers" and "subscriptions" tables rather than "provisions".

This also adds a new Lambda to handle the email verification flow - I've brought some HTML utilities over from `@web3-storage/access-api` for this purpose and wired the handlers up to the same path they were at in the old service `/validate-email`.

This is the part 2 of a larger project (started in https://github.com/web3-storage/w3up/pull/790) to integrate the access service into the w3infra upload service, which will let us move operational details out of w3up, retire D1, consolidate our operations around AWS Lambdas, deprecate now-unused functionality like the voucher flows and generally streamline our UCAN handlers and the infrastructure that supports them.

### TODO

- [x] explore subscription/consumer tables rather than tracking "provisions" 
  -  as suggested in https://github.com/web3-storage/w3infra/pull/194#pullrequestreview-1417122161
  - implemented in https://github.com/web3-storage/w3infra/pull/200
- [x] get delegations and storage provider tests from w3up running here
- [x] explore not storing delegations in a separate bucket, instead pointing to invocations that have already been stored (https://github.com/web3-storage/w3infra/pull/194#discussion_r1187677261)
  - implemented in https://github.com/web3-storage/w3infra/pull/201
- [x] update integration tests (including https://github.com/web3-storage/w3infra/pull/201#discussion_r1210465723)
- [x] update dependencies once a new `@web3-storage/upload-api` is released
- deployment
  - [ ] set new env vars in staging and production ([upload-api/functions/ucan-invocation-router.js](https://github.com/web3-storage/w3infra/pull/194/files#diff-6bc46fb625dd65ffce15531f86723dd48392c2ceadb859182b0acc7090274994))
  - [x] plan, write (and run?) migrations


upload-api/tables/provisions.js
```diff
@@ -0,0 +1,118 @@
...
+            ':cid': { 'S': row.cid },
+            ':consumer': { 'S': row.consumer },
+            ':provider': { 'S': row.provider },
+            ':sponsor': { 'S': row.sponsor }
+          }
+        }))
+      } catch (error) {
+        if (error instanceof Error && error.message === 'The conditional request failed') {
+          return new ConflictError({
+            message: `Space ${row.consumer} cannot be provisioned with ${row.provider}: it already has a provider`
```
Gozala, line 98: It is worth a note that this is intended to be a temporary limitation until we resolve https://github.com/web3-storage/w3up/issues/167#issuecomment-1320468187

I suggest rephrasing an error message in a way that instructs user what they can do, which is create a new space and associate a provider with it or use the space with provider it already has. Mentioning that this limitation may be lifted in the future also might be a good idea.

Gozala (commented):
As a direct port of what we have in d1 this looks all right. That said, I think taking bit more time to avoid another migration down the line might be in our benefit, from that perspective I have following notes:

- We need two tables as opposed to currently 1 provisions table, which
  > That way when you first login web3.storage can delegate an account a subscription that account can later add a consumer (space) to. 
    1. Tracks subscriptions as in provider <-> customer relationship which is where provider could impose any constraints it wishes on the customer.
    2. Tracks subscription consumer provider <-> consumer <-> subscription relationship.

- We need to support different actors to perform following queries
   - provider:
      - find my customers (accounts that have subscriptions)
      - find my consumers (spaces subscribed)
      - cancel subscription
      - block customer
      - stats per customer / consumer / subscription
   - space
      - what are my providers / subscriptions
      - unsubscribe
      - stats / limits
   - customer (account)
      - cancel subscription
      - what are my subscriptions
      - stats per subscriptions

- we need to make provider capable of imposing whatever constraints it wants on accounts.
   - I propose accomplishing this via additional `order` field with uniqueness constraint. This way provider can impose all the constraints it wants by encoding them into the field generation.
      - e.g. web3.storage can derive order from the account email address.
      - nft.storage could choose alternative that would fit it's goals.

upload-api/access-types.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * get number of stored items
+   */
+  count: () => Promise<bigint>
+
+  /**
+   * find all items that match the query
+   */
+  find: (query: Query) => AsyncIterable<Ucanto.Delegation<Ucanto.Tuple<Cap>>>
```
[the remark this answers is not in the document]
travis: yep, works for me!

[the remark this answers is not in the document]
travis: Hm - after looking at this a bit more I'm second guessing this - we can't really know if the underlying system will support streaming results and it does seem valuable to allow novel implementations to stream results back from this function if they want. 

Since the system is already implemented with this async implementation I'm definitely inclined to keep it as-is to both simplify this change and give future implementors maximum flexibility.

@Gozala what do you think? I'm still open to making this a regular `Iterable` but am inclined to keep it as is...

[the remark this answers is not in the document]
Gozala: I'm of a position that this is premature optimization. If we deal with a system that allows us to do streaming we:

1. Should still consider if that would be a right choice, or e.g. whether paging would serve us better.
2. If we want to take advantage of streamed output then we can design solution that fits best the use case, which may or may not be async iterables.
    - For what it's worth in my experience async iterable are rarely the right tool for the job. Even with async streams you often want to have some control of coordination (telling sender to pause / slow down or speed up) and async iterables unlike e.g. web streams don't provide that. They also provide little insight if multiple readers would get same output or they'll get random parts when reading concurrently.

To summarize I think `AsyncIterable`s provide no value now and likely aren't going to be a good choice in the future. That said leaving it out of scope for this PR is reasonable and I'm not going to object to it if that is what you want to do.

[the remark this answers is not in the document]
travis: ok I'm convinced! have changed this interface in https://github.com/web3-storage/w3up/pull/790/commits/026ab8726b73d0ce907a409898325f365bd3cab6

stacks/upload-db-stack.js
```diff
@@ -30,6 +32,23 @@ export function UploadDbStack({ stack, app }) {
...
 
+  /**
+   * This table tracks space provisioning.
+   */
+  const provisionTable = new Table(stack, 'provision', provisionTableProps)
+
+  /**
+   * This table tracks metrics per space.
+   */
+  const delegationTable = new Table(stack, 'delegation', delegationTableProps)
```
vasco-santos, line 43: What kind of metrics this is?

stacks/upload-db-stack.js
```diff
@@ -30,6 +32,23 @@ export function UploadDbStack({ stack, app }) {
...
+
+  const delegationBucket = new Bucket(stack, 'delegation-store', {
+    cors: true,
+    cdk: {
+      bucket: getBucketConfig('delegation', app.stage)
+    }
+  })
```
vasco-santos, line 50: this stack only includes databases, and I would keep it that way. We need to thing where the best place for this bucket is.

upload-api/tables/delegations.js
```diff
@@ -0,0 +1,161 @@
...
+export function useDelegationsTable (dynamoDb, tableName, bucket) {
+  return {
+    putMany: async (...delegations) => {
+      if (delegations.length === 0) {
+        return {
+          ok: {}
+        }
+      }
+      await writeDelegations(bucket, delegations)
+      // TODO: we should look at the return value of this BatchWriteItemCommand and either retry or clean up delegations that we fail to index
```
vasco-santos, line 54: I guess that for metrics might be good to have this output in the receipt for UCAN invocation. What do you think?

upload-api/tables/delegations.js
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * TODO: can we use the function in w3up access-api/src/models/delegations.js?
+ * 
+ * @param {import('../types').DelegationsBucket} bucket
+ * @param {Iterable<import('@ucanto/interface').Delegation>} delegations
+ */
+async function writeDelegations (bucket, delegations) {
+  return writeEntries(
```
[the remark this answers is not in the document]
vasco-santos: Was also thinking about that, that way we do not even need this extra bucket anymore

upload-api/tables/index.js
```diff
@@ -26,3 +26,65 @@ export const uploadTableProps = {
   // space + root must be unique to satisfy index constraint
   primaryIndex: { partitionKey: 'space', sortKey: 'root' },
 }
+
+/** @type TableProps */
+export const delegationTableProps = {
+  fields: {
+    cid: 'string',        // `baf...x`
```
vasco-santos, line 33: we call cid link in other tables. Should we do the same here?

upload-api/tables/index.js
```diff
@@ -26,3 +26,65 @@ export const uploadTableProps = {
...
+  fields: {
+    cause: 'string',        // `baf...x` (CID of invocation that created this provision)
+    consumer: 'string',   // `did:key:space` (DID of the actor that is consuming the provider, e.g. a space DID)
+    provider: 'string',   // `did:web:service` (DID of the provider, e.g. a storage provider)
+    order: 'string',      // string (arbitrary string associated with this subscription)
+    insertedAt: 'string', // `2022-12-24T...`
+    updatedAt: 'string',  // `2022-12-24T...`
+  },
+  // TODO does this index setup seem right?
+  primaryIndex: { partitionKey: 'provider', sortKey: 'order' },
```
vasco-santos, line 89: maybe we want the consumer as `sortKey`? we need to flush out what kind of queries we need here

upload-api/tables/provisions.js
```diff
@@ -0,0 +1,188 @@
...
+ * @param {DynamoDBClient} dynamoDb
+ * @param {string} tableName
+ * @param {import('@ucanto/interface').DID<'web'>[]} services
+ * @returns {import('@web3-storage/upload-api').ProvisionsStorage}
+ */
+export function useProvisionsTable (dynamoDb, tableName, services) {
+  return {
+    services,
+    hasStorageProvider: async (consumer) => {
+      const cmd = new QueryCommand({
```
vasco-santos, line 49: Looks like we only need one. Can't we use GetItemCommand instead here? I believe it will be more efficient than a query

upload-api/tables/provisions.js
```diff
@@ -0,0 +1,188 @@
...
+ * @param {string} subscriptionsTableName
+ * @param {string} consumersTableName
+ * @param {import('@ucanto/interface').DID<'web'>[]} services
+ * @returns {import('@web3-storage/upload-api').ProvisionsStorage}
+ */
+export function useProvisionsStorage (dynamoDb, subscriptionsTableName, consumersTableName, services) {
+  return {
+    services,
+    hasStorageProvider: async (consumer) => {
+      const cmd = new QueryCommand({
```
vasco-santos, line 125: Same as below. Perhaps just get?

upload-api/tables/provisions.js
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/**
+ * @param {DynamoDBClient} dynamoDb
+ * @param {string} subscriptionsTableName
+ * @param {string} consumersTableName
+ * @param {import('@ucanto/interface').DID<'web'>[]} services
+ * @returns {import('@web3-storage/upload-api').ProvisionsStorage}
+ */
+export function useProvisionsStorage (dynamoDb, subscriptionsTableName, consumersTableName, services) {
```
vasco-santos, line 121: These 2 seem quite similar. Maybe we can unify them?

vasco-santos (changes requested):
Thanks for putting this together @travis 
Left some initial feedback. Note that we should also update the integration tests here to include this, given now we have everything in same place

upload-api/buckets/delegations-store.js
```diff
@@ -0,0 +1,78 @@
...
+ * @param {string} bucketName
+ * @param {import('@aws-sdk/client-s3').ServiceInputTypes} [options]
+ */
+export function createDelegationsStore(region, bucketName, options = {}) {
+  const s3client = new S3Client({
+    region,
+    ...options,
+  })
+  return useDelegationsStore(s3client, bucketName)
+}
```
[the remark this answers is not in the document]
travis: we do use this in a couple different places, and I think providing this function makes that code cleaner than if it had to create the S3Client itself, so I'd like to keep this

upload-api/tables/delegations.js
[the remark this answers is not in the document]
travis: ok great - I'm down to make this change. I think this means we'd need to change the `putMany` method of `DelegationsStorage` from this:

```js
  putMany: (
    ...delegations: Array<Ucanto.Delegation<Ucanto.Tuple<Cap>>>
  ) => Promise<Ucanto.Result<{}, never>>
```

to something that takes an invocation CID, something like:

```js
  putMany: (
    invocationCID: CID,
    ...delegations: Array<Ucanto.Delegation<Ucanto.Tuple<Cap>>>
  ) => Promise<Ucanto.Result<{}, never>>
```

and then store the invocationCID in dynamo with each of the rows we create for the delegations. 

Does that sound correct?

stacks/upload-db-stack.js
travis: doh! copypasta fail

stacks/upload-db-stack.js
travis: ok - I've moved this out to `upload-api-stack.js` for now - happy to move it somewhere else but this seems like a decent default for now?

upload-api/tables/delegations.js
travis: hm I'm not sure I follow... are you saying that you think it would be good to record when this fails in the receipt for the invocation of the UCAN that kicked off this write? sounds right to me - I think all we need to do is return an appropriate error when this fails, is that right?

worth noting that this code will be much simpler if we stop saving delegations to a separate bucket!

upload-api/tables/index.js
travis: works for me!

upload-api/tables/index.js
travis: yea - I've started to do that a bit in https://github.com/web3-storage/w3infra/pull/194/files#diff-924a94ba76534a93c7fa4c89f8e70037f21694b6374beaca7a2c6ebf3fdb2006 - I've changed the partition and sort keys for these tables - definitely don't want to use provider as it will be one of two values for a while...

upload-api/tables/provisions.js
travis: removed this in favor of subscription/consumer tables

upload-api/tables/provisions.js
travis: removed this in favor of subscription/consumer tables

stacks/upload-db-stack.js
vasco-santos: I would likely move it to other place. So the reasoning is: we try to keep stacks with state separated from stateless stacks. Having it in `upload-api` adds a dependency that we cannot destroy the stack anymore and recreate.

Anyway, as far as I understand, we are getting rid of this right?

travis: ahh, got it - ok will find it a new home!

re: getting rid of it, good question - it looks like we may need to keep it around for legacy data, but we might be able to just point at R2 rather than move that data to S3 - in that case I'm not entirely sure how this configuration code would work - SST doesn't need to be responsible for creating and managing the bucket, so we should just need to pass some config in somewhere - let's find time to chat about how that would work!

vasco-santos: Yeah, if we want to just keep the R2 bucket, we just need to specify ENV Vars for the R2 bucket. You can see the replicator stack as we also use R2 buckets there

upload-api/tables/delegations.js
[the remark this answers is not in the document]
travis: ok - I have a preliminary implementation of this in https://github.com/web3-storage/w3up/pull/800 and https://github.com/web3-storage/w3infra/pull/201 - @Gozala @vasco-santos what do you think?

stacks/upload-db-stack.js
travis: ok - I've moved to creating this bucket the same way we do in the replicator stack: https://github.com/web3-storage/w3infra/pull/201

upload-api/tables/delegations.js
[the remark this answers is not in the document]
Gozala: I think we should do what the linked PRs do. Have no opinion if we want to land all together or in iterations.

upload-api/tables/consumer.js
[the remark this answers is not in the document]
Gozala: Just found it in `tables/index.js`. I think it would make more sense to export schema for this table from here instead, but I'm fine with keeping thins as as well.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
...
   primaryIndex: { partitionKey: 'space', sortKey: 'root' },
 }
+
+/** @type TableProps */
+export const delegationTableProps = {
+  fields: {
+    link: 'string',        // `baf...x`
+    audience: 'string',   // `did:web:service`
+    issuer: 'string',     // `did:key:agent`
+    expiration: 'string', // `9256939505` (unix timestamp)
```
Gozala, line 36: Should this be a number instead ?

Gozala: Note UCANs use seconds precision so we may want to capture that detail here.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
   // space + root must be unique to satisfy index constraint
   primaryIndex: { partitionKey: 'space', sortKey: 'root' },
 }
+
+/** @type TableProps */
+export const delegationTableProps = {
+  fields: {
```
Gozala, line 32: I wonder if we would want to index by capability as well in the future.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
...
+    link: 'string',        // `baf...x`
+    audience: 'string',   // `did:web:service`
+    issuer: 'string',     // `did:key:agent`
+    expiration: 'string', // `9256939505` (unix timestamp)
+    insertedAt: 'string', // `2022-12-24T...`
+    updatedAt: 'string',  // `2022-12-24T...`
```
Gozala, line 38: I think we need `cause` field for the invocation CID here as well, especially if we'll be looking things up from the UCAN log.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
...
+  primaryIndex: { partitionKey: 'audience', sortKey: 'link' },
+}
+
+/** @type TableProps */
+export const subscriptionTableProps = {
+  fields: {
+    cause: 'string',        // `baf...x` (CID of invocation that created this subscription)
+    provider: 'string',   // `did:web:service` (DID of the provider, e.g. a storage provider)
+    customer: 'string',   // `did:mailto:agent` (DID of the user account)
+    subscription: 'string',      // string (arbitrary string associated with this subscription)
```
Gozala, line 51: Mentioned it elsewhere, but I think we need two fields instead of one here:

1. `order` that MUST be unique per `provider`.
2. `id` (or `subscription`) which is derived from `order` & `provider` that is globally unique.

This would allow providers to enforce whatever limits they want and allow us to store records across all providers without collision risks.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
...
+    insertedAt: 'string', // `2022-12-24T...`
+    updatedAt: 'string',  // `2022-12-24T...`
+  },
+  // TODO does this index setup seem right?
+  primaryIndex: { partitionKey: 'customer', sortKey: 'subscription' },
```
Gozala, line 56: Think so assuming we do the above.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
...
+  globalIndexes: {
+    // TODO: I don't think we should keep this index - partitioning by provider is basically useless and won't be much faster than a table scan
+    provider: { partitionKey: 'provider', projection: ['customer'] },
```
Gozala, line 59: I'm no dynamo expert, we should ask @mikeal for the input. However I would expect that providers may want to have a view over their customers with a good enough query ability.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
...
+    subscription: { partitionKey: 'subscription', projection: ['customer'] }
```
Gozala, line 60: I would expect `subscription` to be unique, primary key. Not sure what that would manifest as in dynamo.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
...
+  }
+}
+
+/** @type TableProps */
+export const consumerTableProps = {
+  fields: {
+    cause: 'string',        // `baf...x` (CID of invocation that created this consumer record)
+    consumer: 'string',   // `did:key:space` (DID of the actor that is consuming the provider, e.g. a space DID)
+    provider: 'string',   // `did:web:service` (DID of the provider, e.g. a storage provider)
+    subscription: 'string',      // string (arbitrary string associated with this subscription)
```
Gozala, line 71: Assuming we integrate `order` here I would expect this field to be `order` as opposed to `subsscription` since later could be derived from `(provider, order)` tuple.

upload-api/tables/index.js
```diff
@@ -26,3 +26,55 @@ export const uploadTableProps = {
...
+    insertedAt: 'string', // `2022-12-24T...`
+    updatedAt: 'string',  // `2022-12-24T...`
+  },
+  // TODO does this index setup seem right?
+  primaryIndex: { partitionKey: 'consumer', sortKey: 'subscription' },
```
Gozala, line 76: I don't think so. I think we want to query what providers does space (consumer in this context) has, or whether consumer has specific provider. `subscription` as sort key seems useless for such queries.

Gozala (changes requested):
Overall looks great. I have provided bunch of improvement suggestions. I think only only blockers are using non-local deps so all the CI checks pass.

It also may be worth merging linked PRs into this so we don't have version that stores delegations and version that indexes invocations, but I'm leaving it up to you to decide.

Gozala (changes requested):
Overall looks great. I have provided bunch of improvement suggestions. I think only only blockers are using non-local deps so all the CI checks pass.

It also may be worth merging linked PRs into this so we don't have version that stores delegations and version that indexes invocations, but I'm leaving it up to you to decide.

package.json
```diff
@@ -53,10 +53,13 @@
...
     "extends": [
-      "./node_modules/hd-scripts/eslint/index.js"
+      "./node_modules/hd-scripts/eslint/preact.js"
     ],
     "parserOptions": {
-      "project": "./tsconfig.json"
+      "project": "./tsconfig.json",
+      "ecmaFeatures": {
+        "jsx": true
+      }
```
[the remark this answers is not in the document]
travis: We do as currently formulated for the email validation flow. We could (and should) move that out to its own app - my preference would be to move it to w3console but I'd like to do that after we extract w3console out of the `w3up` repository.

upload-api/types.ts
```diff
@@ -44,6 +45,54 @@ export interface WorkflowBucket {
...
 
+export interface DelegationsBucket {
+  put: (cid: CID, bytes: ByteView<Delegation>) => Promise<void>
+  get: (cid: CID) => Promise<ByteView<Delegation>|undefined>
+}
+
+export interface SubscriptionInput {
+  customer: DID,
+  provider: DID,
+  subscription: string,
```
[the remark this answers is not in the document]
travis: it is yep! `order` was a reserved token in Dynamo so we couldn't call it that all the way down, and tbh I find it very confusing - I kept thinking it had something to do with sorting (though maybe I'm missing some aspect of it related to sorting?)

upload-api/types.ts
```diff
@@ -44,6 +45,54 @@ export interface WorkflowBucket {
...
+}
+
+export interface SubscriptionTable {
+  insert: (consumer: SubscriptionInput) => Promise<Subscription>
+  count: () => Promise<bigint>
+}
+
+export interface UnstableSubscriptionTable extends SubscriptionTable {
+  findCustomersByProvider: (provider: DID) => Promise<DID[]>
+  findCustomerForSubscription: (subscription: string) => Promise<DID | undefined>
```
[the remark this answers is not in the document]
travis: ah right - so these "unstable" interface functions are mostly here as sketches of future more multi-use functions - I agree with you that we should figure out what fields we want to return in all cases and call it `findSubscription` when we move it into the stable interface

upload-api/types.ts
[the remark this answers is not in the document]
travis: definitely - thanks for prompting, will do a big documentation and refactoring pass

upload-api/stores/provisions.js
```diff
@@ -0,0 +1,155 @@
...
+        provider: item.provider,
+        customer: item.customer,
+      }
+      try {
+        await dynamoDb.send(new PutItemCommand({
+          TableName: subscriptionsTableName,
+          Item: marshall(row),
+          ConditionExpression: `attribute_not_exists(consumer) OR ((cid = :cid) AND (consumer = :consumer) AND (provider = :provider) AND (customer = :customer))`,
+          ExpressionAttributeValues: {
+            ':cid': { 'S': row.cid },
```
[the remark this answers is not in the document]
travis: oh I think this is just a mistake - it should be `cause`

travis:
thanks @Gozala! generally agree with all the feedback and will get it integrated tomorrow.

I do think we need to chat about the indexes - I believe that the current table/index setup actually does enforce the constraints you want without creating composite keys - I'm not a huge fan of composite keys because they constrain some queries - ie, if I create a composite key with `provider` and `order` then a user can't query on that key unless they know the provider and order. I think that in many cases using (eg) `provider` as the partition key and `order` as the sort key provides many of the same benefits without constraining queriers in the way I've described, but it's totally possible I'm missing something that makes composite keys necessary.

I'm going to spend some time creating a HackMD or Notion to organize this discussion and will find a time to chat!

upload-api/functions/validate-email.js
```diff
@@ -0,0 +1,107 @@
...
+  if (!request.queryStringParameters?.ucan) {
+    return new HtmlResponse(
+      <ValidateEmailError msg={'Missing delegation in the URL.'} />
+    )
+  }
+
+  return new HtmlResponse(<PendingValidateEmail autoApprove={true} />)
+}
+
+export const preValidateEmail = Sentry.AWSLambda.wrapHandler(validateEmailGet)
```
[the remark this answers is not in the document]
travis: yea I agree with these - I'm going to defer as I'm hoping to move these out of this repo entirely soon, but agree with your nits!

upload-api/html.js
```diff
@@ -0,0 +1,215 @@
...
+   * @param {import('preact').VNode<{}>} body
+   * @param {ResponseInit} [init]
+   */
+  constructor(body, init = {}) {
+    const headers = {
+      headers: {
+        'content-type': 'text/html; charset=utf-8',
+      },
+    }
+    super(buildDocument(render(body)), { ...init, ...headers })
```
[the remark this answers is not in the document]
travis: yep agree - if we kept these here in the medium-term I think it might be worth moving in this direction, but this is hopefully a short-term bit of ugliness.

upload-api/types.ts
```diff
@@ -44,6 +45,54 @@ export interface WorkflowBucket {
...
+}
+
+export interface Consumer {
+
+}
+
+export interface ConsumerTable {
+  insert: (consumer: ConsumerInput) => Promise<Consumer>
+  count: () => Promise<bigint>
+  hasStorageProvider: (consumer: DID) => Promise<boolean>
```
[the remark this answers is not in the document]
travis: yep agree - I'm going to keep as-is for now but would like to think about reworking this post-merge

upload-api/stores/provisions.js
```diff
@@ -0,0 +1,155 @@
...
+        } else {
+          throw error
+        }
+      }
+    },
+
+    /**
+     * get number of stored items
+     */
+    count: async () => {
```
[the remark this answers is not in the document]
travis: yea I agree - I suspect we should get rid of this function but want to push that out post-merge

upload-api/tables/consumer.js
```diff
@@ -0,0 +1,158 @@
...
+    insert: async ({ consumer, provider, subscription, cause }) => {
+      const insertedAt = new Date().toISOString()
+
+      const row = {
+        consumer,
+        provider,
+        subscription,
+        cause: cause.toString(),
+        insertedAt,
+      }
```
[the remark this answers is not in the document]
travis: ah right - currently the invariant is maintained on both the subscription and consumer tables due to `subscription` being used as the `sortKey` for both, but I think you're right that we should probably create a key from `provider` and `subscription`

I'm going to write up a document summarizing these new tables and listing the queries we need to make on them - should be easier to chat and figure out which changes to make there

upload-api/tables/index.js
travis: yep! that made it in in the PR that moved to using delegations from the invocations they came in

upload-api/tables/index.js
travis: so I think I've managed to do this without introducing a new field - both tables are now indexed by `subscription` (primary key) and `provider` ("sort" key) which means that `subscription` will always be unique per-provider (since there will only ever be a single record for a given (`provider`, `subscription`)

I think this accomplishes what you're looking for without adding any user-space composite keys (as opposed to Dynamo composite keys, which is what we're creating with `subscription` and `provider`)

upload-api/tables/index.js
travis: yea agree - I'm just not sure that maintaining an index makes sense since we'll practically be doing a table scan every time... I think this is a Dynamo best-practices kind of question so yea would definitely love to get @mikeal's brain on this if we can!

travis:
OK @vasco-santos @Gozala I've put together a companion HackMD with a summary of the tables I have here

https://hackmd.io/pQutmD2rR9KeJ6hWzt1QQQ

Note that I've reworked them since you two last looked, bringing them into alignment with @Gozala's spec here:

https://github.com/web3-storage/w3up/pull/746

I'd also love to get @mikeal's eyes on this if he has any space for it!

upload-api/email.js
```diff
@@ -1,3 +1,5 @@
+import fetch from '@web-std/fetch'
```
vasco-santos, line 1: wondering if we should just force node18? thoughts?

travis: I'm down! I don't know how big a change this is likely to be tho - will we need to use 18 for everything in the system or can we just change for one stack?

travis: I'm down! I don't know how big a change this is likely to be tho - will we need to use 18 for everything in the system or can we just change for one stack?

vasco-santos: for everything. it hopefully is just change version in https://github.com/web3-storage/w3infra/blob/main/stacks/index.js#L21 + CI (even in seed we are already using https://github.com/web3-storage/w3infra/blob/main/seed.yml)

stacks/upload-api-stack.js
```diff
@@ -50,6 +61,10 @@ export function UploadApiStack({ stack, app }) {
           STORE_TABLE_NAME: storeTable.tableName,
           STORE_BUCKET_NAME: carparkBucket.bucketName,
           UPLOAD_TABLE_NAME: uploadTable.tableName,
+          CONSUMER_TABLE_NAME: consumerTable.tableName,
+          SUBSCRIPTION_TABLE_NAME: subscriptionTable.tableName,
+          DELEGATION_TABLE_NAME: delegationTable.tableName,
+          DELEGATION_BUCKET_NAME: delegationBucket.bucketName,
```
vasco-santos, line 67:
```suggestion
          DELEGATION_BUCKET_NAME: delegationBucket.bucketName,
          POSTMARK_TOKEN: process.env. POSTMARK_TOKEN ?? '',
```

upload-api/functions/validate-email.jsx
```diff
@@ -0,0 +1,133 @@
...
+    AWS_REGION = '',
+    DELEGATION_TABLE_NAME = '',
+    R2_DELEGATION_BUCKET_ENDPOINT = '',
+    R2_DELEGATION_BUCKET_ACCESS_KEY_ID = '',
+    R2_DELEGATION_BUCKET_SECRET_ACCESS_KEY = '',
+    R2_DELEGATION_BUCKET_NAME = '',
+    INVOCATION_BUCKET_NAME = '',
+    WORKFLOW_BUCKET_NAME = '',
+    POSTMARK_TOKEN = '',
+    PRIVATE_KEY = '',
```
vasco-santos, line 69: PRIVATE_KEY is a secret and not env. SST has something for it :) See https://github.com/web3-storage/w3infra/blob/main/upload-api/functions/ucan-invocation-router.js#L96

stacks/upload-api-stack.js
```diff
@@ -39,6 +48,8 @@ export function UploadApiStack({ stack, app }) {
         permissions: [
           storeTable,
           uploadTable,
+          delegationTable,
+          delegationBucket,
```
vasco-santos, line 52:
```suggestion
          delegationTable,
          delegationBucket,
          consumerTable,
          subscriptionTable,
```

upload-api/email.js
travis: ah shoot - it looks like this will require updating SST. I'd like to avoid that side quest for now, so will stick with 16 I think.

travis: here's the error I was seeing, fwiw: 

Screenshot 2023-06-20 at 3 12 39 PM

upload-api/stores/provisions.js
```diff
@@ -0,0 +1,155 @@
...
+  return {
+    services,
+    hasStorageProvider: async (consumer) => (
+      { ok: await consumerTable.hasStorageProvider(consumer) }
+    ),
+
+    put: async (item) => {
+      const { cause, consumer, customer, provider } = item
+      // by setting subscription to customer we make it so each customer can have at most one subscription
+      // TODO is this what we want?
```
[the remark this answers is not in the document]
travis: ok - I've updated this to use the CID of `customer`-wrapped-in-a-CBOR-map as the "partition key" and the provider as the  "sort key" (see https://github.com/web3-storage/w3infra/pull/194/files#diff-7f8284bee6fcc2ea7426b91d587ca5eff07fb34f5049b2e95823a03c2768b722) which will give us the uniqueness semantics you describe

re: inserting into both tables I agree, and I think the right path to get there is to break `provider/add` into the two invocations you've talked about (I think `provider/get` and `consumer/add` ?) - I'd like to do that in a separate initiative though since this has gotten quite big and I don't think the failure cases here are too bad to handle for now.

vasco-santos (changes requested):
@travis left new round of feedback. We are close, most comments relate to prepare ourselves better for the upcoming follow ups from this PR that were identified

.env.tpl
```diff
@@ -26,3 +26,7 @@ R2_SATNAV_BUCKET_NAME = ''
 R2_SECRET_ACCESS_KEY = ''
 R2_UCAN_BUCKET_NAME = ''
 SATNAV_BUCKET_NAME = ''
+
+# Mailslurp
```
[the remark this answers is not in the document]
travis: ah yes, definitely!

upload-api/errors.js
```diff
@@ -47,6 +47,33 @@ export class NoCarFoundForGivenReceiptError extends HTTPError {
...
+    super(msg, 404)
+    this.name = 'NoDelegationFoundForGivenCid'
+    this.code = NoDelegationFoundForGivenCidError.CODE
+  }
+}
+NoDelegationFoundForGivenCidError.CODE = 'ERROR_DELEGATION_NOT_FOUND_FOR_CID'
+
+export class FailedToDecodeDelegationForGivenCidError extends HTTPError {
+  constructor (msg = 'Failed to decode delegation with given CID') {
+    super(msg, 404)
```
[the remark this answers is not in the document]
travis: yep definitely, great catch

upload-api/functions/ucan-invocation-router.js
```diff
@@ -122,8 +148,15 @@ export async function ucanInvocationRouter(request) {
     access: createAccessClient(
       serviceSigner,
       DID.parse(accessServiceDID),
-      new URL(accessServiceURL)
+      provisionsStorage,
+      delegationsStorage
     ),
+    signer: serviceSigner,
+    // TODO: we should set URL from a different env var, doing this for now to avoid that refactor
```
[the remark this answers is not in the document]
travis: yes - done! https://github.com/web3-storage/w3infra/issues/209

upload-api/tables/delegations.js
```diff
@@ -0,0 +1,260 @@
...
+  FailedToDecodeDelegationForGivenCidError
+} from '../errors.js'
+
+
+// Feature flag for looking up delegations in the invocations in which
+// they were originally embeded.
+// As of 6/6/2023 not all delegations have corresponding invocations
+// because our access/confirm provider creates delegations.
+// Once we change this, we can re-enable this optimization.
+const FIND_DELEGATIONS_IN_INVOCATIONS = false
```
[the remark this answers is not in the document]
travis: done! https://github.com/web3-storage/w3infra/issues/210

upload-api/test/queries.test.js
```diff
@@ -0,0 +1,160 @@
...
+test('finding the customer responsible for a space', async (t) => {
+  const { subscriptions, customer } = await subscriptionsTestFixture(t.context)
+  const { consumers, consumer } = await consumersTestFixture(t.context)
+  const foundSubscriptions = await consumers.findSubscriptionsForConsumer(consumer)
+  const foundCustomer = await subscriptions.findCustomerForSubscription(foundSubscriptions[0])
+
+  t.is(foundCustomer, customer)
+})
+
+test.skip('provider canceling a subscription', async (t) => {
```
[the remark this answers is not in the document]
travis: yea I think that's a great idea - I'm going to move these tests and the "unstable" interfaces out to a separate branch for integration once we actually need these queries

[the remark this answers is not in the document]
travis: great idea - done! https://github.com/web3-storage/w3infra/pull/211

upload-api/tables/subscription.js
```diff
@@ -0,0 +1,166 @@
...
+        return {
+          subscription: i.subscription.toString()
+        }
+      }) : []
+    },
+
+    /**
+     * !!! the following methods are unstable and may be changed at any time !!!
+     * !!!      they are included here for testing and design purposes       !!!
+     */
```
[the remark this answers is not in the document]
travis: moved this out to a new branch: https://github.com/web3-storage/w3infra/pull/211

vasco-santos (commented):
This LGTM! Thanks for the great work here @travis 

As we talked, let's get the migration code here and tested in staging only next week. So that we don't have main branch blocked until then

tools/d1-migration/add-to-dynamo.js
```diff
@@ -0,0 +1,240 @@
...
+}
+
+async function loadFromD1 () {
+  const {
+    STAGE,
+  } = getEnv()
+
+  const dbName = (STAGE === 'prod') ? 'access' : 'access-staging'
+  try {
+    const delegations = JSON.parse(await exec(`wrangler d1 execute ${dbName} --command 'SELECT * from delegations_v3' --json`))[0].results
```
Gozala, line 32: Code comment on why we need `[0].results` would be very welcome.

upload-api/tables/subscription.js
```diff
@@ -0,0 +1,111 @@
...
+
+  return useSubscriptionTable(dynamoDb, tableName)
+}
+
+/**
+ * @param {DynamoDBClient} dynamoDb
+ * @param {string} tableName
+ * @returns {SubscriptionTable}
+ */
+export function useSubscriptionTable (dynamoDb, tableName) {
```
Gozala, line 48: I do not want to block this PR, but I still dislike amount of incidental complexity we have for every single table which I have described a solution for in another PR, that I think applies here just as well

https://github.com/web3-storage/w3filecoin/pull/26#discussion_r1253621904

Perhaps we could remove some of this complexity later as a cleanup.

Gozala (approved):

tools/d1-migration/add-to-dynamo.js
travis: done!

upload-api/tables/subscription.js
travis: I hear you on the incidental complexity, but to be honest this boilerplate doesn't bother me much - having one initializer per table that takes a generic client and another that takes a set of options to set up a client in a default way feels pretty normal and good to me.

That said I'm definitely not opposed to putting together a cleaner/easier-to-use abstraction for this. From my read of the comment thread you linked, @vasco-santos identified some additional complexity that isn't accounted for by the solution you propose. 

If you're up for it, I think it would be a good idea for you to do a quick spike (including integrating it into tests and other places where we use these initializers - migration scripts, etc) of what you're thinking on a single table and then have the three of us do a (high bandwidth maybe?) review of the solution to make sure it fits all our needs and is strictly better than what we have right now. I'd be more than happy to take that spike and run with it, updating our usage across the board. I'd also be down to do the spike, but would definitely need to sync with you and Vasco to summarize and dig into the thinking on this thread so far.

All that said, I appreciate you looking out for these abstractions! I'm adding yet another table so I do think investing in this will help us keep things readable and usable.

PR #195, opened 2023-05-08
fix: add backwards compatible client

Gozala:


travis (approved):

vasco-santos:
@Gozala can we get this in?

PR #196, opened 2023-05-08
fix: use transport sniffing on incoming requests

Gozala:

PR #197, opened 2023-05-08
chore: avoid transcode unless necessary

Gozala:
Cleans up #196 and avoids redundant decode+encode when request is already desired format


vasco-santos (approved):
LGTM

travis (approved):
lgtm

vasco-santos:
@Gozala can we get this in?

PR #198, opened 2023-05-12
fix: disable point in time recovery for dynamodb tables

vasco-santos:
This PR removes point in time recovery for dynamodb tables here per https://github.com/web3-storage/secrets/issues/5

Once this releases, we should check https://us-west-2.console.aws.amazon.com/dynamodbv2/home?region=us-west-2#table?name=prod-w3infra-store&tab=backups to guarantee cloudformation does it. Otherwise, then we can disable it manually in the console

### DIFF
--- a/ucan-invocation/tables/index.js
+++ b/ucan-invocation/tables/index.js
@@ -8,6 +8,11 @@ export const adminMetricsTableProps = {
   },
   // name must be unique to satisfy index constraint
   primaryIndex: { partitionKey: 'name' },
+  cdk: {
+    table: {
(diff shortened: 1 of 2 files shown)


vasco-santos:
Note: We decided to not have this for now in w3infra prod given its low cost at the moment. We can however consider make this enabled for PRs + staging

travis (approved):

vasco-santos:
note: we should still change this PR to set this for non production env

travis:
@alanshaw  closing this for now to clean up PR environments but happy to reopen later

PR #200, opened 2023-05-18
feat: implement `ProvisionStorage` in terms of subscription and consumer tables

travis:
Implement the subscription and consumer tables specified by @Gozala in https://github.com/web3-storage/w3up/pull/746 and use them to implement a new version of ProvisionStorage

This builds on a PR review from @Gozala that lays out some of the queries we will need to support with these new tables:

https://github.com/web3-storage/w3infra/pull/194#pullrequestreview-1417122161

This is configured to merge into the main "D1 to Dynamo" development branch - I'm in favor of going to production with this implementation of `ProvisionStorage` rather than the one in that branch now.


travis:
There seems to be general agreement about moving in this direction so I'm going to merge this into the main PR for review simplicity.

PR #201, opened 2023-05-24
feat: stop storing delegations in a bespoke bucket

travis:
Since we already store delegations inside invocations in the invocation log, we can simply store the invocation CID and delegation CID and then fetch and find the needed delegations when they are requested.

I'm not sure the implementation is correct, as I'm not super familiar with the distinction between invocations, workflows and receipts, but the types all check out!

Would love careful review on the new `findDelegationInInvocation` method I've added to `upload-api/tables/delegations.js`

### DIFF
--- a/upload-api/buckets/delegations-store.js
+++ b/upload-api/buckets/delegations-store.js
@@ -24,6 +24,25 @@ export function createDelegationsStore(region, bucketName, options = {}) {
 }
 
 /**
+ * 
+ * @param {string} endpoint 
+ * @param {string} accessKeyId 
+ * @param {string} secretAccessKey 
+ * @param {string} bucketName
+ */
+export function useR2DelegationsStore(endpoint, accessKeyId, secretAccessKey, bucketName){
+  const s3Client = new S3Client({
+    region: 'auto',
+    endpoint,
+    credentials: {
+      accessKeyId,
+      secretAccessKey,
+    },
+  })
+  return useDelegationsStore(s3Client, bucketName)
+}
+
+/**
  * @param { CID} cid
  */
 function createDelegationsBucketKey (cid) {
--- a/upload-api/buckets/delegations-store.js
+++ b/upload-api/buckets/delegations-store.js
@@ -10,27 +10,13 @@ import { base32 } from 'multiformats/bases/base32'
 
 /**
  * Abstraction layer with Factory to perform operations on bucket.
- *
- * @param {string} region
- * @param {string} bucketName
- * @param {import('@aws-sdk/client-s3').ServiceInputTypes} [options]
- */
-export function createR2DelegationsStore(region, bucketName, options = {}) {
-  const s3client = new S3Client({
-    region,
-    ...options,
-  })
-  return useDelegationsStore(s3client, bucketName)
-}
-
-/**
  * 
  * @param {string} endpoint 
  * @param {string} accessKeyId 
  * @param {string} secretAccessKey 
  * @param {string} bucketName
  */
-export function useR2DelegationsStore(endpoint, accessKeyId, secretAccessKey, bucketName){
+export function createR2DelegationsStore(endpoint, accessKeyId, secretAccessKey, bucketName){
   const s3Client = new S3Client({
     region: 'auto',
     endpoint,
--- a/upload-api/buckets/delegations-store.js
+++ b/upload-api/buckets/delegations-store.js
@@ -15,7 +15,7 @@ import { base32 } from 'multiformats/bases/base32'
  * @param {string} bucketName
  * @param {import('@aws-sdk/client-s3').ServiceInputTypes} [options]
  */
-export function createDelegationsStore(region, bucketName, options = {}) {
+export function createR2DelegationsStore(region, bucketName, options = {}) {
   const s3client = new S3Client({
     region,
     ...options,
--- a/upload-api/functions/ucan-invocation-router.js
+++ b/upload-api/functions/ucan-invocation-router.js
@@ -150,7 +150,7 @@ export async function ucanInvocationRouter (request) {
       /** @type {import('@web3-storage/upload-api').ServiceDID} */
       (accessServiceDID)
     ]),
-    delegationsStorage: createDelegationsTable(AWS_REGION, delegationTableName, delegationBucket)
+    delegationsStorage: createDelegationsTable(AWS_REGION, delegationTableName, delegationBucket, invocationBucket, workflowBucket)
   })
 
   const processingCtx = {
--- a/upload-api/functions/ucan-invocation-router.js
+++ b/upload-api/functions/ucan-invocation-router.js
@@ -19,7 +19,7 @@ import { CAR, Legacy, Codec } from '@ucanto/transport'
 import { Email } from '../email.js'
 import { useProvisionStore } from '../stores/provisions.js'
 import { createDelegationsTable } from '../tables/delegations.js'
-import { createDelegationsStore } from '../buckets/delegations-store.js'
+import { useR2DelegationsStore } from '../buckets/delegations-store.js'
 import { createSubscriptionTable } from '../tables/subscription.js'
 import { createConsumerTable } from '../tables/consumer.js'
 
@@ -86,6 +86,9 @@ export async function ucanInvocationRouter (request) {
     SUBSCRIPTION_TABLE_NAME: subscriptionTableName = '',
     DELEGATION_TABLE_NAME: delegationTableName = '',
     DELEGATION_BUCKET_NAME: delegationBucketName = '',
+    R2_DELEGATION_BUCKET_ENDPOINT: r2DelegationBucketEndpoint = '',
+    R2_DELEGATION_BUCKET_ACCESS_KEY_ID: r2DelegationBucketAccessKeyId = '',
+    R2_DELEGATION_BUCKET_SECRET_ACCESS_KEY: r2DelegationBucketSecretAccessKey = '',
     INVOCATION_BUCKET_NAME: invocationBucketName = '',
     TASK_BUCKET_NAME: taskBucketName = '',
     WORKFLOW_BUCKET_NAME: workflowBucketName = '',
@@ -113,7 +116,7 @@ export async function ucanInvocationRouter (request) {
   )
   const taskBucket = createTaskStore(AWS_REGION, taskBucketName)
   const workflowBucket = createWorkflowStore(AWS_REGION, workflowBucketName)
-  const delegationBucket = createDelegationsStore(AWS_REGION, delegationBucketName)
+  const delegationBucket = useR2DelegationsStore(r2DelegationBucketEndpoint, r2DelegationBucketAccessKeyId, r2DelegationBucketSecretAccessKey, delegationBucketName)
   const subscriptionTable = createSubscriptionTable(AWS_REGION, subscriptionTableName, {
     endpoint: dbEndpoint
   });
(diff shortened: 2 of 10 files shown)


upload-api/buckets/delegations-store.js
```diff
@@ -23,6 +23,25 @@ export function createDelegationsStore(region, bucketName, options = {}) {
...
 }
 
+/**
+ * 
+ * @param {string} endpoint 
+ * @param {string} accessKeyId 
+ * @param {string} secretAccessKey 
+ * @param {string} bucketName
+ */
+export function useR2DelegationsStore(endpoint, accessKeyId, secretAccessKey, bucketName){
```
vasco-santos, line 33:
```suggestion
export function createR2DelegationsStore(endpoint, accessKeyId, secretAccessKey, bucketName){
```

upload-api/functions/ucan-invocation-router.js
```diff
@@ -86,6 +86,9 @@ export async function ucanInvocationRouter (request) {
     SUBSCRIPTION_TABLE_NAME: subscriptionTableName = '',
     DELEGATION_TABLE_NAME: delegationTableName = '',
     DELEGATION_BUCKET_NAME: delegationBucketName = '',
+    R2_DELEGATION_BUCKET_ENDPOINT: r2DelegationBucketEndpoint = '',
+    R2_DELEGATION_BUCKET_ACCESS_KEY_ID: r2DelegationBucketAccessKeyId = '',
+    R2_DELEGATION_BUCKET_SECRET_ACCESS_KEY: r2DelegationBucketSecretAccessKey = '',
```
vasco-santos, line 91: We should also have the name, like we do for replicator https://github.com/web3-storage/w3infra/#r2_carpark_bucket_name (even if it currently the same)

upload-api/tables/delegations.js
```diff
@@ -134,27 +151,54 @@ async function cidToDelegation (bucket, cid) {
...
 
-/**
- * 
- * @param {import('../types').DelegationsBucket} bucket
- * @param {Iterable<readonly [key: CID, value: Uint8Array ]>} entries
- */
-async function writeEntries (bucket, entries) {
-  await Promise.all([...entries].map(([key, value]) => bucket.put(key, value)))
+  if (!agentMessageWithInvocationCid) {
+    throw new NoCarFoundForGivenReceiptError()
```
vasco-santos, line 172: Does this trigger on a receipt? I think we need a different error. This error in `ucan-invocation` is because the flow is:
1. process received CAR file with receipt
2. get CAR file from invocation that corresponds do the receipt, so that we have the execution parameters of the invocation that generated the receipt.

Therefore, here error should be `NoCarFoundForGivenDelegation` or similar.

upload-api/tables/delegations.js
```diff
@@ -134,27 +151,54 @@ async function cidToDelegation (bucket, cid) {
...
+  }
+
+  const agentMessageBytes = await workflowBucket.get(
+    agentMessageWithInvocationCid
+  )
+  if (!agentMessageBytes) {
+    throw new NoCarFoundForGivenReceiptError()
```
vasco-santos, line 179: same as above

upload-api/tables/delegations.js
```diff
@@ -134,27 +151,54 @@ async function cidToDelegation (bucket, cid) {
...
+  }
+
+  const proofDelegations = invocation.proofs.flatMap((proof) =>
+    'capabilities' in proof ? [proof] : []
+  )
+  const foundDelegation = proofDelegations.find(proof => proof.cid)
+
+  if (!foundDelegation) {
+    throw new NoInvocationFoundForGivenReceiptError()
+  }
```
vasco-santos, line 201: this is the part I am less familiar with. The rest of the flow to get the CAR file where delegation is looks right

upload-api/tables/index.js
```diff
@@ -91,6 +74,7 @@ export const consumerTableProps = {
   // TODO does this index setup seem right?
   primaryIndex: { partitionKey: 'consumer', sortKey: 'subscription' },
   globalIndexes: {
+    // TODO: I don't think we should keep this index - partitioning by provider is basically useless and won't be much faster than a table scan
```
vasco-santos, line 77: could we get a small markdown doc with all tables schemas and queries we need to do so that we can align better on which indexes we should use?

upload-api/test/service/delegations.test.js
```diff
@@ -21,12 +23,17 @@ test.before(async (t) => {
...
+
     
     return useDelegationsTable(
       dynamo,
       await createTable(dynamo, delegationTableProps),
-      useDelegationsStore(s3, bucketName)
+      useDelegationsStore(s3, delegationsBucketName),
+      useInvocationStore(s3, invocationsBucketName),
+      useWorkflowStore(s3, workflowBucketName)
     )
```
vasco-santos, line 37: I think that we should be able to test this flow. Would expect something like this https://github.com/web3-storage/w3infra/blob/main/upload-api/test/ucan-invocation.test.js#L140-L157 (using access api invocations for the delegations, meaning it would be stored in the invocations bucket and workflow bucket)

vasco-santos:
Thanks for putting this together @travis

upload-api/tables/index.js
travis: so what I've been working toward is a test that demonstrates the various queries we need to do - I've implemented some of them here but need to do a bit more prototyping/research to be more certain about the set of queries we need to do:

https://github.com/web3-storage/w3infra/pull/194/files#diff-924a94ba76534a93c7fa4c89f8e70037f21694b6374beaca7a2c6ebf3fdb2006

does this give you the info you're looking for? I'd rather work in code at the moment since this is changing fairly quickly but can definitely compile info from this file and the queries tests into one doc if that would help a lot!

upload-api/buckets/delegations-store.js
travis: hm, I'd like to follow the convention used by the other modules in `/buckets` where the `use*` function takes an endpoint, accessKeyId, etc and the `create*` function takes an s3 client. I've updated the name of the other function in this module to match this pattern, but left this one as-is.

upload-api/tables/delegations.js
travis: Done! Introduced a couple new errors that feel like they better reflect what's going on. I could break down the `NoInvocationFoundForGivenCidError` error into a couple more precise errors about different parts of the invocation lookup process, but wanted to start simple.

upload-api/tables/delegations.js
```diff
@@ -134,27 +151,54 @@ async function cidToDelegation (bucket, cid) {
...
+
+  const agentMessageBytes = await workflowBucket.get(
+    agentMessageWithInvocationCid
+  )
+  if (!agentMessageBytes) {
+    throw new NoInvocationFoundForGivenCidError()
+  }
+
+  const agentMessage = await CAR.request.decode({
```
Gozala, line 182: This works, however not exactly how I'd expect one to use it. I think doing [`CAR.decode`](https://github.com/web3-storage/ucanto/blob/d6f1f7b8e8d97db8515a8b32d4b687102580418d/packages/core/src/car.js#L97-L119) first and then [`Message.view`](https://github.com/web3-storage/ucanto/blob/main/packages/core/src/message.js#L32-L40) would be expected use of APIs (where `store` is `blocks` returned by `CAR.decode`).

That is more or less what [`CAR.request.decode` does](https://github.com/web3-storage/ucanto/blob/main/packages/transport/src/car/request.js#L45-L56), although it may change in the future, which is why I don't think it should be used here.

Gozala: Taking closer look this seems unnecessarily complicated way to go about this. It could be simplified by:

1. Doing `const { blocks } = CAR.decode(bytes)` (where CAR is from `core` as opposed to transport).
2. Calling `const delegation = Delegation.view({ blocks, root: delegationCID })`

It is also going to be completely immune to `AgentMessage` format changes, because it simply creates a cursor into CAR blocks.

⚠️ Also once you create a delegation in 2nd step I would also access `delegation.data` to force delegation to be materialized. This is unfortunate side effect of view been lazy, which I'll address someday.

upload-api/tables/delegations.js
```diff
@@ -94,8 +101,16 @@ export function useDelegationsTable (dynamoDb, tableName, bucket) {
...
-        const { link } = unmarshall(result)
-        delegations.push(await cidToDelegation(bucket, CID.parse(link)))
+        const { cause, link } = unmarshall(result)
+        const delegationCid = CID.parse(link)
+        if (cause) {
+          delegations.push(await findDelegationInInvocation(
+            invocationBucket, workflowBucket,
+            CID.parse(cause), delegationCid
+          ))
+        } else {
```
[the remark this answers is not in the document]
travis: ah that's an interesting idea - I'm going to keep this here for now but will explore the viability of that. 

@vasco-santos how hard would it be to backfill some invocations to avoid having to ever look at the legacy R2 bucket?

upload-api/tables/delegations.js
```diff
@@ -94,8 +101,16 @@ export function useDelegationsTable (dynamoDb, tableName, bucket) {
 
       const delegations = []
       for (const result of response.Items ?? []) {
-        const { link } = unmarshall(result)
-        delegations.push(await cidToDelegation(bucket, CID.parse(link)))
+        const { cause, link } = unmarshall(result)
+        const delegationCid = CID.parse(link)
+        if (cause) {
```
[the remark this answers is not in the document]
travis: that's right - will add a comment

upload-api/functions/validate-email.js
```diff
@@ -64,7 +74,7 @@ function createAuthorizeContext () {
     url: new URL(ACCESS_SERVICE_URL),
     email: new Email({ token: POSTMARK_TOKEN }),
     signer: getServiceSigner({ UPLOAD_API_DID, PRIVATE_KEY }),
-    delegationsStorage: createDelegationsTable(AWS_REGION, DELEGATION_TABLE_NAME, delegationBucket),
+    delegationsStorage: createDelegationsTable(AWS_REGION, DELEGATION_TABLE_NAME, delegationBucket, invocationBucket, workflowBucket),
```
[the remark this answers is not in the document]
travis: I'm sympathetic here, but all the other modules in this namespace follow a pattern of exporting a `create*Table` function that take `region` and `tableName` as positional arguments and I'd like to conform to that convention for now. I do think wrapping `bucket`, `invocationBucket` and `workflowBucket` into a single struct is a good idea so I'll do that for now...

upload-api/tables/delegations.js
travis: ok great - chatted in a huddle and got this looking good - much simpler, thanks!!

upload-api/tables/delegations.js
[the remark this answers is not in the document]
vasco-santos: @travis should not be difficult. We can use S3 client to read from R2 and write to S3. How are keys stored there? Only caution is to have a sha256/md5 to guarantee data integrity. If we can't get it, we need to buffer them in memory to copy (which should also be fine for the sizes they will likely have)

upload-api/buckets/delegations-store.js
vasco-santos: Here it is not the case right? Looking at above create + https://github.com/web3-storage/w3infra/blob/main/upload-api/buckets/car-store.js `use` receives an S3 client, while create creates one

travis: oh you're totally right, I was confused, thanks and good catch!

upload-api/test/service/delegations.test.js
travis: yep agree! I'm going to get this merged into the main PR and add a test there

upload-api/tables/delegations.js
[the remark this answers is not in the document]
travis: ok great - will plan on that. I'm going to leave this fallback code in for now but will remove once I have a working migration script that backfills invocations

travis:
have made requested changes so I'm going to merge this into the main PR - can address any additional changes there

PR #207, opened 2023-06-21
feat: roundabout supports redirect to raw key with bucket as query param

vasco-santos:
There is also requirement to be able to create presigned URLs for [dagcargo bucket](https://dash.cloudflare.com/fffa4b4363a7e5250af8357087263b3a/r2/default/buckets/dagcargo). It does not have same format as carpark.

This PR adds a `GET /raw/{key}` endpoint to the API Gateway under roundabout. It enables to pass a raw key of a bucket, rather than a CID that is converted to the carpark key format. By default, it still picks `carpark` bucket if no bucket is provided.

Integration tests were added so that we can actually test the deployment, so that env vars and permissions are validated

Alternatively, we could consider not to be opinated about key key format and always receive a raw key. However, I think it makes sense to move into a way that we create key for carpark. There are a few advantages, like in the future if we have more buckets, bucket sharding or so on, we can easily abstract that logic within this API. Moreover, once we land w3filecoin spade integration, and deprecate old APIs, there will be no need for this format and we can consider to deprecate this new endpoint.

There is also option to be more explicit here, and also change `GET /{cid}` to be `GET /car/{cid}`, which would even allow in the future to get blocks for example, while being more inline with also getting a raw request.

### DIFF
--- a/docs/roundabout.md
+++ b/docs/roundabout.md
@@ -1,6 +1,6 @@
 # Roundabout
 
-> Roundabout allows the creation of short lived presigned URLs for content stored in carpark bucket (in R2). It can also get the HTTP request redirected to the given presigned URL.
+> Roundabout allows the creation of short lived presigned URLs for content stored in web3.storage buckets (in R2). It can also get the HTTP request redirected to the given presigned URL.
 
 ## HTTP API
 
@@ -8,7 +8,13 @@ The given API is currently public.
 
 ### `GET /{carCid}`
 
-Redirects to a URL where the requested CAR file (by its CID) can be downloaded from. The request will return a `302 Redirect` to a created presigned URL.
+Redirects to a presigned URL where the requested CAR file (by its CID) can be downloaded from. This will use web3.storage `carpark` as the location of the requested CARs. The request will return a `302 Redirect` to a created presigned URL.
+
+It also supports a query parameter `expires` with the number of seconds this presigned URL should be valid for. You can set a value from one second to 7 days (604,800 seconds). By default the expiration is set for 3 days (259,200 seconds).
+
+### `GET /raw/{key}?bucket=bucket-name`
+
+Redirects to a presigned URL where the requested bucket value can be downloaded from by its key. Unlike `GET /{carCid}`, this endpoint takes a raw key and is compatible with any web3.storage account bucket. The request will return a `302 Redirect` to a created presigned URL.
 
 It also supports a query parameter `expires` with the number of seconds this presigned URL should be valid for. You can set a value from one second to 7 days (604,800 seconds). By default the expiration is set for 3 days (259,200 seconds).
 
@@ -34,7 +40,7 @@ location: https://carpark-prod-0.fffa4b4363a7e5250af8357087263b3a.r2.cloudflares
 apigw-requestid: G3T3xg1LvHcEPxA=
 ```
 
-### Get presigned URL with custom expiration
+### Get presigned URL for CAR file with custom expiration via CURL
 
 ```console
 curl --head https://roundabout.web3.storage/bagbaiera222226db4v4oli5fldqghzgbv5rqv3n4ykyfxk7shfr42bfnqwua?expires=900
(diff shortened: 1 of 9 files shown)


vasco-santos:
Output on deployment env

```console
$ curl --head https://pr207.roundabout.web3.storage/raw/0000c19bd9cd7fa9c532eba81428eda0_baga6ea4seaqpohse35l4xucs5mtabgewpp4mgtle7yym7em6ouvhgjb7wc2pcmq.car\?bucket\=dagcargo
HTTP/2 302
date: Wed, 21 Jun 2023 13:26:56 GMT
content-length: 0
location: https://dagcargo.fffa4b4363a7e5250af8357087263b3a.r2.cloudflarestorage.com/0000c19bd9cd7fa9c532eba81428eda0_baga6ea4seaqpohse35l4xucs5mtabgewpp4mgtle7yym7em6ouvhgjb7wc2pcmq.car?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Credential=a314d2872c5c092e911e3e2c7b5e5c3f%2F20230621%2Fauto%2Fs3%2Faws4_request&X-Amz-Date=20230621T132656Z&X-Amz-Expires=259200&X-Amz-Signature=938d4ca6ed2e07a008abb18885bb894f7a7d0972a7068a9a9f3fed40e964f0c9&X-Amz-SignedHeaders=host&x-id=GetObject
apigw-requestid: G3wZCjdgPHcEJiw=

$ curl -L https://pr207.roundabout.web3.storage/raw/0000c19bd9cd7fa9c532eba81428eda0_baga6ea4seaqpohse35l4xucs5mtabgewpp4mgtle7yym7em6ouvhgjb7wc2pcmq.car\?bucket\=dagcargo --output 0000c19bd9cd7fa9c532eba81428eda0_baga6ea4seaqpohse35l4xucs5mtabgewpp4mgtle7yym7em6ouvhgjb7wc2pcmq.car
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0 30.9G    0 59.4M    0     0  20.2M      0  0:26:02  0:00:02  0:26:00 48.3M
```

docs/roundabout.md
```diff
@@ -1,14 +1,20 @@
...
 The given API is currently public.
 
 ### `GET /{carCid}`
 
-Redirects to a URL where the requested CAR file (by its CID) can be downloaded from. The request will return a `302 Redirect` to a created presigned URL.
+Redirects to a presigned URL where the requested CAR file (by its CID) can be downloaded from. This will use web3.storage `carpark` as the location of the requested CARs. The request will return a `302 Redirect` to a created presigned URL.
+
+It also supports a query parameter `expires` with the number of seconds this presigned URL should be valid for. You can set a value from one second to 7 days (604,800 seconds). By default the expiration is set for 3 days (259,200 seconds).
+
+### `GET /raw/{key}?bucket=bucket-name`
```
alanshaw, line 15: Given that you can use `GET /raw/{key}` to do the same thing as `GET /{carCid}` like `GET /raw/bagbaiera222226db4v4oli5fldqghzgbv5rqv3n4ykyfxk7shfr42bfnqwua/bagbaiera222226db4v4oli5fldqghzgbv5rqv3n4ykyfxk7shfr42bfnqwua.car` I'd be tempted to simplify and just have one endpoint `GET /{key}`.

stacks/roundabout-stack.js
```diff
@@ -34,6 +34,8 @@ export function RoundaboutStack({ stack, app }) {
     routes: {
       'GET /{cid}':       'functions/redirect.handler',
       'HEAD /{cid}':      'functions/redirect.handler',
+      'GET /raw/{key}':   'functions/redirect.rawHandler',
+      'HEAD /raw/{key}':   'functions/redirect.rawHandler',
```
alanshaw, line 38: Key has to be with URL encoded slashes right or this will not work...?

docs/roundabout.md
vasco-santos: Yeah, I wrote this in original post. My goal to not do this is to not delegate responsability to know key format to the client, because in future we can have multiple buckets (with sharding or other indexing ways). This `/raw/{key}` seems more like a temporary exception for the real goal of roundabout.

I am happy to change to single route if you think that makes more sense that my reasoning written in original post

stacks/roundabout-stack.js
vasco-santos: dagcargo has no `/` so it was fine. But if we do a single endpoint we need to take that into account yes

docs/roundabout.md
alanshaw: ACK, makes sense! Lets ship what's here but please lets restrict the bucket.

alanshaw (approved):

PR #211, opened 2023-06-27
feat: experimental queries and interfaces for subscription and consumer tables

travis:
Per @vasco-santos's suggestion in https://github.com/web3-storage/w3infra/pull/194#discussion_r1241899891 I've moved the experimental queries and interfaces to a separate PR to keep that PR cleaner.

Mostly here as a guide for posterity and future work - this can be closed if we don't use it within a few months.


travis:
closing to free up AWS seed.run resources (IAMs were running low)

PR #212, opened 2023-07-06
docs: fallback secret command changed

alanshaw:
The command for setting the fallback secret changed.


vasco-santos (commented):
Doesn't it depend on sst version? I seem to recall the new command for new projects, but thought it was from sst v2

alanshaw:
Ohhhh ok gotcha

PR #215, opened 2023-07-13
feat: add d1 -> dynamo migration verification script

travis:
grab data from D1 and use it to verify that all the bits got moved correctly

during the course of this I found a bug in the migration script - whew! thanks for the suggestion @vasco-santos

also add a simple utility to list the emails of folks who have registered for w3up


vasco-santos (approved):
From my knowledge of the D1 contents this look good. Could be good to have extra eyes from @Gozala

PR #216, opened 2023-07-28
feat: implement RateLimitStorage and integrate it into upload-api

travis:
Implement the new RateLimitStorage interface from https://github.com/web3-storage/w3up/pull/832 and integrate it into the upload-api.

This gives us the ability to block uploads to a space and to block an email or domain name from authorizing with our service.

Note that we leave a couple of the new capabilities unimplemented for now since we don't need them urgently. We are actively hoping to block some abusive users ASAP, so I'm getting the blocking part of this work in now and will then turn my attention to the remaining capabilities.

TODO
- [x] remove file dependencies from package.json and point at latest versions of deps once https://github.com/web3-storage/w3up/pull/832 is merged and released


olizilla (commented):
~~nit: there is a significant diff on the `package-lock.json` but no change to the `package.json`. plz revert if unintended~~

olizilla:
oh sorry, posted too quickly, the workspace modules have package.json diffs which cause the diff on the root level package.json

upload-api/tables/rate-limit.js
```diff
@@ -0,0 +1,75 @@
+import {
+  DeleteItemCommand,
+  DynamoDBClient, PutItemCommand, QueryCommand,
+} from '@aws-sdk/client-dynamodb'
+import { marshall, unmarshall } from '@aws-sdk/util-dynamodb'
+import { v4 as uuidv4 } from 'uuid'
```
olizilla, line 6: we're currently using `nanoid` in this project for id generation https://github.com/ai/nanoid

upload-api/tables/index.js
```diff
@@ -76,4 +76,20 @@ export const consumerTableProps = {
...
+    cause: 'string',        // `baf...x` (CID of invocation that created record)
+    subject: 'string',      // string (arbitrary string identifying the subject to be limited)
+    rate: 'number',         // unitless number representing the rate to which the subject is limited
+    insertedAt: 'string',   // `2022-12-24T...`
+    updatedAt: 'string',    // `2022-12-24T...`
+  },
+  primaryIndex: { partitionKey: 'id' },
+  globalIndexes: {
+    subject: { partitionKey: 'subject', projection: ['rate', 'id'] },
+  }
```
olizilla, line 94: if folks will mostly query by `subject` then consider modelling the table so that the subject is the partition (primary) key, so we can get back all rate limits for that subject without a secondary index. Something like

```suggestion
export const rateLimitTableProps = {
  fields: {
    id: 'string',           // arbitrary identifier for this limit
    cause: 'string',        // `baf...x` (CID of invocation that created record)
    subject: 'string',      // string (arbitrary string identifying the subject to be limited)
    rate: 'number',         // unitless number representing the rate to which the subject is limited
    insertedAt: 'string',   // `2022-12-24T...`
    updatedAt: 'string',    // `2022-12-24T...`
  },
  primaryIndex: { partitionKey: 'subject', sortKey: 'id' }
```
if the `id` includes a timestamp such that lexigraphical sorting preserves rough ordering, then great!

The general recommendation with dynamo is to abuse the heck out of the primary key in ways you'd never consider if you've ever modelled data in anything else. It's wild.

> well-optimized single-table DynamoDB layout looks more like machine code than a simple spreadsheet
- https://www.trek10.com/blog/dynamodb-single-table-relational-modeling

(we're not ready for single table design yet, so this is a more advanced walk through then is strictly relevant but i think i shows the point...)

olizilla (approved):
👍 nice work!

non-blocking comments are inline. I'd switch out the uuid dep for nanoid, and meditate on the dynamo table structure some more, but what you have will work, so is up for discussion.

test/integration.test.js
```diff
@@ -65,6 +70,48 @@ test('upload-api /metrics', async t => {
...
+    Item: marshall({
+      id: Math.random().toString(10),
+      subject: 'example2.com',
+      rate: 0
+    })
+  }))
+  
+  // it would be nice to use t.throwsAsync here, but that doesn't work with errors that aren't exceptions: https://github.com/avajs/ava/issues/2517
+  try {
+    await client.authorize('travis@example2.com')
```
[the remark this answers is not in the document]
travis: it does yea - I think this happens fairly frequently when we throw the `error` of a Ucanto `Result` - in this particular case it happens here:

https://github.com/web3-storage/w3up/blob/main/packages/access-client/src/agent-use-cases.js#L31

I think this is a pattern that @Gozala has advocated for, but I may have misunderstood and don't have a good handle on the reasons - fwiw I would be down to say that we should always wrap these `error` objects in a real JavaScript `Exception` before we `throw` them - does that sound right to you @olizilla and @Gozala ?

upload-api/tables/rate-limit.js
travis: oh awesome, will update!

upload-api/tables/index.js
travis: mmm, that's a good thought... given your comments in the spec repo and a general desire to get this out I'm going to stick with the current model, but I think this would be a good idea to consider when we iterate on this - the migration should be straightforward when we get there!

PR #217, opened 2023-08-11
fix: re-enable support for multiple providers

travis:
we lost this in the D1 migration work - go back to parsing the PROVIDERS env var and use the result as the list of supported providers

### DIFF
--- a/upload-api/config.js
+++ b/upload-api/config.js
@@ -9,11 +9,23 @@ import * as DID from '@ipld/dag-ucan/did'
  * @param {string} [config.UPLOAD_API_DID] - public DID for the upload service (did:key:... derived from PRIVATE_KEY if not set)
  * @returns {import('@ucanto/principal/ed25519').Signer.Signer}
  */
- export function getServiceSigner(config) {
+export function getServiceSigner(config) {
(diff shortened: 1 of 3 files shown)


upload-api/config.js
```diff
@@ -9,11 +9,29 @@ import * as DID from '@ipld/dag-ucan/did'
...
   return signer
 }
+
+/**
+ * Given a string, parse into provider ServiceDIDs. 
+ * 
+ * @param {string} providersEnvVar 
+ * @returns {import('@web3-storage/upload-api').ServiceDID[]}
+ */
+export function parseProviders(providersEnvVar) {
```
alanshaw, line 27: This doesn't really have anything to do with providers - it's parsing a CSV of did:web prefixed DIDs. I'd call it something to that effect.

alanshaw (approved):

upload-api/config.js
travis: Cool - renamed to parseServiceDids since that's what feels natural for the types. I think this could be further generalized to `parseDidWebs` or something but I do want to return `ServiceDIDs` in this case and have a ServiceDID-specific error message and don't feel like diving down the generics rabbithole right this sec, so feel pretty good about this level of abstraction for now.

PR #218, opened 2023-08-14
feat: index upload table by root CID

travis:
@dchoi27 needs to be able to figure out which space an upload went to, ie, he needs to be able to search space DIDs by upload CID

I'm not sure that we need `insertedAt` in here, but I do think it will be helpful to be able to see when a CID was uploaded to various spaces in the admin UI, so I'm including it in the projection. It would not be hard to convince me that we should leave it out...


olizilla (approved):
Having `insertedAt` gives us some potentially useful info. Go for it.

PR #220, opened 2023-08-18
feat: implement `consumer/get` and `subscription/get` invocation handlers

travis:
`consumer/get` required wiring up the space metrics table in the upload-api stack to get space allocation metrics, which we call "allocation" in w3admin. Currently just returning 1 billion as the "total" space for all consumers since we don't currently check anyway, but we should figure out where this lives when we tackle billing in the near future.

Also implement `getStorageProviders` per https://github.com/web3-storage/w3up/pull/862

Also add `subscriptions` to the `customer/get` response, per https://github.com/web3-storage/w3up/pull/863


upload-api/stores/provisions.js
```diff
@@ -95,7 +96,20 @@ export function useProvisionStore (subscriptionTable, consumerTable, services) {
...
-      return { error: { name: 'ConsumerNotFound', message: 'unimplemented' } }
+      const [consumerRecord, allocated] = await Promise.all([
+        consumerTable.get(provider, consumer),
+        spaceMetricsTable.getAllocated(consumer)
+      ])
+      return consumerRecord ? ({
+        ok: {
+          did: consumer,
+          allocated,
+          total: 1_000_000_000, // set to an arbitrarily high number because we currently don't enforce any limits
```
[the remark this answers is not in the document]
travis: yep I agree! 

in your proposal, would "available" be the total remaining available bytes or the "max" available bytes - ie, is it:

1. allocated + available == total (and we return allocated and available)
or
2. allocated == bytes used and available == total (and we return allocated and available)





[10:05](https://filecoinproject.slack.com/archives/D04AENLQQ3V/p1692723933722439)
(1) makes sense to me, if (2) is what you had in mind I think maximum as you suggest is clearer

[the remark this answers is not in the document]
travis: oh chatted out of band and we decided to go with `limit` as in "you can raise your storage limit by upgrading your subscription"

upload-api/tables/space-metrics.js
```diff
@@ -0,0 +1,48 @@
...
+     * Return the total amount of of storage a space has used.
+     * 
+     * @param {import('@ucanto/interface').DIDKey} consumer the space whose current allocation we should return
+     */
+    getAllocated: async (consumer) => {
+      const response = await dynamoDb.send(new GetItemCommand({
+        TableName: tableName,
+        Key: marshall({
+          consumer,
+          name: 'store/add-size-total'
```
vasco-santos, line 41: could we import const from `ucan-invocation/contants.js` side of things instead of hardcoding?

upload-api/tables/subscription.js
```diff
@@ -72,12 +73,39 @@ export function useSubscriptionTable (dynamoDb, tableName) {
...
+     */
+    get: async (provider, subscription) => {
+      const response = await dynamoDb.send(new GetItemCommand({
+        TableName: tableName,
+        Key: {
+          provider: {
+            S: provider
+          },
+          subscription: {
+            S: subscription
```
vasco-santos, line 97: could we use marshal instead of manually do it?

upload-api/tables/consumer.js
```diff
@@ -41,13 +42,91 @@ export class ConflictError extends Failure {
...
+    getBySubscription: async (provider, subscription) => {
+      const response = await dynamoDb.send(new GetItemCommand({
+        TableName: tableName,
+        Key: {
+          provider: {
+            S: provider
+          },
+          subscription: {
+            S: subscription
+          }
```
vasco-santos, line 120: could we use marshal instead of manually do it?

upload-api/tables/consumer.js
```diff
@@ -41,13 +42,91 @@ export class ConflictError extends Failure {
...
+    },
+    AttributesToGet: ['provider']
+  })
+  const response = await dynamoDb.send(cmd)
+  return response.Items ? response.Items.map(item => {
+    const row = unmarshall(item)
+    return /** @type {import('@web3-storage/upload-api').ProviderDID} */ (row.provider)
+  }) : []
+}
+
```
vasco-santos, line 70: any plan to handle pagination? Probably future problem, but maybe we should comment it and create an issue

vasco-santos (dismissed):
This looks good, just pointed out some small things. More critical would be to make sure we use constants for metrics names to avoid breaking things on changes

upload-api/tables/subscription.js
travis: ah, yes!

upload-api/tables/space-metrics.js
travis: oh great idea, ty

upload-api/tables/consumer.js
travis: ah right yea not handling that - will comment and file an issue, it probably won't be a problem for a long time (if ever?) but worth tracking

travis: huh apparently Query will return 1MB of results? https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html

I can't imagine this ever being larger than 1MB, so maybe we just don't need to worry about it?

vasco-santos: I would add a comment in the code just in case

vasco-santos (approved):
LGTM

PR #222, opened 2023-08-21
feat: add date index to uploads table

travis:
The last piece of the blocking workflow that's not possible right now is being able to get a list of recent uploads. Add an index to the upload table to enable this.

I think this is the right thing to do for now while we figure out what kind of data warehousing setup we want, but we may want to consider removing this later.


travis:
I don't think we need this anymore, since most of the queries I was imagining this using can be handled by Athena once #191  is merged. Closing.

PR #224, opened 2023-09-11
fix: limit max upload size

vasco-santos:
Adds limit based on previous decisions, considering that S3 Put command has hard-limit of 5GiB.

- By limiting CAR size to 127*(1<<25), we guarantee max-4GiB-padded Filecoin pieces and have better utilization of Fil sector space.
- By receiving one more byte, we would immediatly get to 8GiB padded piece.

Closes https://github.com/web3-storage/w3up/issues/827

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4826,9 +4826,9 @@
       }
     },
     "node_modules/@web3-storage/upload-api": {
-      "version": "5.3.0",
-      "resolved": "https://registry.npmjs.org/@web3-storage/upload-api/-/upload-api-5.3.0.tgz",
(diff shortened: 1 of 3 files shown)


vasco-santos:
Also updated `upload-api` to enforce the limit right in the start

Gozala (approved):
Looks good to me. I do think however that it might be a good idea to pull in the data-segment lib and use it's functions to derive this number. That way future changes will propagate without us having to find all the places we have calculated the sizes.

vasco-santos:
> Looks good to me. I do think however that it might be a good idea to pull in the data-segment lib and use it's functions to derive this number. That way future changes will propagate without us having to find all the places we have calculated the sizes.

I would say this is implementation detail. Other users of data-segment can actually have other limits like 8Gb. I will merge but happy to iterate this further, even though I am not clear on what would mean using `data-segment` here. I guess starting from S3 limit of 5GB and find what is the bigger height that fit there. Afterwards, seeing what that size would map for a car size.

PR #225, opened 2023-09-13
feat: implement `getCID` methods on store and upload tables

travis:
this enables the new capability invocation handlers in https://github.com/web3-storage/w3up/pull/918


alanshaw (commented):
Is it possible to call it something closer to what it's retrieving `getCID` to me says retrieve a CID. Is it metadata? You say "information" in the description - `getInfo`, `stat`?

travis:
oh yea - we actually switched to "inspect" in the capability name (https://github.com/web3-storage/w3up/pull/918) - @alanshaw how would you feel about `inspectCID` ?

alanshaw:
Just `inspect`?

travis:
ok updated! I think the errors here may be due to some cleanup @vasco-santos did with the Kinesis streams - I think they should be safe to ignore. @alanshaw anything else blocking approval?

alanshaw (approved):

PR #227, opened 2023-09-14
fix: add needed env variables for local testing

vasco-santos:
My local env was not working and after digging through the dev deployment to find out what was going on, noticed that there were no docs for the Delegations Bucket, so added it now.

Also ended up adding a validation for the environment

### DIFF
--- a/.env.tpl
+++ b/.env.tpl
@@ -14,6 +14,7 @@
 EIPFS_INDEXER_SQS_ARN = 'arn:aws:sqs:us-west-2:505595374361:staging-ep-indexer-topic'
 EIPFS_INDEXER_SQS_URL = 'https://sqs.us-west-2.amazonaws.com/505595374361/staging-ep-indexer-topic'
 
+PROVIDERS = ''
 ACCESS_SERVICE_DID = ''
 UPLOAD_API_DID = ''
 ACCESS_SERVICE_URL = ''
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -140,6 +140,10 @@ Bucket name to replicate written .idx files.
 
 Bucket name to replicate root CID to car CIDs mapping.
 
+#### `R2_DELEGATION_BUCKET_NAME`
```
olizilla, line 143: Do we need to describe `PROVIDERS` here in the read me as well?

olizilla (approved):

.env.tpl
```diff
@@ -14,6 +14,7 @@
 EIPFS_INDEXER_SQS_ARN = 'arn:aws:sqs:us-west-2:505595374361:staging-ep-indexer-topic'
 EIPFS_INDEXER_SQS_URL = 'https://sqs.us-west-2.amazonaws.com/505595374361/staging-ep-indexer-topic'
 
+PROVIDERS = ''
 ACCESS_SERVICE_DID = ''
```
olizilla, line 18: is `ACCESS_SERVICE_DID` used anywhere?

vasco-santos: looks like no, good catch. Removing it

travis:
fabulous, thank you!!

travis (approved):

PR #228, opened 2023-09-14
feat: compute piece for uploaded cars

vasco-santos:
This is the first of a series of PRs to integrate w3up with w3filecoin.

This PR adds Filecoin Stack, where CARPARK bucket is hooked up with a lambda to compute Piece CIDs via Event Bus events. A queue is in the middle to keep the flow steady. Once a PR is computed, its details are written into a DynamoDB table. This will have all the information users might need to ask, in order to then ask Deal Tracker for information for a given deal. In other words, it maps `piece`, `link` (Car CID), `aggregate` and `inclusionProof` (TBD what it actually is, is it a CID for a stored file in a bucket?).

Also integration tests were added.

In a follow up PR, a event will be hooked up from DynamoDB table insert to trigger `segment/add` capability from Aggregator together with writing a claim to equivalency claims.

TODO:
- [x] add https://github.com/web3-storage/fr32-sha2-256-trunc254-padded-binary-tree-multihash/ published module

### DIFF
--- a/filecoin/errors.js
+++ b/filecoin/errors.js
@@ -1,22 +1,7 @@
-export class Failure extends Error {
-  describe() {
-    return this.toString()
-  }
-
-  get message() {
-    return this.describe()
-  }
-
-  toJSON() {
-    const { name, message, stack } = this
-    return { name, message, stack }
-  }
-}
-
 export const DatabaseOperationErrorName = /** @type {const} */ (
   'DatabaseOperationFailed'
 )
-export class DatabaseOperationFailed extends Failure {
+export class DatabaseOperationFailed extends Error {
   get reason() {
     return this.message
   }
@@ -29,7 +14,7 @@ export class DatabaseOperationFailed extends Failure {
 export const GetCarErrorName = /** @type {const} */ (
   'GetCarFailed'
 )
-export class GetCarFailed extends Failure {
+export class GetCarFailed extends Error {
   get reason() {
     return this.message
   }
@@ -42,7 +27,7 @@ export class GetCarFailed extends Failure {
 export const ComputePieceErrorName = /** @type {const} */ (
   'ComputePieceFailed'
 )
-export class ComputePieceFailed extends Failure {
+export class ComputePieceFailed extends Error {
   get reason() {
     return this.message
   }
--- a/filecoin/errors.js
+++ b/filecoin/errors.js
@@ -1,0 +1,53 @@
+export class Failure extends Error {
+  describe() {
+    return this.toString()
+  }
+
+  get message() {
+    return this.describe()
+  }
+
+  toJSON() {
+    const { name, message, stack } = this
+    return { name, message, stack }
+  }
+}
+
+export const DatabaseOperationErrorName = /** @type {const} */ (
+  'DatabaseOperationFailed'
+)
+export class DatabaseOperationFailed extends Failure {
+  get reason() {
+    return this.message
+  }
+
+  get name() {
+    return DatabaseOperationErrorName
+  }
+}
+
+export const GetCarErrorName = /** @type {const} */ (
+  'GetCarFailed'
+)
+export class GetCarFailed extends Failure {
+  get reason() {
+    return this.message
+  }
+
+  get name() {
+    return GetCarErrorName
+  }
+}
+
+export const ComputePieceErrorName = /** @type {const} */ (
+  'ComputePieceFailed'
+)
+export class ComputePieceFailed extends Failure {
+  get reason() {
+    return this.message
+  }
+
+  get name() {
+    return ComputePieceErrorName
+  }
+}
--- a/filecoin/functions/piece-cid-compute.js
+++ b/filecoin/functions/piece-cid-compute.js
@@ -37,6 +37,8 @@ async function computeHandler (event) {
   })
 
   if (error) {
+    console.error(error)
+
     return {
       statusCode: 500,
       body: error.message
--- a/filecoin/functions/piece-cid-compute.js
+++ b/filecoin/functions/piece-cid-compute.js
@@ -1,0 +1,97 @@
+import { S3Client } from '@aws-sdk/client-s3'
+import * as Sentry from '@sentry/serverless'
+
+import { computePieceCid } from '../index.js'
+import { createPieceTable } from '../tables/piece.js'
+
+Sentry.AWSLambda.init({
+  environment: process.env.SST_STAGE,
+  dsn: process.env.SENTRY_DSN,
+  tracesSampleRate: 1.0,
+})
+
+const AWS_REGION = process.env.AWS_REGION || 'us-west-2'
+
+/**
+ * Get EventRecord from the SQS Event triggering the handler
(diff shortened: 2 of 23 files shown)


stacks/index.js
```diff
@@ -21,7 +22,7 @@ export default function (app) {
   app.setDefaultFunctionProps({
     runtime: 'nodejs16.x',
     environment: {
-      NODE_OPTIONS: "--enable-source-maps",
+      NODE_OPTIONS: "--enable-source-maps --experimental-fetch",
```
vasco-santos, line 25: https://github.com/web3-storage/fr32-sha2-256-trunc254-padded-binary-tree-multihash relies on `globalThis.fetch` and we will need to rely on this until [we can upgrade SST here to bump node version](https://github.com/web3-storage/w3infra/pull/152)

filecoin/errors.js
```diff
@@ -0,0 +1,38 @@
+export const DatabaseOperationErrorName = /** @type {const} */ (
+  'DatabaseOperationFailed'
+)
+export class DatabaseOperationFailed extends Error {
+  get reason() {
+    return this.message
```
alanshaw, line 6: Is this necessary? Why not just use the message property?

filecoin/functions/piece-cid-compute.js
```diff
@@ -0,0 +1,97 @@
...
+  const s3Client = new S3Client({ region: record.bucketRegion })
+  const pieceTable = createPieceTable(AWS_REGION, pieceTableName)
+
+  const { error, ok } = await computePieceCid({
+    record,
+    s3Client,
+    pieceTable,
+  })
+
+  if (error) {
```
alanshaw, line 39: Log it so we can get stack if we ever need to!

```suggestion
  if (error) {
    console.error(error)
```

filecoin/index.js
```diff
@@ -0,0 +1,80 @@
...
+    for await (const chunk of res.Body) {
+      hasher.write(chunk)
+    }
+    hasher.digestInto(digestBytes, 0, true)
+    const digest = Digest.decode(digestBytes)
+    // @ts-expect-error some properties from PieceDigest are not present in MultihashDigest
+    piece = Piece.fromDigest(digest)
+  } catch (/** @type {any} */ error) {
+    return {
+      error: new ComputePieceFailed(error.cause)
```
alanshaw, line 66:
```suggestion
      error: new ComputePieceFailed(`failed to compute piece CID for CAR: ${cidString}`, { cause: error })
```

filecoin/index.js
```diff
@@ -0,0 +1,80 @@
...
+  const getCmd = new GetObjectCommand({
+    Bucket: record.bucketName,
+    Key: key,
+  })
+  const res = await s3Client.send(getCmd)
+  if (!res.Body) {
+    return {
+      error: new GetCarFailed(`failed to get CAR file with key ${key} in bucket ${record.bucketName}`)
+    }
+  }
```
alanshaw, line 49: To catch the thrown 404 (and others) and return a nicer error:

```suggestion
  let res
  try {
    res = await s3Client.send(getCmd)
    if (!res.Body) throw new Error('missing body')
  } catch (err) {
    return {
      error: new GetCarFailed(`failed to get CAR: ${record.bucketName}/${key}`, { cause: err })
    }
  }
```

filecoin/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "scripts": {
+    "test": "ava --verbose --timeout=60s **/*.test.js"
+  },
+  "dependencies": {
+    "@aws-sdk/client-dynamodb": "^3.211.0",
+    "@aws-sdk/client-s3": "^3.211.0",
+    "@aws-sdk/client-sqs": "^3.226.0",
+    "@sentry/serverless": "^7.22.0",
+    "@web3-storage/data-segment": "^3.0.1",
+    "fr32-sha2-256-trunc254-padded-binary-tree-multihash": "github:web3-storage/fr32-sha2-256-trunc254-padded-binary-tree-multihash",
```
alanshaw, line 14: I think this can be npm now?

filecoin/tables/piece.js
```diff
@@ -0,0 +1,64 @@
...
+          piece: input.piece.toString(),
+          insertedAt
+        }),
+      })
+
+      try {
+        await dynamoDb.send(cmd)
+      } catch (/** @type {any} */ error) {
+        return {
+          error: new DatabaseOperationFailed(error.cause)
```
alanshaw, line 55: Use the message to communicate what was being attempted when something went wrong?

```suggestion
          error: new DatabaseOperationFailed(`failed put to dynamo piece table, link: ${input.link}, piece: ${input.piece}`, { cause: error })
```

filecoin/test/compute-piece-cid.test.js
```diff
@@ -0,0 +1,127 @@
...
+  } = await createDynamodDb({ port: 8000 })
+
+  Object.assign(t.context, {
+    s3Client: client,
+    dbEndpoint,
+    dynamoClient
+  })
+})
+
+test('computes piece cid from a CAR file in the bucket', async t => {
```
alanshaw, line 37:
```suggestion
test('computes piece CID from a CAR file in the bucket', async t => {
```

stacks/index.js
alanshaw: ACK, lets get that upgrade done soon!

test/helpers/table.js
```diff
@@ -17,7 +18,41 @@ export async function getTableItem (dynamo, tableName, key) {
...
+      return r
+    }, {
+      maxTimeout: 2000,
+      minTimeout: 1000,
+      retries: 100
+    })
+  } catch {}
+
+  console.log('items', response?.Items)
+
```
alanshaw, line 50:
```suggestion
```

test/helpers/table.js
```diff
@@ -17,7 +18,41 @@ export async function getTableItem (dynamo, tableName, key) {
...
+      if (r.$metadata.httpStatusCode === 404 || !r.Count) {
+        throw new Error('not found in dynamoDB yet')
+      }
+      return r
+    }, {
+      maxTimeout: 2000,
+      minTimeout: 1000,
+      retries: 100
+    })
+  } catch {}
```
alanshaw, line 47: I'd let this throw - if we don't find after 100 retries it ain't coming.

alanshaw (approved):
This LGTM, I left a few nits. Please can we also publish an equivalency claim?

vasco-santos:
> This LGTM, I left a few nits. Please can we also publish an equivalency claim?

Yes, it is mentioned in main comment and will come as follow up PR

filecoin/errors.js
vasco-santos: It follows the same pattern as `ucanto` does https://github.com/web3-storage/ucanto/blob/c8999a59852b61549d163532a83bac62290b629d/packages/core/src/result.js#L63 for failures, so that the types of failures from handlers of ucanto server are all well aligned. So, not strictly necessary, but consistent

PR #231, opened 2023-09-18
fix: reduce costs on kinesis streams

vasco-santos:
We are currently being billed quite a bit in AWS for all the Kinesis streams we keep open for dev/CI.

This PR addresses this by:
- not deploying 2 kinesis streams per env to keep historical content. We really just need the production stream still deployed until we read it.
- we can also decrease the persistence of the streams to keep costs down

As follow ups, we should also consider (I will create issues):
- tool that proactively tears down running infrastructure for stalled PRs (or setup tests in a way that kinesis is put on and off for the tests)
- get historical content kinesis stream flushed out

### DIFF
--- a/stacks/config.js
+++ b/stacks/config.js
@@ -87,6 +87,21 @@ export function getKinesisEventSourceConfig (stack) {
   }
 }
 
+/**
+ * @param {import('@serverless-stack/resources').Stack} stack
+ */
+export function getKinesisStreamConfig (stack) {
+  if (stack.stage !== 'production') {
+    return {
+      retentionPeriod: Duration.hours(24)
+    }
+  }
+
+  return {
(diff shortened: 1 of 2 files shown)


stacks/config.js
```diff
@@ -87,6 +87,21 @@ export function getKinesisEventSourceConfig (stack) {
   }
 }
 
+/**
+ * @param {import('@serverless-stack/resources').Stack} stack
+ */
+export function getKinesisStreamConfig (stack) {
+  if (stack.stage !== 'production') {
+    return {
+      retentionPeriod: Duration.hours(24)
```
vasco-santos, line 96: Minimum value supported

travis (approved):
this is great, thanks @vasco-santos!

PR #239, opened 2023-10-04
feat: implement new `RevocationsStorage` methods

travis:
implement new `RevocationsStorage` methods introduced in https://github.com/web3-storage/w3up/pull/941

Added a new "revocations" table to index revocations. The revocations table is designed to be batch-GET-queried by "revoke CIDs", (which means the primary key MUST be just the revoke CID) but needs to accommodate multiple possible "revocation scope DIDs" each with its own "cause CID".

If we could batch-get indices by primary key we could add fields for scope DIDs and cause CIDs, but because `BatchGetItem` only works on primary tables, not indices, we decided to store scope and cause information as scopeDID-keyed Map-type attributes on the row, like:

| revoke       | `scopeDID` (ie, `did:key:z...alice`) | `scopeDID` (ie, `did:key:z...bob`)|
| ---------- |------------------|-----------------|
| `revokeCID` | {cause: `causeCID`} | {cause: `causeCID`} |


upload-api/tables/index.js
```diff
@@ -98,4 +98,25 @@ export const rateLimitTableProps = {
...
+ * the context CID and the cause CIDs into a "set" field of :-separated contextCID:causeCID
+ * strings.
+ * 
+ * @type TableProps 
+*/
+export const revocationTableProps = {
+  fields: {
+    delegation: 'string',     // `baf...x`(CID of the revoked delegation)
+    // @ts-ignore SST seems to incorrectly limit the set of field types to the three types suitable for index keys - hopefully this is fixed in the next versoin
+    contextsAndCauses: 'set'  // a set of :-separated contextCID:causeCID pairs
```
Gozala, line 119: should we consider using serialized JSON instead so it's possible to upgrade, I don't know if it does but thought I'd ask so tradeoffs could be assesed.

upload-api/tables/revocations.js
```diff
@@ -0,0 +1,87 @@
...
+      await dynamoDb.send(new UpdateItemCommand({
+        TableName: tableName,
+        Key: marshall({
+          delegation: delegationCID.toString(),
+        }),
+        UpdateExpression: 'ADD contextsAndCauses :candc',
+        ExpressionAttributeValues: marshall({
+          ':candc': new Set([`${contextCID.toString()}:${causeCID.toString()}`])
+        })
+      }))
```
Gozala, line 43:
```suggestion
      await dynamoDb.send(new UpdateItemCommand({
        TableName: tableName,
        Key: marshall({
          delegation: delegationCID.toString(),
        }),
        UpdateExpression: "SET scope.#scope.cause = :cause",
        ConditionExpression: "attribute_not_exists(scope.#revoke) AND attribute_not_exists(scopes.#scope)",
        ExpressionAttributeNames: {
           "#scope": contextCID.toString(),
           "#revoke": delegationCID.toString()
         },
        ExpressionAttributeValues: {
          ":cause": causeCID.toString()
        },
      }))
```

Assumption above is that instead of modeling records as

```ts
type Model = {
  delegation: ToString<UCANLink>
  contextsAndCauses: `${ToString<UCANLink>}:${ToString<UCANLink>}` 
}
```

We would model records as follows instead

```ts
type Model = {
  delegation: ToString<UCANLink>
  scope: Record<RevocationScope, { cause: ToString<UCANLink> }>
}

type RevocationScope = ToString<UCANLink>
```

I am not sure if above is actually correct, but if it is, I think it's probably better option than stringifying into sets as maps naturally have set semantics. Also by using maps we could in theory write other fields if we wanted to.

I have also added `ConditionExpression` to deal with following two scenarios:

1. We MAY get a revocation with same `delegationCID`, `contextCID`, but different `causeCID` e.g.  revocations was invoked once than again an hour later. In such second invocation is NOOP and should be discarded, storing it will provide no value. Revocation should have taken effect on first revocation anyway.
2. If actual delegation is revoked as in  `contextCID == delegationCID` we no longer care about any other scopes in which `delegationCID` is revoked as it had been revoked, so we probably can stop storing any other redundant revocations.

upload-api/tables/revocations.js
```diff
@@ -0,0 +1,87 @@
...
+    },
+
+    async getRevocations(delegationCIDs) {
+      // BatchGetItem only supports batches of 100 and return values under 16MB
+      // https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html
+      // limiting to 100 should be fine for now, and since we don't return much data we should always
+      // stay under the 16 MB limit
+      if (delegationCIDs.length > 100) {
+        throw new Error('checking for more than 100 revocations in a single call is currently not supported')
+      }
```
Gozala, line 53: I think we should do multiple queries here as opposed to failing, caller will not have a way of handle this other than passing less CIDs and then doing another query.

I can't believe I'm saying this but perhaps return value of this function should be `AsyncIterable<{revoke: UCANLink, scope: UCANLink, cause:UCANLink}[]>` that way caller could check first page of scopes and if invocation still valid, get another page and etc... Which I think it would be better API as it does not require you to handle errors and instead just does query at time.

Gozala: Also please do not throw, just return Result type so errors could be typed.

Gozala (dismissed):

upload-api/tables/index.js
travis: serialized JSON seems fine yep! though I'll probably use CBOR here because (answering your second question as well) the max size of any item in Dynamo is 400KB (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ServiceQuotas.html#limits-items) 

unfortunately I don't think `BS` is workable - the reason I'm using a `set` here is so that we can do an atomic update using the `ADD` update expression you noticed below - if we use just about anything else we'll need to fetch the item before updating it, which imho is a dealbreaker

travis: worth noting - we might be able to use a `list` or `map` type for this column, but I don't think they would be more compact or efficient than `set` and I have not tried to get them working.

also, after thinking about this, I'm not sure serialized JSON or CBOR would be more compact than the `:`-separated value I'm using now, so I think I'd prefer to stick with what I have for now... open to being convinced otherwise, but I'd like to optimize for the smallest set values possible!

upload-api/tables/revocations.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+export function useRevocationsTable(dynamoDb, tableName) {
+  return {
+    async put (delegationCID, contextCID, causeCID) {
+      await dynamoDb.send(new UpdateItemCommand({
+        TableName: tableName,
+        Key: marshall({
+          delegation: delegationCID.toString(),
+        }),
+        UpdateExpression: 'ADD contextsAndCauses :candc',
```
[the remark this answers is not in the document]
travis: oh yea great idea - this is the whole reason I'm using a `set` above, so definitely needs comments

upload-api/tables/revocations.js
```diff
@@ -0,0 +1,87 @@
...
+        ExpressionAttributeValues: marshall({
+          ':candc': new Set([`${contextCID.toString()}:${causeCID.toString()}`])
+        })
```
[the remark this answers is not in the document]
travis: yep! the `ADD` operation in the `UpdateExpression` above handles adding this to the set of `contextCID:causeCID` strings. in the case where `delegationCID` and `contextCID` match, we'll just have one of the items in this set whose first half matches the partition key, shouldn't be a problem.

upload-api/tables/revocations.js
travis: oh wow! I'll take this for a spin - if we can use the `map` type like this it does seem conceptually cleaner to me

upload-api/tables/revocations.js
travis: yeaaaa I just figured we won't see this case for a while - if someone is sending us a 100+ item delegation chain I'm pretty comfortable just throwing an error for now....

re: `Result` this is the pattern we tend to use for `w3infra`-internal tables (ie, tables that are not part of the interface between `w3up` and `w3infra`):

https://github.com/web3-storage/w3infra/blob/main/upload-api/tables/consumer.js
https://github.com/web3-storage/w3infra/blob/main/upload-api/tables/store.js
https://github.com/web3-storage/w3infra/blob/main/upload-api/tables/subscription.js
https://github.com/web3-storage/w3infra/blob/main/upload-api/tables/upload.js

so I'd prefer to stick with this style for consistency. I'm not opposed to switching to using `Result` for these tables, but it feels like it's worth a broader conversation and a concerted effort to convert them all if we do go that way... what do you think?

travis: adding - with the new interface I think this will actually be a standalone store instead of an internal table, so I'll use `Result` after all!

upload-api/tables/revocations.js
travis: ok so unfortunately it looks like setting values on a new record will fail:

https://thomasstep.com/blog/handling-dynamodb-validation-exception-for-maps

this means we'd need to do two requests, one that will fail, for the most common case where we are receiving the first revocation for a particular delegation, which seems like a bad tradeoff to me since the `set` implementation works and is probably more compact - definitely open to continuing to iterate on this but I think the `set` is still the way to go for now!

upload-api/tables/revocations.js
travis: done

upload-api/tables/revocations.js
Gozala: That is surprising I was under impression it will be created automatically. That said I think there is a pretty straightforward solution that is,

```
UpdateExpression: "SET scope = if_not_exists(scopes, :emptyMap), SET scope.#scope.cause = :cause"
```

There are also [some examples here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html#Expressions.UpdateExpressions.SET.AddingListsAndMaps%23Expressions.UpdateExpressions.SET.AddingListsAndMaps:~:text=after%20the%20update.-,Adding%20lists%20and%20maps,L%22%3A%20%5B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%7B%20%22S%22%3A%20%22Best%20product%20ever!%22%20%7D%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%5D%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%7D,-Adding%20elements%20to)

travis:
Ah thanks! Will give this another go in the morning!

upload-api/stores/revocations.js
```diff
@@ -0,0 +1,100 @@
...
+            revoke: revocation.revoke.toString(),
+          }),
+          // when we add or update this item, use the update expression to 
+          // add a string to the "details" set that contains revokeCID and
+          // causeCID in a :-separated string
+          // see https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html
+          // for more information about update expressions
+          UpdateExpression: 'ADD details :details',
+          ExpressionAttributeValues: marshall({
+            ':details': new Set([`${revocation.scope.toString()}:${revocation.cause.toString()}`])
```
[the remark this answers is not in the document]
vasco-santos: Using the first option will require a QueryCommand below, is it problematic?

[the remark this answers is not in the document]
vasco-santos: should we set a limit on the reason string size? Allowing user to provide huge string that we would accept MAY not be good. And that might help to not get into limits

upload-api/tables/revocations.js
travis: shoot, I don't think this is going to be workable either - I changed this to:

```
          UpdateExpression: 'SET details = if_not_exists(details, :emptyMap), details.#scope.cause = :cause',
          ExpressionAttributeNames: {
            '#scope': revocation.scope.toString()
          },
```

and got a `ValidationException` with this message:

```
    message: 'Invalid UpdateExpression: Two document paths overlap with each other; must remove or rewrite one of these paths; path one: [details], path two: [details, bafyreibjhuiw57hsbbf6u5t5plcfglq6gpz263wtxthbz5d3ylaycqeifq, cause]',
```

I don't see any examples in the document you linked of doing precisely this - I did find some examples that use two roundtrips on Stack Overflow - revocation will probably be infrequent enough that two requests isn't the end of the world, but it definitely isn't ideal... then again my set-based implementation isn't ideal either!

travis: ok we chatted in high bandwidth and came up with a design where we use the scopeCIDs as the attribute key of a value containing the per-revocation metadata. so the table looks like:

| revoke       | `scopeCID` (ie, bafy....asd1) | `scopeCID` (ie, bafy...asd2)|
| ---------- |------------------|-----------------|
| `revokeCID` | {cause: `causeCID`} | {cause: `causeCID`} |

this lets us query with `BatchGetItem` and we use the SST schema definition to filter out the static column keys, and then assume the rest are scopeCID: causeMetadataMaps  pairs

upload-api/stores/revocations.js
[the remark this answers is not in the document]
travis: oh the reason will always be a CID, so it should be pretty limited!

alanshaw (approved):
🚀

PR #240, opened 2023-10-16
feat: billing

alanshaw:
I've built so that `billing/lib` can be extracted to w3up, published separately and used here.

Screenshot 2023-11-01 at 10 52 52

TODO:

* [x] Env vars for queue URLs etc.
* [x] Add test for UCAN invocation handler
* [x] DLQs
* [x] Add code to invoke Stripe

Post merge:

* [ ] Add space size snapshots for each space in the system
* [x] Ensure `Customer` table get populated when user selects plan


travis (approved):
a few small tweaks but this looks great to me, love the increased rigor around values coming in and out of the system, hope we can propagate these patterns out to the rest of the repo!

billing/functions/usage-table.js
```diff
@@ -0,0 +1,50 @@
...
+
+export const handler = Sentry.AWSLambda.wrapHandler(
+  /**
+   * @param {import('aws-lambda').DynamoDBStreamEvent} event
+   * @param {import('aws-lambda').Context} context
+   */
+  async (event, context) => {
+    // /** @type {CustomHandlerContext|undefined} */
+    // const customContext = context?.clientContext?.Custom
+    // TODO: stripe publishable key or something?
```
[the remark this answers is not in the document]
alanshaw: 🤪 no you're right, publishable is for the client...seeing cross eyed at this point

billing/functions/billing-cron.js
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * @param {import('aws-lambda').ScheduledEvent} event
+   * @param {import('aws-lambda').Context} context
+   */
+  async (event, context) => {
+    /** @type {CustomHandlerContext|undefined} */
+    const customContext = context?.clientContext?.Custom
+    const customerTable = customContext?.customerTable ?? notNully(process.env, 'CUSTOMER_TABLE_NAME')
+    const customerBillingQueueURL = new URL(customContext?.customerBillingQueueURL ?? notNully(process.env, 'CUSTOMER_BILLING_QUEUE_URL'))
+    const region = customContext?.region ?? notNully(process.env, 'AWS_REGION')
```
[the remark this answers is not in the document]
alanshaw: ACK yes I invented it because I wanted it for getting properties from a config object as well as `process.env` but not required here so I'll switch.

billing/lib/api.ts
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Captures details about a space that should be billed for a given customer in
+ * the given period of usage.
+ */
+export interface SpaceBillingInstruction extends CustomerBillingInstruction {
+  /** Space DID (did:key:...). */
+  space: DID
+  /** Storage provider for the space. */
+  provider: DID<'web'>
```
Gozala, line 143: Could be a good idea to define `type Provider = DID<'web'>` and use that, same for `Consumer`  and `Customer` that way it reads more clearly and we can update definitions in one place when necessary.

billing/lib/api.ts
```diff
@@ -0,0 +1,290 @@
...
+> extends UcanMessage<C> {
+  type: 'receipt'
+  out: R
+}
+
+export interface UcanWorkflowMessage<C extends Capabilities = Capabilities> extends UcanMessage<C> {
+  type: 'workflow'
+}
+
+export type UcanStreamMessage<C extends Capabilities = Capabilities> = UcanWorkflowMessage<C> | UcanReceiptMessage<C>
```
Gozala, line 211: Something about these types are kind of off, [here is definition in ucanto](https://github.com/web3-storage/ucanto/blob/43ea497ca2200f9e1e15dbbdd033ec24895edb7d/packages/interface/src/lib.ts#L802-L821) and few things to note:

1. We could have more than one invocation / receipt in a single message.
2. Message can contain both invocation and receipt even though often it's either on or another.
3. We intentionally did not use term `workflow` because IPVM calls it something else.

Probably none of the above is critical, but might be worth considering.

billing/lib/util.js
```diff
@@ -0,0 +1,15 @@
+export const startOfMonth = () => {
+  const d = new Date()
```
Gozala, line 2: I think it would be a good idea to pass seed here

```suggestion
export const startOfMonth = (now) => {
  const d = new Date(now)
```

billing/lib/util.js
```diff
@@ -0,0 +1,15 @@
...
+  d.setUTCDate(1)
+  d.setUTCHours(0)
+  d.setUTCMinutes(0)
+  d.setUTCSeconds(0)
+  d.setUTCMilliseconds(0)
+  return d
+}
+
+export const startOfLastMonth = () => {
+  const d = startOfMonth()
```
Gozala, line 12:
```suggestion
export const startOfLastMonth = (now) => {
  const d = startOfMonth(now)
```

billing/lib/billing-cron.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * @param {{
+ *   customerStore: import('./api.js').CustomerStore
+ *   customerBillingQueue: import('./api.js').CustomerBillingQueue
+ * }} ctx
+ * @returns {Promise<import('@ucanto/interface').Result>}
+ */
+export const handleCronTick = async ctx => {
+  const from = startOfLastMonth()
+  const to = startOfMonth()
```
Gozala, line 12: I think it's a good idea to pass seed time instead of making those utils effectful, also in theory you may run into two moth range if this happens to run at really unlucky time.

```suggestion
  const now = Date.now()
  const from = startOfLastMonth(now)
  const to = startOfMonth(now)
```

billing/lib/billing-cron.js
```diff
@@ -0,0 +1,37 @@
...
+
+  /** @type {string|undefined} */
+  let cursor
+  while (true) {
+    const customerList = await ctx.customerStore.list({}, { cursor, size: 1000 })
+    if (customerList.error) return customerList
+
+    for (const c of customerList.ok.results) {
+      console.log(`Adding customer billing instruction for: ${c.customer}`)
+      const queueAdd = await ctx.customerBillingQueue.add({
```
Gozala, line 22: Is queue idempotent here, if so worth calling out in the comment above.

billing/lib/billing-cron.js
```diff
@@ -0,0 +1,37 @@
...
+        customer: c.customer,
+        account: c.account,
+        product: c.product,
+        from,
+        to
+      })
+      if (queueAdd.error) return queueAdd
```
Gozala, line 29: Comment explaining what happens on error would be really nice, do we restart everything from start or do we have some way of resuming from where error occured.

billing/lib/customer-billing-queue.js
```diff
@@ -0,0 +1,43 @@
...
+
+  /** @type {string|undefined} */
+  let cursor
+  while (true) {
+    const subsList = await ctx.subscriptionStore.list({ customer: instruction.customer }, { cursor })
+    if (subsList.error) return subsList
+
+    // TODO: this is going to be inefficient for any client with many spaces
+    // and may eventually cause billing to fail.
+    for (const s of subsList.ok.results) {
```
Gozala, line 22: Could we just iterate over subscription store as opposed to first iterating over customers and then getting all the subscriptions that customer has ?

billing/lib/space-billing-queue.js
```diff
@@ -0,0 +1,67 @@
...
+  while (true) {
+    const spaceDiffList = await ctx.spaceDiffStore.listBetween(
+      { customer: instruction.customer },
+      instruction.from,
+      instruction.to,
+      { cursor, size: 1000 }
+    )
+    if (spaceDiffList.error) return spaceDiffList
+    for (const diff of spaceDiffList.ok.results) {
+      if (diff.provider !== snap.provider) continue
```
Gozala, line 41: Should we amend store to allow passing provider into a query ?

Gozala (approved):
I have provided bunch of feedback geared towards making it easier for others to understand whats going on. That said I don't think I have caught any critical issues to block this from merging. Leaving it up to you to decide what to do with all the feedback.

billing/data/consumer.js
```diff
@@ -0,0 +1,83 @@
...
+
+/** @type {import('../lib/api').Encoder<Consumer, ConsumerStoreRecord>} */
+export const encode = input => {
+  try {
+    return {
+      ok: {
+        consumer: input.consumer,
+        provider: input.provider,
+        subscription: input.subscription,
+        cause: input.cause.toString(),
```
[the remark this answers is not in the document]
alanshaw: Well yes this is kind of what I was thinking with the UCAN invocation stream but as you rightly pointed out dag-json serialization of DIDs wouldn't have worked. I think since we have precedence for strings everywhere else here I'm inclined to keep as is.

billing/data/consumer.js
```diff
@@ -0,0 +1,83 @@
...
+
+/** @type {import('../lib/api').Decoder<StoreRecord, Consumer>} */
+export const decode = input => {
+  try { 
+    return {
+      ok: {
+        consumer: asDID(input.consumer),
+        provider: asDIDWeb(input.provider),
+        subscription: String(input.subscription),
+        cause: Link.parse(String(input.cause)),
```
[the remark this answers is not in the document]
alanshaw: Should be a type cast not an actual cast.

billing/lib/api.ts
alanshaw: Yes agreed they are not great. They are an improvement on https://github.com/web3-storage/w3infra/blob/fe3e3c6f3385e2b519eb1b296c2f3de0e12d1390/ucan-invocation/types.ts#L47-L76 which would have needed a whole lot of type casting / use of `any` to be usable. Unfortunately the UCAN stream is coupled to the current invocations as they exist in ucanto plus our hacked in receipts. I really don't want to increase the scope of this PR to fix this, but propose instead that when we implement the invocation spec we do it.

Issue #243, opened 2023-10-18
Athena tables should be automatically generated from a list of capabilities


travis:
`@web3-storage/capabilities` exposes Ucanto `Capability` objects that capture schema information for `nb` fields, so we should have everything we need to generate [Glue table schemas](https://github.com/web3-storage/w3infra/blob/5591d86b4384086802710dc5bb78993f24468c11/stacks/firehose-stack.js#L235) automatically, which would make it easier to make data available to Athena for querying.

The idea is that something like:

```ts
generateTable(Store.add)
```

would result in something like [the following SST code](https://github.com/web3-storage/w3infra/blob/5591d86b4384086802710dc5bb78993f24468c11/stacks/firehose-stack.js#L235):

```js
  const storeAddTableName = getCdkNames('store-add-table', app.stage)
  const storeAddTable = new glue.CfnTable(stack, storeAddTableName, {
    catalogId: Aws.ACCOUNT_ID,
    databaseName,
    tableInput: {
      name: storeAddTableName,
      partitionKeys: [
        { name: 'day', type: 'date' }
      ],
      parameters: {
        classification: "json",
        typeOfData: "file",
        // @see https://docs.aws.amazon.com/athena/latest/ug/partition-projection-kinesis-firehose-example.html for more information on projection
        // configuration - this should match the "day" parameter and S3 prefix configured in the delivery stream
        "projection.enabled": "true",
        "projection.day.type": "date",
        "projection.day.format": "yyyy-MM-dd",
        "projection.day.range": "2023-01-01,NOW",
        "projection.day.interval": "1",
        "projection.day.interval.unit": "DAYS",
        "storage.location.template": `s3://${streamLogBucket.bucketName}/logs/receipt/store_add/\${day}/`
      },
      storageDescriptor: {
        location: `s3://${streamLogBucket.bucketName}/logs/receipt/store_add/`,
        columns: [
          { name: 'carcid', type: 'string' },
          // STRUCT here refers to the Apache Hive STRUCT datatype - see https://aws.amazon.com/blogs/big-data/create-tables-in-amazon-athena-from-nested-json-and-mappings-using-jsonserde/
          { name: 'value', type: 'STRUCT<att:ARRAY<struct<can:STRING,with:STRING,nb:STRUCT<size:BIGINT,link:STRUCT<_cid_slash:STRING>>>>,iss:STRING,aud:STRING>' },
[...]

PR #244, opened 2023-10-18
feat: add two more capability tables for Athena

travis:
Add two more capability tables to Glue so Athena can query them.

I'd like to start doing this in an automated way at some point, but that doesn't feel high priority enough for the moment so I've filed an issue here:

https://github.com/web3-storage/w3infra/issues/243

### DIFF
--- a/stacks/firehose-stack.js
+++ b/stacks/firehose-stack.js
@@ -9,6 +9,7 @@ import {
   aws_athena as athena,
   aws_sam as sam
 } from 'aws-cdk-lib'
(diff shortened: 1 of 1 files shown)


stacks/firehose-stack.js
```diff
@@ -9,6 +9,7 @@ import {
   aws_athena as athena,
   aws_sam as sam
 } from 'aws-cdk-lib'
+import { Access, Provider, Space, Store, Upload, Consumer, Customer, RateLimit, Subscription, Filecoin, Admin, UCAN } from '@web3-storage/capabilities'
```
vasco-santos, line 12:
```suggestion
```

vasco-santos (approved):

PR #246, opened 2023-10-19
feat: allow customers to create more than one space

travis:
this is just a matter of changing how we generate subscription IDs. they are now generated from customer AND consumer, which means they will be unique for a given customer/consumer pair - in other words, a given customer will only be able to provision a given space one time.

### DIFF
--- a/upload-api/stores/provisions.js
+++ b/upload-api/stores/provisions.js
@@ -4,14 +4,14 @@ import { CBOR, Failure } from '@ucanto/server'
 
 /**
  * Create a subscription ID for a given provision. Currently 
- * uses a CID generated from `customer` and `consumer` which ensures each customer
- * will get at most one subscription per space.
+ * uses a CID generated from `consumer` which ensures a space
+ * can be provisioned at most once.
  * 
  * @param {import('@web3-storage/upload-api').Provision} item 
  * @returns string
  */
 export const createProvisionSubscriptionId = async ({ customer, consumer }) =>
-  (await CBOR.write({ customer, consumer })).cid.toString()
+  (await CBOR.write({ consumer })).cid.toString()
 
 /**
  * @param {import('../types').SubscriptionTable} subscriptionTable
--- a/upload-api/stores/provisions.js
+++ b/upload-api/stores/provisions.js
@@ -4,16 +4,14 @@ import { CBOR, Failure } from '@ucanto/server'
 
 /**
  * Create a subscription ID for a given provision. Currently 
- * uses the CID of `customer` which ensures each customer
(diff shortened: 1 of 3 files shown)


travis:
yea, surprisingly there isn't - in retrospect I definitely should have added one during the migration from D3 but it looks like I overlooked it! 

There is a test in `w3up` that broke during that migration because we started restricting users to one space per user, which I skipped because there wasn't an easy way to get it working and we were already talking about loosening the restriction - I've re-enabled that in a separate PR:

https://github.com/web3-storage/w3up/pull/989

I do think it would be a good idea to have a test that verifies a user can create multiple spaces - thinking through that I think the original test slipped through the cracks because most of the service-level tests live in w3up, but this is explicitly a behavior we'd like implementors to be able to control with their `ProvisionsStorage` implementation - I think the right test for this lives at the `ProvisionsStorage` level, so I'll get that into this PR asap.

travis:
ok - updated with an integration test!

upload-api/stores/provisions.js
```diff
@@ -4,16 +4,14 @@ import { CBOR, Failure } from '@ucanto/server'
 
 /**
  * Create a subscription ID for a given provision. Currently 
- * uses the CID of `customer` which ensures each customer
- * will get at most one subscription. This can be relaxed (ie,
- * by deriving subscription ID from customer AND consumer) in the future
- * or by other providers for flexibility.
+ * uses a CID generated from `customer` and `consumer` which ensures each customer
+ * will get at most one subscription per space.
```
alanshaw, line 8: It'll also allow existing spaces that have already been provisioned to be provisioned again...can we fix/avoid?

travis: oh are you thinking of the case where someone else comes along and provisions a space I created? that would definitely create a thorny billing situation, though in the long run it would be nice to let more than one person sign up to pay for a space.

maybe we can just generate the billing ID from the space DID?

travis: oh HMMM good point, maybe this should just be generated from "consumer" so that a given space can only ever be provisioned once... will make that tweak and add a test!

travis: oh it turns out we already check for this in provisioning! I would like to leave the door open for a "multi-payer" space in the future so I think we should leave as currently implemented!

gobengo:
@travis btw I tested w3infra from your commit at aafebb21b85036aa074b947d279dcae8f6c8d657 locally and confirmed that it resolves https://github.com/web3-storage/w3cli/issues/112

I'll test again once deployed

PR #250, opened 2023-10-31
feat: compute piece cid behind env var

vasco-santos:
So that we can transition into new deployments for w3filecoin without creating noise in w3infra with errors. 

This was test data and is also still with outdated PieceCIDv2

### DIFF
--- a/filecoin/functions/piece-cid-compute.js
+++ b/filecoin/functions/piece-cid-compute.js
@@ -21,7 +21,17 @@ const AWS_REGION = process.env.AWS_REGION || 'us-west-2'
 async function computeHandler (event) {
   const {
     pieceTableName,
+    disablePieceCidCompute
   } = getEnv()
+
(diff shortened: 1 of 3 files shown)


filecoin/functions/piece-cid-compute.js
```diff
@@ -60,6 +70,7 @@ export const handler = Sentry.AWSLambda.wrapHandler(computeHandler)
 function getEnv () {
   return {
     pieceTableName: mustGetEnv('PIECE_TABLE_NAME'),
+    disablePieceCidCompute: process.env.DISABLE_PIECE_CID_COMPUTE
```
alanshaw, line 73:
```suggestion
    disablePieceCidCompute: process.env.DISABLE_PIECE_CID_COMPUTE === 'true'
```

alanshaw (approved):

PR #252, opened 2023-11-01
feat: integrate new storefront service and piece cid

vasco-santos:
Integrates new storefront service and events, together with the new PieceCIDV2.

Note that:
- New Piece store DB
  - the older one has outdated PieceCID versions, however we should be able to re-use those to re-compute the new version and add them to new table. Can be follow up migration when we want to also add backlog to Filecoin
- As previously agreed, for MVP we allow agent to send us `piece/offer` invocation, but we do not queue it as in https://github.com/web3-storage/w3filecoin/blob/main/docs/w3filecoin.svg
  - we rely on https://github.com/web3-storage/w3up/blob/main/packages/filecoin-api/src/storefront/api.ts#L31 to skip submission queue on `piece/offer`. Instead, on bucket put, we put it to the queue right away as we will not trust piece computations at this point. However, this allows user sending correct piece to follow receipt chain :)
- Unfortunately, when storing receipts we only store the out field, which means we won't have the effects. Therefore, both receipt and task stores will need to read from invocation bucket + workflow bucket.
  - Irakli and I already talked about doing changes here once we land new UCAN spec and have the tasks, so I suggest we deal with this as we do in the PR for MVP


alanshaw:
> As previously agreed, for MVP we allow agent to send us piece/offer invocation

Do you mean `filecoin/offer`?

vasco-santos:
> Do you mean filecoin/offer?

yes!

filecoin/store/piece.js
```diff
@@ -0,0 +1,315 @@
...
+    has: async (key) => {
+      const getCmd = new GetItemCommand({
+        TableName: tableName,
+        Key: marshall(encodeKey(key)),
+      })
+      let res
+      try {
+        res = await dynamoDb.send(getCmd)
+      } catch (/** @type {any} */ error) {
+        console.log('errr', error)
```
alanshaw, line 218:
```suggestion
        console.error(error)
```

filecoin/store/piece.js
```diff
@@ -0,0 +1,315 @@
...
+      }
+
+      return {
+        ok: decodeRecord(
+          /** @type {PieceStoreRecord} */ (unmarshall(res.Item))
+        )
+      }
+    },
+    has: async (key) => {
+      const getCmd = new GetItemCommand({
```
alanshaw, line 210:
```suggestion
      const getCmd = new HeadItemCommand({
```

filecoin/store/receipt.js
```diff
@@ -0,0 +1,131 @@
...
+ * @param {S3Client} s3client
+ * @param {string} invocationBucketName
+ * @param {string} workflowBucketName
+ * @returns {import('@web3-storage/filecoin-api/storefront/api').ReceiptStore}
+ */
+export const useReceiptStore = (s3client, invocationBucketName, workflowBucketName) => {
+  return {
+    put: async (record) => {
+      return {
+        error: new StoreOperationFailed('no new receipt should be put by storefront')
```
alanshaw, line 36: ...then why does the `ReceiptStore` type require it? I had the same problem with billing, you could either use `Omit` or have multiple store types that you combine for different requirements e.g. https://github.com/web3-storage/w3infra/blob/72a628f2e3ae20cd04ed81391a2ba5959c6d6def/billing/lib/api.ts#L294-L310

Not blocking!

stacks/upload-api-stack.js
```diff
@@ -68,6 +76,9 @@ export function UploadApiStack({ stack, app }) {
           WORKFLOW_BUCKET_NAME: workflowBucket.bucketName,
           UCAN_LOG_STREAM_NAME: ucanStream.streamName,
           ADMIN_METRICS_TABLE_NAME: adminMetricsTable.tableName,
+          PIECE_TABLE_NAME: adminMetricsTable.tableName,
```
alanshaw, line 79:
```suggestion
          PIECE_TABLE_NAME: pieceTable.tableName,
```

stacks/upload-db-stack.js
```diff
@@ -49,7 +49,7 @@ export function UploadDbStack({ stack, app }) {
    * This table takes a stored CAR and makes an entry in the piece table
    * Used by the filecoin/* service capabilities. // TODO
    */
-  const pieceTable = new Table(stack, 'piece', {
+  const pieceTable = new Table(stack, 'piece-1', {
```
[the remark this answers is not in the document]
alanshaw: Some prior art here as `v1`:

* https://github.com/web3-storage/content-claims/blob/main/stacks/db.js#L7
* https://github.com/elastic-ipfs/infrastructure/blob/cce465fc63265037d10fa4b7d8e6122ed08d7d9d/terraspace/app/stacks/shared/tfvars/base.tfvars#L15

```suggestion
  const pieceTable = new Table(stack, 'piece-v1', {
```

alanshaw (changes requested):

filecoin/functions/handle-piece-offer-message.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  const context = {
+    aggregatorService: {
+      connection,
+      invocationConfig: {
+        issuer: storefrontSigner,
+        with: storefrontSigner.did(),
+        audience: connection.id,
+        proofs: aggregatorServiceProofs
```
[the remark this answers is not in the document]
vasco-santos: [that is for storefront to invoke `piece/offer`](https://github.com/web3-storage/w3up/blob/main/packages/filecoin-api/src/storefront/events.js#L59), maybe the name should be `aggregatorProofs`

filecoin/package.json
```diff
@@ -16,17 +16,20 @@
     "@ucanto/principal": "^9.0.0",
     "@ucanto/transport": "^9.0.0",
     "@web3-storage/content-claims": "^3.2.0",
-    "@web3-storage/data-segment": "^3.0.1",
-    "@web3-storage/filecoin-client": "^1.4.0",
-    "fr32-sha2-256-trunc254-padded-binary-tree-multihash": "^1.0.0",
+    "@web3-storage/data-segment": "^5.0.0",
```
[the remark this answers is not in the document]
vasco-santos: no, looks like is v5 https://github.com/web3-storage/data-segment/releases 🤔

filecoin/store/piece.js
vasco-santos: this is [dynamo](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/), there is no Head command

filecoin/store/receipt.js
vasco-santos: We need PUT in the tests interface, I agree it is not ideal. I will fill in an issue to make it specific for src and test

filecoin/store/piece.js
alanshaw: haha sorry!

filecoin/package.json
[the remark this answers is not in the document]
alanshaw: 🤦‍♂️

filecoin/functions/handle-piece-status-update.js
```diff
@@ -0,0 +1,113 @@
...
+  const { PRIVATE_KEY: privateKey } = Config
+  const { serviceDid, serviceUrl, did, delegatedProof } = getEnv()
+  let storefrontSigner = getServiceSigner({
+    privateKey
+  })
+  const connection = getServiceConnection({
+    did: serviceDid,
+    url: serviceUrl
+  })
+  const aggregatorServiceProofs = []
```
alanshaw, line 57: Should be used somewhere?

filecoin/functions/handle-piece-status-update.js
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Get Env validating it is set.
+ */
+function getEnv () {
+  return {
+    did: mustGetEnv('DID'),
+    serviceDid: mustGetEnv('SERVICE_DID'),
+    serviceUrl: mustGetEnv('SERVICE_URL'),
+    delegatedProof: mustGetEnv('PROOF'),
```
alanshaw, line 99: I'm not sure this one got updated?

stacks/config.js
```diff
@@ -4,7 +4,7 @@ import { StartingPosition } from 'aws-cdk-lib/aws-lambda'
 import git from 'git-rev-sync'
 
 /**
- * Get nicer bucket names
+ * Get nicer burcket names
```
alanshaw, line 7:
```suggestion
 * Get nicer bucket names
```

alanshaw (approved):
LGTM, just a couple of questions where I think things were not updated, but afterwards I think it's fine to merge.

filecoin/store/receipt.js
alanshaw: Ok cool. I defined extended test stores for put/get etc. https://github.com/web3-storage/w3infra/blob/72a628f2e3ae20cd04ed81391a2ba5959c6d6def/billing/test/lib/api.ts#L25

PR #254, opened 2023-11-02
fix: ucan message types and parser

alanshaw:
Turns out `invocationCid` is only set on receipt messages.


billing/lib/api.ts
```diff
@@ -220,13 +219,15 @@ export interface UcanMessageValue<C extends Capabilities = Capabilities> {
   aud: DID,
   iss?: DID,
   prf?: Array<LinkJSON<Link>>
+  cid: Link
```
vasco-santos, line 222: 👍🏼

vasco-santos (approved):
LGTM

PR #256, opened 2023-11-03
fix: allow optional fields

alanshaw:
Some fields i.e. `cause` because required after a certain time so need to be marked optional in the types and expected to be `undefined` sometimes when encoding/decoding.

Other fields like `updatedAt` are explicitly only set when the record has been updated.


vasco-santos (approved):
LGTM

PR #264, opened 2023-11-07
feat: let user select a plan when authorizing

travis:
After the user verifies their email, show them a pricing table if they haven't already selected a plan. This will let them finish this part of the flow without being redirected to the console again later.

### DIFF
--- a/upload-api/functions/validate-email.jsx
+++ b/upload-api/functions/validate-email.jsx
@@ -1,6 +1,7 @@
 import * as Sentry from '@sentry/serverless'
 import { authorize } from '@web3-storage/upload-api/validate'
 import { Config } from '@serverless-stack/node/config/index.js'
+import * as DidMailto from '@web3-storage/did-mailto'
 import { getServiceSigner, parseServiceDids } from '../config.js'
 import { Email } from '../email.js'
 import { createDelegationsTable } from '../tables/delegations.js'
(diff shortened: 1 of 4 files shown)


upload-api/functions/validate-email.jsx
```diff
@@ -134,11 +144,21 @@ export async function validateEmailPost (request) {
 
   const { email, audience, ucan } = authorizeResult.ok
 
+  const planCheckResult = await context.customerStore.get({ customer: DidMailto.fromEmail(email) })
+  let stripePricingTableId
+  let stripePublishableKey
+  if (!planCheckResult.ok?.product){
```
gobengo, line 150: is the idea here to only show the pricing table in the case where that account hasn't already chosen one?

travis:
Yep exactly. This could be cleaner but I think we will move this all to
console sooner rather than later

upload-api/functions/validate-email.jsx
travis: Yep exactly. This could be cleaner but I think we will move this all to
console sooner rather than later

alanshaw:
I adopting this

alanshaw:
Screenshot 2023-11-13 at 15 52 04

alanshaw (approved):

PR #265, opened 2023-11-07
fix: change partition key for space diff to not use customer

alanshaw:
There's no reason for space diffs to store the customer, or include it in the partition key 🤦. If the customer for a given space changes we cannot accurately account for space size changes. I'm not sure why it would, but if it did we'd not be able to get an accurate list of changes for a space.

Also, for retrieving usage in the frontend it's also problematic as we have to somehow look up the customer for a given space before we can get hold of the diffs.

### DIFF
--- a/billing/data/space-diff.js
+++ b/billing/data/space-diff.js
@@ -10,7 +10,6 @@ import { EncodeFailure, DecodeFailure, Schema } from './lib.js'
  */
 
 export const schema = Schema.struct({
-  customer: Schema.did({ method: 'mailto' }),
   space: Schema.did(),
   provider: Schema.did({ method: 'web' }),
   subscription: Schema.text(),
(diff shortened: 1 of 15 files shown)


vasco-santos:
@alanshaw can we fix the tests?

I was wondering if that would be possible given dynamoDB, but learnt that you compute the `pk` and is not a "single column". However, also relevant here to know how we will migrate old records?

billing/data/space-diff.js
```diff
@@ -28,9 +27,8 @@ export const encode = input => {
   try {
     return {
       ok: {
-        pk: `${input.customer}#${input.provider}#${input.space}`,
+        pk: `${input.provider}#${input.space}`,
```
vasco-santos, line 30: can we please update `billing/tables/space-diff` comment on how `pk` is formated?

alanshaw:
> @alanshaw can we fix the tests?

Gah sorry, I ran them locall and they passed but i guess I broke something after that...

vasco-santos (changes requested):

alanshaw:
> However, also relevant here to know how we will migrate old records?

I have made a note to migrate them before the end of the month.

I just need to iterate the table and swap key from `customer#provider#space` to `provider#space`. It'll be a delete and insert.

vasco-santos (approved):
This looks good, lets get merged https://github.com/web3-storage/w3infra/pull/266 and see if it fixes the CI here

Issue #267, opened 2023-11-07
Create integration tests that use the Stripe webhook handler to configure a plan


travis:
It would be nice to have a test for our Stripe integration. I think this can be added to the existing integration test flow in [`test/integration.test.js`](https://github.com/web3-storage/w3infra/blob/main/test/integration.test.js), but we will need to coordinate webhook signing secrets and call into the billing service for the first time, so I think it will take a bit of work to do properly. The following is a rough outline of how I think this should work:

1) we will need to add an additional step to [`setupNewClient`](https://github.com/web3-storage/w3infra/blob/main/test/helpers/up-client.js#L47) that uses [`stripe.webhooks.generateTestHeaderString`](https://github.com/stripe/stripe-node#testing-webhook-signing) to create a signed webhook request
2) we'll need to POST that signed request to the billing service's new `/stripe` endpoint - this will require us to teach `setupNewClient` about the billing service, since it uses a different HTTP endpoint than the upload API we've interacted with thus far
3) we'll need to ensure that the test environment's secrets are configured with a `STRIPE_ENDPOINT_SECRET` matching the secret used to create the signed webhook request in step (1) - this probably means adding a `STRIPE_ENDPOINT_TEST_SECRET` (name flexible) to the `process.env` of the integration test and ensuring that the two match. We'll want to be careful to avoid leaking secrets if possible here.
4) we'll need to ensure that the `REQUIRE_PAYMENT_PLAN` environment variable is set to `true` in the integration environment.

PR #268, opened 2023-11-07
feat: require payment plans to provision spaces

travis:
Most of the heavy lifting here is done by upgrading the `@web3-storage/upload-api` dependency. There is a new "feature flag" in the context that changes whether payment plans are required for provisioning - I'm going to set it to `false` for now in production and we can flip it once we're ready to turn this on.

I spent a bit of time working toward extending our integration tests to exercise the Stripe webhook handler, but it turned into a big enough project that I decided to defer and recorded what I learned in an issue here:

https://github.com/web3-storage/w3infra/issues/267


.env.tpl
```diff
@@ -38,3 +38,5 @@ SATNAV_BUCKET_NAME = ''
 MAILSLURP_API_KEY = ''
 MAILSLURP_TIMEOUT = '120000'
 
+# Feature flags
+REQUIRE_PAYMENT_PLAN = 'true'
```
vasco-santos, line 42: would be good to test this in staging. I am not 100% sure if swapping env var is enough, or we need to re-deploy. So knowing that can help us know what we need to do

vasco-santos (approved):
changes look good to me. We just need to see what is going on in inetgration tests by upgrading the `upload-api`

upload-api/test/helpers/ucan.js
```diff
@@ -295,6 +295,7 @@ export async function executionContextToUcantoTestServerContext (t) {
 
   return {
     ...serviceContext,
+    grantAccess: async () => {},
```
alanshaw:
```suggestion
```

alanshaw (approved):

upload-api/test/helpers/ucan.js
travis: oh lol we were looking at this at precisely the same time - fixed, thx!

PR #269, opened 2023-11-08
wip: demonstrate integration test failure with latest clients

travis:
One of the recent changes to auth flows broke the integration tests. I suspect https://github.com/web3-storage/w3up/pull/1059 is the root cause, but could be wrong - I believe there have been a couple recent changes to authorization logic.


travis:
closing for now to free up IAMs

PR #275, opened 2023-11-13
feat: add receipts endpoint

vasco-santos:
Adds receipts endpoint with redirect to workflow bucket where receipt for asked task will be available.

Bucket was made public to accept GETs, you can see the generated policy [here](https://s3.console.aws.amazon.com/s3/buckets/workflow-store-pr275-0?region=us-east-2&tab=permissions) . IAM Role has general access to the bucket, while `"Principal": "*"` can perform `GetObject`.

Integration tests were added to inspect receipt from `filecoin/offer` invocation

### DIFF
--- a/upload-api/buckets/invocation-store.js
+++ b/upload-api/buckets/invocation-store.js
@@ -116,12 +116,12 @@ export const useInvocationStore = (s3client, bucketName) => {
       })
       const listObject = await s3client.send(listObjectCmd)
       const carEntry = listObject.Contents?.find(
-        content => content.Key?.endsWith('.workflow')
+        content => content.Key?.endsWith('.out')
       )
       if (!carEntry) {
         return
       }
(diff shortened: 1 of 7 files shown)


upload-api/buckets/invocation-store.js
```diff
@@ -116,12 +116,12 @@ export const useInvocationStore = (s3client, bucketName) => {
       })
       const listObject = await s3client.send(listObjectCmd)
       const carEntry = listObject.Contents?.find(
-        content => content.Key?.endsWith('.workflow')
+        content => content.Key?.endsWith('.out')
```
vasco-santos, line 119: 😱  this had a bug, but this code path was not being used. See https://github.com/web3-storage/w3infra/blob/main/docs/ucan-invocation-stream.md#buckets as this first was `.workflow` but then was iterated to `.in` and `.out`

alanshaw (approved):
\o/

PR #278, opened 2023-11-14
fix: diffs for done stores

alanshaw:
If an upload is "done", we should only count it as a diff to the space when it's NOT already stored in the space.

### DIFF
--- a/billing/lib/ucan-stream.js
+++ b/billing/lib/ucan-stream.js
@@ -6,9 +6,10 @@ import * as StoreCaps from '@web3-storage/capabilities/store'
  * the delta.
  *
  * @param {import('./api.js').UcanStreamMessage[]} messages
- * @returns {import('./api.js').UsageDelta[]}
+ * @param {{ storeTable: Pick<import('@web3-storage/upload-api').StoreTable, 'exists'> }} ctx
+ * @returns {Promise<import('./api.js').UsageDelta[]>}
  */
-export const findSpaceUsageDeltas = messages => {
+export const findSpaceUsageDeltas = async (messages, ctx) => {
   const deltas = []
   for (const message of messages) {
     if (!isReceipt(message)) continue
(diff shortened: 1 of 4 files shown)


billing/test/lib/ucan-stream.js
```diff
@@ -60,10 +85,40 @@ export const test = {
...
+        exists: async (s, l) => filteredReceipts.some(r => {
+          const out = /** @type {any} */ (r.out.ok)
+          return s.toString() === out.with?.toString() && out.link?.toString() === l.toString()
+        })
+      }
+    })
     assert.equal(deltas.length, receipts.length)
 
-    // ensure we have a delta for every receipt
+    // ensure we have a delta for every receipt, except
```
olizilla, line 121: except...?

billing/lib/ucan-stream.js
```diff
@@ -6,16 +6,23 @@ import * as StoreCaps from '@web3-storage/capabilities/store'
...
   const deltas = []
   for (const message of messages) {
     if (!isReceipt(message)) continue
 
     /** @type {number|undefined} */
     let size
     if (isReceiptForCapability(message, StoreCaps.add) && isStoreAddSuccess(message.out)) {
+      // If status is done, we need to check if the item is stored in _this_
+      // space or not. If it is, then there is no delta.
+      if (message.out.ok.status === 'done') {
```
olizilla, line 22: do we increment when the space is allocated by calling `store/add`? If I keep calling `store/add` but not uploading the thing do i keep getting charged additional increments each time?

olizilla (approved):
Change looks good, just 1 question about how this works if i call store/add succesfully, repeatedly for the same cid, but never manage to upload that cid.

alanshaw:
Hmm, thinking about it I think this doesn't work as I thought....

alanshaw:
Plan is to return something `exists: boolean` in the receipt that would signal to us whether this item was _already_ stored in the space or not.

PR #280, opened 2023-11-15
fix: filecoin stack cron increase timeout

vasco-santos:
Default is 10s, which is unrealistic to go through table, get around 4000 pieces, get receipt chain for each and if there is a deal, update table.


stacks/filecoin-stack.js
```diff
@@ -121,6 +121,7 @@ export function FilecoinStack({ stack, app }) {
           AGGREGATOR_DID,
           PROOF: STOREFRONT_PROOF,
         },
+        timeout: '4 minutes',
```
olizilla, line 124: is there some maths or experiments to back this up, or just bump it up an see?

test/filecoin.test.js
```diff
@@ -202,18 +202,20 @@ test('w3filecoin integration flow', async t => {
...
   console.log(`wait for aggregate/accept receipt ${aggregateAcceptReceiptCid.toString()} ...`)
   await waitForStoreOperationOkResult(
-    () => receiptStoreFilecoin.get(aggregateAcceptReceiptCid),
+    async () => {
+      // Trigger cron to update and issue receipts based on deals
+      const callDealerCronRes = await fetch(`https://staging.dealer.web3.storage/cron`)
+      t.true(callDealerCronRes.ok)
+
+      return receiptStoreFilecoin.get(aggregateAcceptReceiptCid)
+    },
```
olizilla, line 218: this seems unrelated to the PR title... can you explain why change?

stacks/filecoin-stack.js
vasco-santos: we run each 5 minutes, the amount of work to be done depends on the inflow of uploads. Timing out with 10 seconds would not work when we may need to get 1000s of receipts and do 1000s of dynamo ops. Timing out after 4min and restarting the job after will restart the process and will be fine, so there is not really math behind this

test/filecoin.test.js
vasco-santos: Added 

> Tests improved so that when things are still being processed in queue when we trigger cron, we don't await on outcome of cron run that ended up being NOP because queue is still processing.

Not documented before as I was trying this out when PR made...

PR #284, opened 2023-11-21
feat: fetch delegations concurrently

travis:
This is used in `access/claim` and has been throwing errors when there are large numbers of delegations to claim. I think this is because we are fetching them serially, and am hoping that by fetching delegations concurrently we speed that up a lot.

### DIFF
--- a/upload-api/tables/delegations.js
+++ b/upload-api/tables/delegations.js
@@ -121,7 +121,7 @@ export function useDelegationsTable (dynamoDb, tableName, { bucket, invocationBu
       })
       const response = await dynamoDb.send(cmd)
 
-      const delegations = []
+      const delegationPromises = []
       for (const result of response.Items ?? []) {
         const { cause, link } = unmarshall(result)
         const delegationCid = /** @type {Ucanto.Link} */ (parseLink(link))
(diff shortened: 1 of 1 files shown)


upload-api/tables/delegations.js
```diff
@@ -121,36 +121,35 @@ export function useDelegationsTable (dynamoDb, tableName, { bucket, invocationBu
...
-          const result = await cidToDelegation(bucket, delegationCid)
-          if (result.ok) {
-            delegations.push(result.ok)
-          } else {
-            return {
-              error: new Failure(`failed to get delegation ${delegationCid} from legacy delegations bucket`, {
-                cause: result.error
-              })
-            }
+          delegationPromises.push(cidToDelegation(bucket, delegationCid))
```
vasco-santos, line 140: nit: from my experience with R2, and their rate limits, I would suggest wrapping this with a `p-retry` (if error is not 404), or maybe better within the function itself.

vasco-santos (approved):
LGTM, just a small suggestion inline

upload-api/tables/delegations.js
alanshaw: This might even be the cause for the 500 errors when there are lots to retrieve. We shouldn't block this PR on adding that though but it would be a good idea to have in a follow up ;)

upload-api/tables/delegations.js
```diff
@@ -121,36 +121,35 @@ export function useDelegationsTable (dynamoDb, tableName, { bucket, invocationBu
       })
       const response = await dynamoDb.send(cmd)
 
-      const delegations = []
+      const delegationPromises = []
```
alanshaw, line 124: Would be good to retain capturing the problematic CID in the error. Perhaps if this was:

```js
/** @type {Array<{ delegationCid, promise }>} */
const delegationPromises = []
```

You could use later to augment the error message appropriately.

upload-api/tables/delegations.js
```diff
@@ -121,36 +121,35 @@ export function useDelegationsTable (dynamoDb, tableName, { bucket, invocationBu
...
+      for (const result of delegationResults) {
+        if (result.ok) {
+          delegations.push(result.ok)
+        } else {
+          return {
+            error: new Failure(`could not find delegation: ${result.error.message}`, {
+              cause: result.error
+            })
           }
         }
```
alanshaw, line 154: Alternatively turn `response.Items` into the promises to retain the delegation CID:

```suggestion
      const delegationResults = await Promise.all(response.Items.map(async result => {
        // do the stuff in the loop above
        // augmenting the error message with the delegation CID as before
      }))
      const delegations = []
      for (const result of delegationResults) {
        if (result.ok) {
          delegations.push(result.ok)
        } else {
          return result
        }
```

alanshaw (approved):
Changes look fine but I think it would ease debuggering if we retain the delegation CID in the error.

PR #285, opened 2023-11-28
fix: filecoin integration tests more robust

vasco-santos:
Depending on the aggregate buffering and timings in queue, MAY happen that not all pieces are needed to create the aggregate. If this is the case, there is a chance that one piece in the 4 uploaded will not make it to the aggregate, being still in queue for next one. In these integration tests, we just care that one Piece made it, so changed test to instead of expecting that first piece made it, we now do a race for all the pieces, and just expect one to make it


alanshaw (commented):
LGTM!

PR #289, opened 2023-11-30
feat: more access/claim improvements

travis:
implement suggestions in PR feedback from https://github.com/web3-storage/w3infra/pull/284

### DIFF
--- a/upload-api/buckets/delegations-store.js
+++ b/upload-api/buckets/delegations-store.js
@@ -69,7 +69,7 @@ export const useDelegationsStore = (s3client, bucketName) => {
         Bucket: bucketName,
         Key: createDelegationsBucketKey(cid),
       })
-      const s3Object = await s3client.send(getObjectCmd)
+      const s3Object = await pRetry(() => s3client.send(getObjectCmd))
       const bytes = await s3Object.Body?.transformToByteArray()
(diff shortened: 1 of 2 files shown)


gobengo (approved):

upload-api/buckets/delegations-store.js
```diff
@@ -69,7 +69,7 @@ export const useDelegationsStore = (s3client, bucketName) => {
         Bucket: bucketName,
         Key: createDelegationsBucketKey(cid),
       })
-      const s3Object = await s3client.send(getObjectCmd)
+      const s3Object = await pRetry(() => s3client.send(getObjectCmd))
```
vasco-santos, line 72: I assume it is possible to have 404? If so, we should have something more in the lines of

```js
    const s3Object = await pRetry(async () => {
    let res
      try {
        res = await this.client.send(cmd)
      } catch (cause) {
        if (cause?.$metadata?.httpStatusCode === 404) {
          return undefined
        }
        throw new Error(`Failed to check bucket ${this.name}`)
      }
      return res
    }, {
      retries: 3
    })
```

vasco-santos (changes requested):

vasco-santos (approved):

PR #290, opened 2023-11-30
feat: filecoin metrics for ucan stream prometheus and updates metrics endpoint

vasco-santos:
Adds filecoin metrics for ucan stream in the same pattern we track other metrics in https://up.web3.storage/metrics

We want to rely on UCAN Stream metrics table to in a "prometheus style" increment metrics over time for totals.  There are some metrics in Firehose that we can rely on for ad-hoc queries and business queries (like monthly usage). More than the price and efficiency in the long run for totals in the Firehose, there are some limitations that are explicit here:

- we use PieceCids to encode Piece size, which we will never be able to handle in Athena level as we cannot decode the CID to get the size
- UCAN invocations (like aggregate/offer) may rely on attached blocks to move attachments with the invocation, while only identifying the CID for them. E.g in `aggregate/offer` we encode the very large list of pieces to put into the aggregate in a CBOR block, and inline it in the CAR of the invocation, while providing `pieces` as CID of the block in the nb object.

We will need most measures I wanted to handle, except for tracking execution time between the invocations. We need https://github.com/web3-storage/w3up/issues/970 to get the power to track execution time of a UCAN effect (e.g. time between receipt of `aggregate/offer` and `aggregate/accept`

### DIFF
--- a/ucan-invocation/types.ts
+++ b/ucan-invocation/types.ts
@@ -2,6 +2,16 @@ import type { DID, Link } from '@ucanto/interface'
 import { ToString, UnknownLink } from 'multiformats'
 import { Ability, Capability, Capabilities } from '@ucanto/interface'
 
+export interface FilecoinMetricsStore {
+  incrementTotal: (metricName: string, n: number) => Promise<void>
+  incrementTotals: (metricsToUpdate: Record<string, number>) => Promise<void>
+}
+
+export interface FilecoinMetricsCtx {
+  filecoinMetricsStore: FilecoinMetricsStore
+  workflowStore: WorkflowBucket
+}
+
 export interface MetricsTable {
   incrementStoreAddTotal: (incrementSizeTotal: Capability<Ability, `${string}:${string}`, unknown>[]) => Promise<void>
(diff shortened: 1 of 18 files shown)


ucan-invocation/types.ts
```diff
@@ -48,16 +62,17 @@ export type UcanInvocationType = 'workflow' | 'receipt'
 
 export interface UcanInvocation {
   carCid: string
+  invocationCid: string
```
vasco-santos, line 65: in https://github.com/web3-storage/w3infra/blob/9897890c65968f7d47a4b3b520f736863343642b/upload-api/ucan-invocation.js#L169

alanshaw: Yes but, only for receipts.

I ended up re-typing these definitions in `billing` to more accurately reflect what's there and not get errors for properties that don't always exist:

https://github.com/web3-storage/w3infra/blob/9897890c65968f7d47a4b3b520f736863343642b/billing/lib/api.ts#L210-L240

alanshaw (approved):
LGTM

travis (approved):

PR #295, opened 2023-12-04
feat: glue table for dealer aggregate caps

vasco-santos:
Adds glue table for dealer `aggregate/offer` and `aggregate/accept` capabilities.

My initial plan after talking with @travis was to simply rely on [receipts table](https://github.com/web3-storage/w3infra/blob/3aee776405c195e3ff7b0ac5eb9ec737e9d5dbbc/stacks/firehose-stack.js#L183-L229) to grab Athen metrics for times for an aggregate to be dealt after being offered.

However, due to the dag-json encoding for the receipt value in these invocations while querying receipts, it won't be possible to fetch the CID of the aggregate. Which is needed, so that we can build a query where we select `aggregate/accept` and join them with `aggregate/offer` to get to know the timestamp differences.

```json
{"aggregate":{"/":"bafkzcibcaapphztgcff34xlqsvyudubnagnbz2ltlcvdarala5tkuhvm7j6kiey"}}
```

With this PR, we do the same approach as other tables like [`storeAdd`](https://github.com/web3-storage/w3infra/blob/3aee776405c195e3ff7b0ac5eb9ec737e9d5dbbc/stacks/firehose-stack.js#L235-L278) where we configure serialization to be able to grab the aggregate in a query. We cannot do this at the receipt table level, as we need to be sure about the format we expect.

### DIFF
--- a/stacks/firehose-stack.js
+++ b/stacks/firehose-stack.js
@@ -473,6 +473,104 @@ export function UcanFirehoseStack ({ stack, app }) {
   })
   providerAddTable.addDependsOn(glueDatabase)
 
+  // creates a table that can be seen in the AWS Glue table browser at 
+  // https://console.aws.amazon.com/glue/home#/v2/data-catalog/tables
+  // and in the data browser in the Athena Query editor at
+  // https://console.aws.amazon.com/athena/home#/query-editor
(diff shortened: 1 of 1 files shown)


alanshaw (approved):
I can't really authoritatively review as I'm not familiar with this code.

PR #303, opened 2023-12-07
feat(billing): use allocated bytes in store/add receipt

alanshaw:
Uses `allocated` property in the `store/add` receipt to count bytes added to the space. The `allocated` property will be 0 when the user tries to add a shard that is already present in the space.

❓ should we store a diff anyway? even though the size did not change?

refs https://github.com/web3-storage/w3up/pull/1213


vasco-santos (approved):
LGTM

PR #309, opened 2023-12-18
fix: sort key for space diffs

alanshaw:
🙈 🙈 🙈 

We currently have sort keys like:

```
Wed Nov 15 2023 16:23:16 GMT+0000 (Coordinated Universal Time)#bafyreiaaqkc6e2rcij7tipadffmqrdfcwbztcms3g7ayktyterfascn3xm
```

...which does not sort as expected

### DIFF
--- a/billing/data/space-diff.js
+++ b/billing/data/space-diff.js
@@ -28,7 +28,7 @@ export const encode = input => {
     return {
       ok: {
         pk: `${input.provider}#${input.space}`,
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):
lgtm

what is the plan here? re-run the billing job and clear the DB?

alanshaw:
I will migrate the existing diffs (add new, delete old)

PR #313, opened 2023-12-22
feat: roundabout with content claims

vasco-santos:
Makes roundabout use content claims by default. It looks for location claims for requested CAR and attempts to return to a presigned URL for an object in a R2 bucket. In case there are multiple location claims, they are filtered by R2 URLs. If there are no R2 URLs for configured buckets, fallbacks to find S3 URLs configured (e.g. `carpark` in S3 that also has CARs in R2) and attempts sibling bucket in R2.

We can also in the future easily make these bucket names coming from ENV vars. Given they need to be tied with CF Access key (and other ENVs), we should have a list of buckets we are willing to have roundabout looking for.

Notes:
- we are currently fallbacking on no location claims found, to attempt `carpark-prod-0`
- kept key route with bucket parameter for the time being. This is heavily used by dagcargo created aggregates and we should keep it until we stop dagcargo deals and GC dagcargo R2 bucket

### DIFF
--- a/roundabout/functions/redirect.js
+++ b/roundabout/functions/redirect.js
@@ -2,9 +2,18 @@ import * as Sentry from '@sentry/serverless'
 import { S3Client } from '@aws-sdk/client-s3'
 import { CID } from 'multiformats/cid'
 
-import { getSigner } from '../index.js'
-import { findEquivalentCarCids, asPieceCidV1, asPieceCidV2, asCarCid } from '../piece.js'
-import { getEnv, parseQueryStringParameters } from '../utils.js'
+import {
+  getSigner,
+  carLocationResolver,
+  resolveCar,
+  resolvePiece,
+  redirectTo
+} from '../index.js'
+import {
+  getEnv,
+  parseQueryStringParameters,
+  parseKeyQueryStringParameters,
+} from '../utils.js'
 
 Sentry.AWSLambda.init({
   environment: process.env.SST_STAGE,
@@ -13,11 +22,16 @@ Sentry.AWSLambda.init({
 })
 
 /**
- * AWS HTTP Gateway handler for GET /{cid} by CAR CID or Piece CID
+ * AWS HTTP Gateway handler for GET /{cid} by CAR CID or an equivalent CID,
+ * such as a Piece CID.
  *
  * @param {import('aws-lambda').APIGatewayProxyEventV2} request
  */
 export async function redirectCarGet(request) {
+  const {
+    BUCKET_NAME,
+  } = getEnv()
+
   let cid, expiresIn
   try {
     const parsedQueryParams = parseQueryStringParameters(request.queryStringParameters)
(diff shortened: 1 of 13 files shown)


roundabout/functions/redirect.js
```diff
@@ -29,7 +39,6 @@ export async function redirectCarGet(request) {
   }
 
   const locateCar = carLocationResolver({ 
-    bucket: getEnv().BUCKET_NAME,
     s3Client: getS3Client(),
     expiresIn
   })
```
vasco-santos, line 44: we are currently not fallbacking on no location claims found, to attempt carpark-prod-0, probably we should?

roundabout/utils.js
```diff
@@ -3,7 +3,74 @@ export const MAX_EXPIRES_IN = 3 * 24 * 60 * 60 // 7 days in seconds
...
+  const validR2Buckets = options.validR2Buckets || VALID_R2_BUCKETS_DEFAULT
+  const validS3Buckets = options.validS3Buckets || VALID_S3_BUCKETS_DEFAULT
+  // Filter by Cloudflare R2 URLs
+  const r2Urls = Array.from(locations)
+    .filter(
+      // CF Domain
+      l => l.includes(CF_R2_DOMAIN) &&
+      // Bucket name valid for CF
+      validR2Buckets.filter(b => l.includes(b)).length
+    )
```
olizilla, line 31: lets tighten this up to map the locations to URLs and match on the hostname and path explicitly rather than using string.includes. If the plan is to let anyone post a claim, then we have to be defensive when processing them.

roundabout/utils.js
```diff
@@ -3,7 +3,74 @@ export const MAX_EXPIRES_IN = 3 * 24 * 60 * 60 // 7 days in seconds
...
+      validS3Buckets.filter(b => l.includes(b)).length
+    )
+  
+  // Transform S3 URLs if existent
+  if (s3Urls.length) {
+    return s3Urls.map(url => {
+      // Format 'https://bucket-name.s3.amazonaws.com/key'
+      const domainParts = url.split(`.${AWS_S3_DOMAIN}`)
+      const bucketName = domainParts[0].replace('https://', '')
+      const key = domainParts[1].slice(1)
```
olizilla, line 63: same here, let's use URLs and be explicit.... `3.amazonaws.com` would be a valid s3 key name i think. Even if not, a user could put anything in a location claim url

vasco-santos: I mean it would work as a key, but not as a CID later on? so for this to work would never be a valid key

roundabout/utils.js
vasco-santos: added the `account-id` in R2. However, it would actually make sense to spec how we plan to send these URLs for R2 as they vary quite a lot in R2 docs...

vasco-santos: I am not sure if we can actually do this nicely with S3 given bucket name is in the hostname and we can have several. I think we will need to sync on this

vasco-santos: see https://github.com/web3-storage/w3infra/issues/316

travis:
@hannahhoward @alanshaw closing for now to clean up PR environments but happy to re-open

PR #314, opened 2023-12-30
fix: ensure integer quantity

alanshaw:
Quantity must always be an integer, otherwise it is recorded as 0.


vasco-santos (approved):
lgtm

alanshaw:
Note: this is an speculative PR and I'm holding off on merging until I can verify.

PR #318, opened 2024-01-23
feat: update Stripe plan in `PlansStorage#set`

travis:
We'd like to let users change their plan directly from console or the CLI, which means we need to teach `PlansStorage#set` how to update Stripe.

I've introduced a new `BillingProvider` interface to start abstracting Stripe functionality and especially to make it easier to test systems that interact with Stripe. I have not done a thorough audit of the various ways we interact with Stripe, but I'd like to move toward having all Stripe interactions go through `BillingProvider`.

The linter is currently failing because I made the `account` field in the the billing customer schema optional and a fair amount of the billing logic relies on it being not-`undefined`. I could just go in and update logic to deal with undefined `account` but wanted to get some feedback from @alanshaw before I dig into that: I can see arguments for either option (roughly "we might want to track customers who don't have an account set up in Stripe" vs "the point of this table is to track Stripe accounts") but testing will require a bit more work if we want to preserve non-optionality (since the test stores don't get initialized from the test billing provider in the same way - in production they get initialized via the webhook handler, which doesn't currently have an analog in testing code). Either of the options here will require a bit more work, so I'd love to know if anyone has strong feelings before I finish this off.


billing/tables/customer.js
```diff
@@ -56,8 +56,8 @@ export const createCustomerStore = (conf, { tableName }) => ({
         Key: marshall({ customer }),
         UpdateExpression: 'SET product = :product, updatedAt = :updatedAt',
         ExpressionAttributeValues: marshall({
-          product,
-          updatedAt: new Date().toISOString()
+          ':product': product,
+          ':updatedAt': new Date().toISOString()
```
alanshaw, line 60: Ooo is that an existing bug?

upload-api/billing.js
```diff
@@ -0,0 +1,89 @@
...
+        if (subscriptions?.length != 1) return (
+          { error: new InvalidSubscriptionState(`found ${subscriptions?.length} Stripe subscriptions(s) for customer with email ${email} - cannot set plan`) }
+        )
+
+        subscriptionItems = customer.subscriptions?.data[0].items.data
+        if (subscriptionItems?.length != 1) return (
+          { error: new InvalidSubscriptionState(`found ${subscriptionItems?.length} Stripe subscriptions item(s) for customer with email ${email} - cannot set plan`) }
+        )
+      } catch (/** @type {any} */ err) {
+        return { error: new InvalidSubscriptionState(err.message, { cause: err }) }
```
alanshaw, line 77: It seems we've handled all expected invalid states - if any of these calls to stripe throw it's because of a network or config error or something. I would just not catch this and let it be thrown (and caught and handled by ucanto).

upload-api/billing.js
```diff
@@ -0,0 +1,89 @@
...
+      }
+
+      try {
+        await stripe.subscriptionItems.update(subscriptionItems[0].id, { price: priceID })
+
+        return { ok: {} }
+      } catch (/** @type {any} */ err) {
+        return { error: new BillingProviderUpdateError(err.message, { cause: err }) }
```
alanshaw, line 85: I don't know if it's worth wrapping a completely unexpected error...personally I'd let this throw.

upload-api/stores/plans.js
```diff
@@ -22,7 +25,21 @@ export function usePlansStore(customerStore) {
...
+      try {
+        const hasCustomerResponse = await billingProvider.hasCustomer(customer)
+        if (hasCustomerResponse.ok) {
+          await billingProvider.setPlan(customer, plan)
+          return await customerStore.updateProduct(customer, plan)
+        } else {
+          if (hasCustomerResponse.error) {
+            return hasCustomerResponse
+          } else {
+            return { error: new Failure(`billing provider does not have customer for ${customer}`) }
```
alanshaw, line 37: So this would maybe be a `BillingCustomerNotFoundError` that we can use in the types for `PlansStorage`?

upload-api/stores/plans.js
```diff
@@ -22,7 +25,21 @@ export function usePlansStore(customerStore) {
...
+          }
+        }
+      } catch (/** @type {any} */ err) {
+        return { error: err }
```
alanshaw, line 41: Just let it throw...this is what ucanto will do for us essentially.

alanshaw (changes requested):
Please can you explain why `account` needs to become optional? I don't see anything in this PR that requires it.

travis:
> Please can you explain why account needs to become optional?

For sure - tried to explain this in the PR description but it was end-of-day stream of consciousness.

Tests for `PlansStorage` (which live in [`w3up`](https://github.com/web3-storage/w3up/blob/main/packages/upload-api/test/storage/plans-storage-tests.js)) don't currently do any setup, which means there isn't a record in the test Dynamo when we invoke `plan/set`. Since that doesn't currently require a full record (ie, doesn't error when it doesn't find a record) I was seeing errors in the test when it subsequently invoked `plan/get`.

Taken on its own, this suggests to me that we should rework the test and perhaps make `plan/set` fail if there isn't already a record in Dynamo, but it also made me wonder if account _should_ be optional - it seems like we might want to track billing for a customer even if they have not yet set up a Stripe account and allow them to connect a Stripe account later. 

It's entirely possible that this case is already supported without making `account` optional, so I figured I'd pause and check in with you on the intent behind making `account` non-optional. Would love to huddle this morning to kick this around if you have a few minutes!

upload-api/billing.js
travis: Oh I didn't know Ucanto had default handling for errors here, to be honest. A couple thoughts:

1) I'm not sure this `BillingProvider` will always be used in Ucanto - I'm using the `Result` type here because @Gozala has been advocating for it but otherwise this is something I'd like to use, eg, in the Stripe webhook where top level error handling will be different
2) I've found it's annoying when trying to call a function that might either return an `{error: ...}` response OR throw an exception - I think @Gozala has also been advocating for trying to make sure all errors in functions that return `Result`s are caught and turned into typed `{error: ...}` style responses. I could totally be going overboard here! Curious for your thoughts @Gozala!

upload-api/stores/plans.js
travis: ah yep - I want to replace all the `Failure`s in this PR with typed error responses like that, great catch!

billing/tables/customer.js
travis: it is! the `PlanStorage` tests from `w3up` should have caught this but they weren't being run, since the test runner was not added before this PR

upload-api/billing.js
alanshaw: We talked about this async and came up with this: https://github.com/web3-storage/RFC/pull/7

upload-api/test/helpers/billing.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export function createTestBillingProvider() {
+  /**
+   * 
+   * Initialize this with data that matches the test defined in @web3-storage/w3up.
+   * 
+   * Normally this will be set up in the billing provider itself (in the 
+   * current implementation this means sending them to Stripe) so we 
+   * initialize it with values that will allow the test to pass.
+   *
```
alanshaw, line 12:
```suggestion
```

alanshaw (approved):

travis:
created https://github.com/web3-storage/w3infra/pull/320 because the integration environment for this PR is broken - integration tests pass there!

PR #319, opened 2024-01-31
fix: pull secrets from SST correctly

travis:
I was not doing it correctly, which is I think why the integration tests are failing


travis:
whoop - this should have been added to #318

PR #320, opened 2024-01-31
feat: update Stripe plan in `PlansStorage#set`

travis:
We'd like to let users change their plan directly from console or the CLI, which means we need to teach PlansStorage#set how to update Stripe.

I've introduced a new BillingProvider interface to start abstracting Stripe functionality and especially to make it easier to test systems that interact with Stripe. I have not done a thorough audit of the various ways we interact with Stripe, but I'd like to move toward having all Stripe interactions go through BillingProvider.


travis:
I created this PR since the integration test env for https://github.com/web3-storage/w3infra/pull/318 seems to be broken - integration tests run fine here but consistently fail there

upload-api/billing.js
```diff
@@ -0,0 +1,89 @@
...
+      try {
+        const prices = await stripe.prices.list({ lookup_keys: [plan] })
+        priceID = prices.data.find(price => price.lookup_key === plan)?.id
+        if (!priceID) return (
+          { error: new InvalidSubscriptionState(`could not find Stripe price with lookup_key ${plan} - cannot set plan`) }
+        )
+
+        const email = toEmail(/** @type {import('@web3-storage/did-mailto').DidMailto} */(customerDID))
+        const customers = await stripe.customers.list({ email, expand: ['data.subscriptions'] })
+        if (customers.data.length != 1) return (
```
gobengo, line 62:
```suggestion
        if (customers.data.length !== 1) return (
```

Just curious, but were all the non-triple-equality checks in here on purpose? (e.g. because 'length' is a string or something?)

upload-api/stores/plans.js
```diff
@@ -1,12 +1,42 @@
...
  * @param {import("@web3-storage/w3infra-billing/lib/api").CustomerStore} customerStore
+ * @param {import('../types.js').BillingProvider} billingProvider
  * @returns {import("@web3-storage/upload-api").PlansStorage}
  */
-export function usePlansStore(customerStore) {
+export function usePlansStore(customerStore, billingProvider) {
   return {
-    get: async (customer) => {
-      const result = await customerStore.get({ customer })
+    initialize: async (account, externalID, plan) => {
```
gobengo, line 9: At first glance, this logic could be subject to a race condition. i.e. if two request come in in parallel and do `customerStore.get` and get a result indicating the customer does not exist, then both requests will move on to calling `customerStore.put` with similar values.

It would be helpful for a comment to acknowledge the seeming race condition and, if there are any mitigations, explain them. If there are no mitigations and it's just a known race condition, also good to have a comment for that.
(a similar race condition affected last stripe integration)

upload-api/billing.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * 
+ * @param {import('stripe').Stripe} stripe 
+ * @returns {import("./types").BillingProvider}
+ */
+export function createStripeBillingProvider(stripe) {
+  return {
+    async hasCustomer(customer) {
+      const customersResponse = await stripe.customers.list({ email: toEmail(/** @type {import('@web3-storage/did-mailto').DidMailto} */(customer)) })
```
gobengo, line 44: this layer should probably *normalize* these email addresses before reading/writing.

e.g. lowercase the domain part of the email address before passing to stripe.
https://stackoverflow.com/questions/9807909/are-email-addresses-case-sensitive

(ideally stripe would do this for you but I wouldn't count on it)

travis: I wonder if we should also normalize the part before the `@` for email providers (like gmail, outlook and yahoo) that treat it as case-insensitive: https://www.linkedin.com/pulse/unraveling-mystery-email-addresses-case-sensitive-warmy#:~:text=Gmail's%20Approach%20to%20Case%20Sensitivity.&text=This%20means%20that%20regardless%20of,email%20address%20in%20Gmail's%20system.

for gmail I think this would also mean stripping out `.` from that part of the email address: https://support.google.com/mail/answer/7436150?hl=en

upload-api/billing.js
travis: no just a mistake on my end! will review and fix them all

upload-api/billing.js
```diff
@@ -0,0 +1,89 @@
...
+
+        const email = toEmail(/** @type {import('@web3-storage/did-mailto').DidMailto} */(customerDID))
+        const customers = await stripe.customers.list({ email, expand: ['data.subscriptions'] })
+        if (customers.data.length != 1) return (
+          { error: new InvalidSubscriptionState(`found ${customers.data.length} Stripe customer(s) with email ${email} - cannot set plan`) }
+        )
+
+        const customer = customers.data[0]
+        const subscriptions = customer.subscriptions?.data
+        if (subscriptions?.length != 1) return (
```
travis, line 68:
```suggestion
        if (subscriptions?.length !== 1) return (
```

upload-api/billing.js
```diff
@@ -0,0 +1,89 @@
...
+          { error: new InvalidSubscriptionState(`found ${subscriptions?.length} Stripe subscriptions(s) for customer with email ${email} - cannot set plan`) }
+        )
+
+        subscriptionItems = customer.subscriptions?.data[0].items.data
+        if (subscriptionItems?.length != 1) return (
```
travis, line 73:
```suggestion
        if (subscriptionItems?.length !== 1) return (
```

upload-api/stores/plans.js
travis: ah true! added a comment: 

```js
      // Note that there is a possible race condition here - if two `initialize` calls come in at
      // the same time, they may both get here since the account hasn't been initialized yet.
      // In that case, the last `put` will win, which should be fine - as far as I can tell
      // there's no way we can really determine which one was "valid" so accepting whichever of them
      // happens to go last should be fine.
```

gobengo (approved):

billing/tables/customer.js
```diff
@@ -56,8 +56,8 @@ export const createCustomerStore = (conf, { tableName }) => ({
         Key: marshall({ customer }),
         UpdateExpression: 'SET product = :product, updatedAt = :updatedAt',
         ExpressionAttributeValues: marshall({
-          product,
-          updatedAt: new Date().toISOString()
+          ':product': product,
```
vasco-santos, line 59: good catch!

upload-api/types.ts
```diff
@@ -264,4 +264,19 @@ declare module '@serverless-stack/node/config' {
   }
 }
```
vasco-santos, line 266: here on top should come type for STRIPE_SECRET_KEY right?

upload-api/test/billing.test.js
```diff
@@ -0,0 +1,96 @@
...
+      payment_method: 'pm_card_visa',
+    }
+  )
+  const paymentMethod = /** @type {string} */(setupIntent.payment_method)
+  await stripe.customers.update(customer.id, { invoice_settings: { default_payment_method: paymentMethod } })
+  return customer
+}
+
+test('stripe plan can be updated', async (t) => {
+  const stripeSecretKey = process.env.STRIPE_TEST_SECRET_KEY
```
vasco-santos, line 52: so this means test is dependent on a real deployment of stripe? Perhaps I would put this into the "integration" tests, so that we also run it in Post deploy? 
What do you think? I usually like to avoid to have tests that require external things, but also would like to have Stripe somehow tested

travis: ah, that's a good call! it'll require a bit of refactoring to make this work, so I've created a separate issue to track this refactoring: https://github.com/web3-storage/w3infra/issues/324

upload-api/types.ts
travis: ah good call! done!

PR #322, opened 2024-02-05
docs: ucan stream new diagram

vasco-santos:
Adds new diagram with Glue and Athena, removing the speculative Redshift from before. Also changes log bucket out of Glacier for now


vasco-santos:
merging in to remove stacks as we ran out of labels

PR #325, opened 2024-02-07
wip: first draft of UCAN bridge

travis:
To support users in languages that do not have existing UCAN invocation implementations, we are going to launch a bridge that allows them to make simple HTTP requests with JSON bodies that we transform into proper UCAN invocations.

So far this PR has:

1) an untested (but type-checking!) implementation of such a bridge 2) a markdown description of the bridge protocol, intended to be the first draft of an eventual specification

Notable design choices:

1) I chose to include JUST the base64pad-encoded "secret" in the Authorization header to avoid running afoul of maximum header size restrictions that exist in [some HTTP environments](https://stackoverflow.com/questions/686217/maximum-on-http-header-values). 2) The `proof` field of the JSON body is a base64pad encoded "delegation archive" (created with `ucanto`'s `Delegation.archive` function)

TODO
- [ ] factor core bridge logic out to a separate library
- [ ] factor HTTP input wrangling out to a separate function
- [ ] rename `UPLOAD_API_DID` and `ACCESS_SERVICE_URL` environment variables to `W3UP_SERVICE_DID` and `W3UP_SERVICE_URL`
- [ ] add tests
- [ ] expand and formalize bridge specification, move it to the specs repo (?)


upload-api/BRIDGE.md
```diff
@@ -0,0 +1,48 @@
...
+## Summary
+
+We have implemented a "bridge" that allows w3up users to interact with the service
+without implementing the UCAN invocation wire protocols. 
+
+A user can submit an HTTP request like (simplified for clarity):
+
+```
+POST /bridge
+Authorization: MYmI4NWUwMGFhNzNlZDlkM2Y2NDYxYWEwZjk1NDdjOWY=
```
gobengo, line 12: wdyt of implementing this instead? https://github.com/ucan-wg/ucan-http-bearer-token

alanshaw: If we use UCAN bearer token we have to know the inputs, create and sign a UCAN prior to the request. The point of the bridge is to allow a client that does not know how to do this to still use the service.

upload-api/BRIDGE.md
```diff
@@ -0,0 +1,48 @@
...
+}
```
alanshaw, line 22: Can we please support multiple invocations?

upload-api/BRIDGE.md
```diff
@@ -0,0 +1,48 @@
...
+A user can submit an HTTP request like (simplified for clarity):
+
+```
+POST /bridge
+Authorization: MYmI4NWUwMGFhNzNlZDlkM2Y2NDYxYWEwZjk1NDdjOWY=
+
+{
+  "ability": "store/add",
+  "resource": "did:key:z6Mkm5qHN9g9NQSGbBfL7iGp9sexdssioT4CzyVap9ATqGqX",
+  "inputs": {
```
alanshaw, line 17: I'd probably map these to [invocation spec](https://github.com/ucan-wg/invocation) field names.

alanshaw (changes requested):
❤️ love it - thanks for pushing this!

Worth mentioning what you're doing is https://github.com/web3-storage/RFC/pull/1#issuecomment-1901018099

Can you please define response format?

gobengo:
> The proof field of the JSON body is a base64pad encoded

Originally I read this as meaning it was base64 encoded with padding. https://datatracker.ietf.org/doc/html/rfc4648#section-4

It could be good to clarify you mean it is *multibase* encoded using the multibase encoding for base64+padding. https://www.ietf.org/archive/id/draft-multiformats-multibase-07.html#name-base-64-with-padding-and-mi

upload-api/BRIDGE.md
gobengo: Sure, this gateway may not technically match one requirement in that spec https://github.com/ucan-wg/ucan-http-bearer-token?tab=readme-ov-file#332-insufficient-capability-scope

But other than that enough is similar that it seems like we could reuse the syntax and a lot of other thinking in that spec, e.g.:
* using a syntactically valid HTTP Authorization header instead of the one here that lacks an `<auth-scheme>`
* a way of dealing with large UCAN chains that is different than what this PR does, but unlike this PR keeps the authorization orthogonal from the request body (in headers) and not tightly coupled to any one request body media type (in this case JSON).
* encodes the UCAN as its canonical JWT encoding instead of a multibased car file. I thought this feature was motivated by people not wanting to have to deal with new encoding formats compared to old.web3.storage, and the UCAN signatures are over the canonical JWT, so it seems ideal to avoid novel encoding formats here and use a JWT like old.web3.storage does. It's also how the UCAN spec says is the canonical way of encoding UCANs ('UCANs MUST be formatted as JWTs').
* I'm not that attached to anyone solution here. But partially I wanted to link to that spec bcause @Gozala's writeup you linked to below says "They POST to the RPC gateway endpoint passing UCAN delegation as JWT bearer token" so passing something other than a bearer token, and other than a JWT, felt like it was deviating from that plan as described.

If this goes down the path of not reusing that fairly well defined way of authorizing http requests using UCANs and invents a new one, please at least add an [`<auth-scheme>`](https://www.rfc-editor.org/rfc/rfc9110#name-credentials) to this Authorization header value (e.g. `Bearer`) even if the rest of the value is not a UCAN JWT.

gobengo: as an aside, we can put the secret seed into a UCAN as a UCAN fct like 'did:zAlice derives from this KDF and this seed'. I can't think of a reason we need to keep them separate?

travis: 1) +1 to adding an <auth-scheme> - definitely need to do this
2) I'm hesitant to use the `ucan-http-bearer-token` spec because it seems to implicitly limit the complexity of UCANs that can be used to authorize a request - the spec even notes that "the limit of [the size of an Authorization header in] most browsers [is] 8KB." I'm open to an argument that 8KB is a high enough limit that we should just accept it, but it imho it feels bad to bake this limit into our bridge without a compelling benefit (and imho "compliance with a spec" is not particularly compelling on its own in this case).
3) As @alanshaw mentioned above, we're trying to avoid forcing users of this bridge to implement UCAN signing, and I think that's directly at odds with using the ucan-http-bearer-token spec - in other words, using that spec feels like a total non-starter given the goals of this project - am I misunderstanding?
4) @bengo what you're proposing is a pretty different direction than what I've proposed - I am suggesting we only put the "identity" into this header and you're proposing we try to get the full authorization chain into it. In your proposed solution, where do you imagine the "identity" being passed? HTTP Basic auth does this by :-separating the "username" and "password" - do you think that would be a good idea or do you have other ideas about how this could work?

travis: whoop - worth noting that you answered (4) above! 

One design goal that I haven't stated anywhere afaik is to keep the really sensitive secret token isolated to the Authorization header - I think there's less of a chance intermediaries (or our systems) will log the Authorization header somewhere than the body - put another way, I'd expect the body of this request to end up in something equivalent to our UCAN log, and that's much worse if it contains both the private key AND the delegation UCAN(s) than if it just contained the delegation. 

If we do decide to put both the secret AND the delegation in the Authorization field then I'm much less concerned, and what you're proposing could work, though (I think) it would mean that we'd mix up the UCAN invocation inputs and the secret key, and would need to be diligent to filter out the key before we put the inputs into the invocation that we create and execute...

travis: Oh - reading the `ucan-http-bearer-token` spec just now I realized that it doesn't really specify whether the UCAN in the bearer token should be an invocation or a delegation. I do think we could use that spec to pass delegations! We'd just need to figure out where we'd like the `secret` to live. I'm still hesitant to go down this path because of the 8kb limit, but that is definitely the least strong of my objections above!

@Gozala would definitely love your thoughts here!

travis:
> ❤️ love it - thanks for pushing this!

<3 

> Worth mentioning what you're doing is [web3-storage/RFC#1 (comment)](https://github.com/web3-storage/RFC/pull/1#issuecomment-1901018099)

Yes! Though there's a pretty big difference in the authorization - I'm not "passing UCAN delegation as JWT bearer token" because of the 8KB limit I mention in the thread with Ben above, instead passing the secret there and embedding the delegation in the JSON body.

> Can you please define response format?

Yep definitely - added a TODO!

upload-api/BRIDGE.md
travis: I'm definitely open to this, though curious whether we could just rely on HTTP features (ie https://en.wikipedia.org/wiki/HTTP_persistent_connection) to make multiple requests fast - is there a particular reason you'd want to support "concurrency" at this level rather than just relying on HTTP?

Worth noting that I think @Gozala also wanted to support multiple invocations here!

upload-api/BRIDGE.md
alanshaw: Thanks for clarifying @gobengo. I had assumed you meant to use it to send an invocation which we wouldn't be able to do, but agreed it makes sense to follow for sending the delegation. I like that it moves the auth data out of the invocation instruction in the body.

We could just use an `X-Auth-Secret` header. Cloudflare for example, uses [headers like this](https://developers.cloudflare.com/api/) in it's own API. I appreciate the comment about not wanting to include this header in logs but in practice `Authorization` is just as susceptible as any other header to being logged.

I'm not keen on putting the secret in the facts of the delegation, I think we _would_ end up logging this UCAN and so would be logging secrets. It also wouldn't allow alice to be delegated permissions from bob without bob knowing alice's secret.

I'm ok with an 8KB limit. If it gets problematic, maybe we send a multipart request with extra UCANs in the body (or something)...but we can implement that if/when we need it.

re: multiple invocations - I assume a single HTTP request translates into a single UCAN invocation to our service. If multiple requests are sent then I think it won't necessarily translate into a single request to our service. i.e. I assume the handler at the bridge would be called for every request on a keepalive connection, causing multiple requests to our service upstream.

@travis WDYT can we move forward with this?

travis: yea after noodling on this I'm also pro-`ucan-http-bearer-token`! agree re: possible mitigation strategies if the `Authorization` header gets too big. 

I'll get this updated, I think this is probably enough shared agreement to move forward with at least another iteration!

upload-api/BRIDGE.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+POST /bridge
+X-Auth-Secret: NGY2YTQ1YjYwNWFiYWU2YWNmYmY4NWFhODc4YjEwYzQ=
+Authorization: Bearer eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCIsInVjdiI6IjAuOS4xIn0.eyJhdHQiOlt7ImNhbiI6InVwbG9hZC9saXN0Iiwid2l0aCI6ImRpZDprZXk6ejZNa3JUblpIRU1aQnYzMjRIMlV5N2N1cjZIR29weXRuZkc4V3RBbzEyTFByQjk0In1dLCJhdWQiOiJkaWQ6a2V5Ono2TWtyczNVbkRZVndVZ2FDWDl5OGdVeGY0c2VKblFxSGE5OWltQkhLa2hiekV5dSIsImV4cCI6MTcwNzUyMzIzNCwiaXNzIjoiZGlkOmtleTp6Nk1ralJ4QmkycDdHelRrTFFRSE5RNGZIY1ExWHQzaVBKVVpxRGVKMnd3UTRlVVUiLCJwcmYiOlsiYmFmeXJlaWQ2dXNwNnZncmprNjRuNXZ6ZGlkZ2gyeW9mbHA0NnRwcmZvdnFwdHozM283eTRvcmxyM3EiXX0.VH09YeLZjT28QpipB4kDRHWdnHq08GiwjlCIaxD2z8XXr5-WC2eR39scKYC8_kAxiRc5EdJ8Vj25hwld2eTyBw
+Content-Type: application/json
+
+{
+  "call": "store/add",
+  "on": "did:key:z6Mkm5qHN9g9NQSGbBfL7iGp9sexdssioT4CzyVap9ATqGqX",
```
Gozala, line 18: Given that UCAN spec is planning to use names `do, sub, args` in places of `call, on, inputs` it might be worth doing the same here.
https://github.com/ucan-wg/invocation/blob/high-level/README.md#invocation-payload

Although it can all change again so 🤷‍♂️

Gozala: One other alternative that was considered at some point which might be a good fit here is following format

```json
{
  "store/add": {
    "did:key:z6Mkm5qHN9g9NQSGbBfL7iGp9sexdssioT4CzyVap9ATqGqX": {
      "link": { "/": "bafybeicxsrpxilwb6bdtq6iztjziosrqts5qq2kgali3xuwgwjjjpx5j24" },
      "size": 42
    }
  }
}
```

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,156 @@
...
+    if (!authorizationHeader.startsWith('Bearer ')) {
+      return {
+        statusCode: 401,
+        body: 'authorization header must use the Bearer directive'
+      }
+    }
+    const jwt = authorizationHeader.replace('Bearer ', '')
+    const delegation = UCAN.parse(jwt)
+
+    // parse body
```
Gozala, line 79: Can we support content type negotiation here ? If content-type was application/json we can parse as such, but we could also accept `dag-cbor` in which case we should parse body as such.

Gozala: `accept` header can be used to decide how to format response.

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,156 @@
...
+    if (
+      (typeof jsonBody.on !== 'string') ||
+      (jsonBody.on.split(':').length < 2)
+    ) {
+      return {
+        statusCode: 400,
+        body: 'on must be a URI'
+      }
+    }
+    const resource = /** @type {import('@ucanto/interface').Resource} */(jsonBody.on)
```
Gozala, line 110: I would say use ucanto `Schema.DID.parse` here, it's never a URI and new spec has also decided to make resource be a DID so that seems like better choice IMO.

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,156 @@
...
+        body: 'request has no authorization header'
+      }
+    }
+    if (!authorizationHeader.startsWith('Bearer ')) {
+      return {
+        statusCode: 401,
+        body: 'authorization header must use the Bearer directive'
+      }
+    }
+    const jwt = authorizationHeader.replace('Bearer ', '')
```
Gozala, line 76: I think jwt assumption here is flawed. JWT can not represent proof chains it will only contain outmost proof and link to the rest. In practice our system will refuse service if proofs aren't included.

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,156 @@
...
+    const delegation = UCAN.parse(jwt)
```
Gozala, line 77: Same as above `UCAN.parse` only understands JWTs and does not support proof chains, which is what typescript is complaining about below. `Delegation` interface in ucanto can represent multiblock DAGs while UCAN codec (or any ipld codec) can only represent single block.

Gozala: I think we should either embrace linked multiheader thing that fission came up with, or simply embed CARs in the authorization header and call it a day. That way you'll be able to use archive / extract functionality in ucanto.

I personally favor CAR option here because it's simpler and I don't expect target audience here to care what they put in HTTP header. CAR version also would be less bytes.

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,156 @@
...
+      })
+    }))
+    const result = receipt.out
+    if (result.ok) {
+      return {
+        statusCode: 200,
+        headers: {
+          "Content-Type": "application/json"
+        },
+        body: JSON.stringify(result.ok)
```
Gozala, line 137: Please use dag-json codec here not JSON.stringify as has no knowledge of IPLD links. On the other hand maybe we don't care about the receipt structure been correct ?

If we do not use dag-json codec receipt signature will not appear valid as payload been signed is not the payload that will come out from `JSON.stringify`.

Gozala (changes requested):

upload-api/BRIDGE.md
gobengo: > Given that UCAN spec is planning to use names do, sub, args in places of call, on, inputs it might be worth doing the same here.

let's do both and identify which the client means via a media type...

upload-api/functions/bridge.js
gobengo: > simply embed CARs in the authorization header and call it a day

my understanding is we'd love to have a plan for how to share UCANs whose CAR serialization is greater than the infra providers http header size limit of ~16kb.

but if we can punt on that, sure why not `Authorization: multiformats/cid <cid-of-ucan-car-identity>`

upload-api/BRIDGE.md
vasco-santos: yeah, also I think for DX of user will be easier to send all in the payload

travis: great, I'm in! I'll make it an array of "invocations" - @gobengo any thoughts on how this might work in light of the content negotiation idea? maybe we should handle multiple invocations with multipart? if we do want to go that route, maybe we could skip multiple invocations in this iteration - handle 1 invocation now and extend that to N invocations using HTTP multipart stuff?

upload-api/functions/bridge.js
travis: yea I think this would be great! also aligns with @gobengo's idea to use content negotiation to accept various iterations of the UCAN spec. 

I have a soft preference to pick one format to accept in this iteration, and follow pretty quickly with support for other content types - are json and dag-cbor the only two formats you'd imagine accepting here or is there a bigger list @Gozala ?

upload-api/functions/bridge.js
travis: cool - I am convinced that this should be a CAR!

upload-api/BRIDGE.md
```diff
@@ -0,0 +1,66 @@
...
+  "inputs": {
+    "link": "bafybeicxsrpxilwb6bdtq6iztjziosrqts5qq2kgali3xuwgwjjjpx5j24",
+    "size": 42
+  }
+}
+```
+
+And receive a JSON-encoded UCAN receipt in response.
```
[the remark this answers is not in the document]
travis: @Gozala I'm currently just returning the `Result` accessed via `receipt.out` in the receipt I get back from `Connection#execute` - that serializes nicely to JSON. Do you think it would be better to return a JSON-serialized version of the whole receipt? If so, what in the receipt do you think end users would want to access? I don't doubt it - just don't know much about receipts yet!

upload-api/functions/bridge.js
travis: I think this is just a question of picking a type - I've moved this to `@ucanto/interface`'s `DID` type, does that sound right to you @Gozala ?

upload-api/functions/bridge.js
travis: got it, done!

upload-api/functions/bridge.js
alanshaw: I think @Gozala is saying if you use `Schema.DID.parse` you don't need the jsdoc type or the conditional above checking for ":".

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,331 @@
...
+      context.serviceURL,
+      tasks,
+      delegation
+    )
+
+    const bridgeReceipts = await buildBridgeReceipts(receipts)
+    return bridgeReceipts.ok ? {
+      statusCode: 200,
+      headers: {
+        "Content-Type": "application/json"
```
alanshaw, line 288: https://www.iana.org/assignments/media-types/application/vnd.ipld.dag-json

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/**
+ * @type {import('../bridge/types').BodyParser}
+ */
+async function parseBody(contentType, body) {
+  const bodyBytes = await streamToArrayBuffer(body)
+  let parsedBody
+  if (contentType === 'application/json' || contentType === 'application/vnd.ipld.dag-json') {
+    parsedBody = dagJSON.parse(new TextDecoder().decode(bodyBytes))
```
alanshaw, line 106: It has a `decode` method no?

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,331 @@
...
+      return {
+        statusCode: 400,
+        body: 'request has no body'
+      }
+    }
+
+    if (!contentType) {
+      return {
+        statusCode: 400,
+        body: 'request has no Content-Type header'
```
alanshaw, line 244: Personally I'd just assume `dag-json` if no header...

alanshaw (approved):
🚀 I'm approving because I think this is fine and can merge if the spec does not change. 

However, I have left comments on the spec because I think we could align request/response formats a little more with the UCAN invocation spec.

upload-api/functions/bridge.js
travis: ah so - when I did this, AWS stopped handling the response correctly, but I think I can tweak it to use their base64 encoding stuff and this should work - will give it a shot!

upload-api/functions/bridge.js
travis: oh look at that it sure does!

upload-api/functions/bridge.js
travis: works for me!

upload-api/functions/bridge.js
travis: ah, stringifying already fixed it! done!

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,324 @@
...
+ * @param {import('../bridge/types').BridgeRequestContext} context
+ */
+export async function handleBridgeRequest(request, context) {
+  try {
+    const { authSecretHeader, authorizationHeader, body, contentType } = parseAwsLambdaRequest(request)
+
+    if (!authorizationHeader) {
+      return {
+        statusCode: 401,
+        body: 'request has no Authorization header'
```
alanshaw, line 223: I like to say "missing" rather than "request has no" 😬

upload-api/functions/bridge.js
```diff
@@ -0,0 +1,324 @@
...
+      context.serviceURL,
+      tasks,
+      delegation
+    )
+
+    const bridgeReceipts = await buildBridgeReceipts(receipts)
+    return bridgeReceipts.ok ? {
+      statusCode: 200,
+      headers: {
+        "Content-Type": "application/vnd.ipld.dag-json"
```
alanshaw, line 281: Would be nice if these had a content type export [like CAR](https://github.com/ipld/js-car/blob/1c50e4421cdd51a7c98507722abac96e1d41fccd/src/index.js#L10)...how do you feel about sending a couple of PRs?

travis: oo I like that - will dig into that today!

PR #327, opened 2024-02-07
fix: find equivalent cids also if not car cid

vasco-santos:
We should keep also equivalent CIDs that are not CARs to attempt in the bucket to get presigned URL

We will try for each of them to check if it is in bucket and generate presigned URL per:
- https://github.com/web3-storage/w3infra/blob/fix/find-equivalent-cids-also-if-not-car-cid/roundabout/functions/redirect.js#L81C23-L81C36
- https://github.com/web3-storage/w3infra/blob/fix/find-equivalent-cids-also-if-not-car-cid/roundabout/index.js#L25

### DIFF
--- a/roundabout/piece.js
+++ b/roundabout/piece.js
@@ -19,7 +19,6 @@ export const PIECE_V2_MULTIHASH = 0x10_11
 
 /** 
  * @typedef {import('multiformats/cid').CID} CID
- * @typedef {import('@web3-storage/w3up-client/types').CARLink} CARLink
  * @typedef {import('@web3-storage/content-claims/client/api').Claim} Claim
  **/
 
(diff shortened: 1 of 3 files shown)


roundabout/piece.js
```diff
@@ -57,13 +56,13 @@ export function asCarCid(cid) {
...
+ * Find the set of CIDs that are claimed to be equivalent to the Piece CID.
  * 
  * @param {CID} piece
  * @param {(CID) => Promise<Claim[]>} [fetchClaims] - returns content claims for a cid
  */
-export async function findEquivalentCarCids (piece, fetchClaims = createClaimsClientForEnv()) {
-  /** @type {Set<CARLink>} */
+export async function findEquivalentCids (piece, fetchClaims = createClaimsClientForEnv()) {
+  /** @type {Set<import('multiformats').UnknownLink>} */
   const cids = new Set()
```
alanshaw, line 66: Hmm, I think this is not going to dedupe like you think it will.

roundabout/piece.js
```diff
@@ -72,11 +71,9 @@ export async function findEquivalentCarCids (piece, fetchClaims = createClaimsCl
       continue
     }
     // an equivalence claim may have the pieceCid as the content cid _or_ the equals cid
-    // so check both properties for the car cid.
-    const carCid = asCarCid(claim.equals) ?? asCarCid(claim.content)
-    if (carCid) {
-      cids.add(carCid)
-    }
+    // so if content does not equal piece, we can grab the content. Otherwise equals
+    const equivalentCid = !claim.content.equals(piece) ? claim.content : claim.equals
```
alanshaw, line 75: IDK, maybe a little easier to read?:

```suggestion
    const equivalentCid = claim.content.equals(piece) ? claim.equals : claim.content
```

alanshaw (approved):

roundabout/piece.js
vasco-santos: yeah, I also think this set will probably not guarantee . will create an issue, as this was like it and look after

vasco-santos:
this was already happening on https://protocol-labs-it.sentry.io/issues/4439567707/?project=4506587250819072&project=4504253926539264&project=4504515009970176&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=22 and is a failure from content claims service

PR #334, opened 2024-02-21
fix: temporary disable unsupported lambda runtimes

vasco-santos:


vasco-santos:
this won't fix, so turning it off

PR #335, opened 2024-02-21
fix: update sst to latest

vasco-santos:
Fixes Deployment issues by finally updating to the LATEST sst and update aws cdk/sdk deps to be compatible. See [migration guide](https://docs.sst.dev/upgrade-guide) for details. In reality, the core changes apply right to [upgrade-guide#upgrade-to-v20](https://docs.sst.dev/upgrade-guide#upgrade-to-v20), given remaining items is things we do not use today

You can run to know what versions should be used:
```sh
➜  w3infra git:(fix/update-sst-to-v2) npx sst version
SST: v2.40.3
CDK: v2.124.0
Constructs: v10.3.0
```

❗ while the github PR diff seems massive, actually most is on the `package-lock`, and changes and mostly methodical and the same things all over the place, like changing import paths

A few notes:
- srcPath was sadly removed on sst, which makes us need to use the entire path
- updated `multiformats` + `@ipld/car` dependencies because of the big changes in package-lock there were mismatch versions causing type issues in Links
- updated `testcontainers`  given billing already had a new one and local tests were flakey because sometimes too many containers open take too long to stop/start

### DIFF
--- a/test/helpers/deployment.js
+++ b/test/helpers/deployment.js
@@ -13,12 +13,7 @@ export function getStage () {
     return stage
   }
 
-  const f = fs.readFileSync(path.join(
-    process.cwd(),
-    '.sst/stage'
-  ))
-
-  return f.toString()
+  return fs.readFileSync(path.join(process.cwd(), '.sst', 'stage'), 'utf8')
 }
 
 export const getStackName = () => {


filecoin/test/compute-piece-cid.test.js
```diff
@@ -14,20 +14,29 @@ const AWS_REGION = 'us-west-2'
...
 
   Object.assign(t.context, {
     s3Client: client,
     dbEndpoint,
-    dynamoClient
+    dynamoClient,
+    stop: async () => {
+      await s3Stop()
+      await dynamoStop()
+    }
```
vasco-santos, line 32: forcing stop to decrease likelihood of flakey tests

package.json
```diff
@@ -12,17 +12,15 @@
     "console": "sst console",
     "lint": "tsc && eslint '**/*.js'",
     "clean": "rm -rf dist node_modules package-lock.json ./*/{.cache,dist,node_modules}",
-    "test": "npm test -w billing -w upload-api -w carpark -w replicator -w satnav -w ucan-invocation -w roundabout -w filecoin",
```
vasco-santos, line 15: we do not have that package anymore since https://github.com/web3-storage/w3infra/pull/315

sst.config.ts
```diff
@@ -0,0 +1,62 @@
+import type { SSTConfig } from 'sst'
```
vasco-santos, line 1: basically `stacks/index.js` is now this file :)

test/helpers/deployment.js
```diff
@@ -31,11 +42,15 @@ export const getAwsBucketClient = (region = getAwsRegion()) => new S3Client({
...
   if (process.env.SEED_APP_NAME) {
+    const stage = getStage()
     return `https://${stage}.up.web3.storage`
   }
 
   const require = createRequire(import.meta.url)
-  const testEnv = require('../../.test-env.json')
+  const testEnv = require(path.join(
+    process.cwd(),
+    '.sst/outputs.json'
```
vasco-santos, line 52: this was changed from new version, but of course too internal to be included in the migration guide... one of those "needle in a haystack" that took forever to get tests working with dynamic deployment of infra per PR stage

olizilla:
you make it looks so easy! I thought this was gonna be much more of faff.

Do you know if sst will figure out that the stateful resources like buckets and tables are the same thing across this upgrade. Need to verify that it's not gonna try and delete and recreate them.

vasco-santos:
> Do you know if sst will figure out that the stateful resources like buckets and tables are the same thing across this upgrade. Need to verify that it's not gonna try and delete and recreate them.

yeah, that is something I pointed out to @alanshaw . We need to see what the diff will be when doing the release with seed.run in staging to be totally sure. But I would say that it won't recreate, at least from the output I see on seed.run at the moment:

image

test/helpers/deployment.js
```diff
@@ -1,18 +1,29 @@
...
+  if (stage) {
+    return stage
+  }
+
+  const f = fs.readFileSync(path.join(
+    process.cwd(),
+    '.sst/stage'
+  ))
+
+  return f.toString()
```
alanshaw, line 21:
```suggestion
  return fs.readFileSync(path.join(process.cwd(), '.sst', 'stage'), 'utf8')
```

test/helpers/deployment.js
```diff
@@ -79,14 +102,18 @@ export const getSatnavBucketInfo = () => {
...
       Region: 'us-east-2'
     }
   }
 
   const require = createRequire(import.meta.url)
-  const testEnv = require('../../.test-env.json')
+  const testEnv = require(path.join(
+    process.cwd(),
+    '.sst/outputs.json'
+  ))
```
alanshaw, line 116: Pull out as a function?

PR #342, opened 2024-03-12
feat: billing dry run

alanshaw:
A script to do dry run of the billing pipeline. It reads production data, but does not write to any production tables.


billing/scripts/helpers.js
```diff
@@ -0,0 +1,66 @@
...
+      return item ? { ok: item } : { error: new EndOfQueue() }
+    }
+  }
+}
+
+/**
+ * @template T
+ * @returns {import('../lib/api.js').StorePutter<T> & import('../lib/api.js').StoreLister<any, T> & import('../lib/api.js').StoreGetter<any, T>}
+ */
+export const createMemoryStore = () => {
```
vasco-santos, line 28: I assume based on your ran that it does not load a lot of memory in the host running. wondering if we should write in disk to avoid getting out of memory, but probably nothing close to happen that we can start with in memory?

vasco-santos (approved):
lgtm ✨

PR #343, opened 2024-03-18
feat: upgrade storefront api with filecoin offer from bucket event instead of directly compute

vasco-santos:
This PR intends to [upgrade](https://github.com/web3-storage/w3up/pull/1332) storefront package. This enables `filecoin/offer` invocations lead to messages to the submit queue, where message consumer will compute Piece for the content. The main goal of this is to have the client to compute Filecoin Pieces and offer them, instead of pieces being directly computed from bucket event and written into the DB (skipping the submit validation).

Note that:
- given we will need to deal with client upgrades and old clients still running for the time being, I opted in this PR to still keep a version of the bucket event in place. However, instead of it running computation of piece, it also triggers `filecoin/offer` so that we can keep old clients running where `filecoin/offer` is not invoked. This of course means that we may need to run Piece computation twice for the old path. However, looking at our AWS bill, we paid [200 USD in February](https://us-east-1.console.aws.amazon.com/cost-management/home?region=us-west-2#/cost-explorer?chartStyle=STACK&costAggregate=unBlendedCost&endDate=2024-02-29&excludeForecasting=false&filter=%5B%7B%22dimension%22:%7B%22id%22:%22TagKey%22,%22displayValue%22:%22Tag%22%7D,%22operator%22:%22INCLUDES%22,%22values%22:%5B%7B%22value%22:%22w3infra%22,%22displayValue%22:%22w3infra%22%7D%5D,%22growableValue%22:%7B%22value%22:%22Project%22,%22displayValue%22:%22Project%22%7D%7D%5D&futureRelativeRange=CUSTOM&granularity=Monthly&groupBy=%5B%22Service%22%5D&historicalRelativeRange=LAST_6_MONTHS&isDefault=true&reportName=New%20cost%20and%20usage%20report&showOnlyUncategorized=false&showOnlyUntagged=false&startDate=2023-09-01&usageAggregate=undefined&useNormalizedUnits=false) for lambda execution of whole w3infra. This assumes then that we will not incur a lot of additional cost while making migration much simpler
- datastore currently implements a composed store with S3+R2, which allows us to move on an iterative approach where we can deploy first step without client upgrade, while we can iterate on writes to anywhere with the overall picture.

Needs:
- [x] https://github.com/web3-storage/w3up/pull/1332


filecoin/test/helpers/resources.js
```diff
@@ -140,7 +140,7 @@ export const createSQS = async (opts = {}) => {
   const port = opts.port || 9324
 
   const queue = await pRetry(() =>
-    new Container('softwaremill/elasticmq')
+    new Container('softwaremill/elasticmq:1.5.4')
```
vasco-santos, line 143: looks like latest queue docker image is bad 😱

filecoin/functions/handle-piece-insert-to-content-claim.js
```diff
@@ -56,24 +55,24 @@ async function pieceCidReport (event) {
...
+      invocationConfig: {
+        issuer: claimsIssuer,
+        audience: connection.id,
+        with: claimsIssuer.did(),
+        proofs: claimsProofs
+      },
+    },
+  }
 
+  const { ok, error } = await storefrontEvents.handlePieceInsertToEquivalencyClaim(context, record)
```
vasco-santos, line 70: we now use the API 🎉

filecoin/functions/handle-filecoin-submit-message.js
```diff
@@ -34,9 +36,26 @@ async function handleFilecoinSubmitMessage (sqsEvent) {
...
+    s3BucketName,
+    r2BucketName,
+    r2BucketEndpoint,
+    r2BucketAccessKeyId,
+    r2BucketSecretAccessKey
+  } = getEnv()
   const context = {
-    pieceStore: createPieceTable(AWS_REGION, pieceTableName)
+    pieceStore: createPieceTable(AWS_REGION, pieceTableName),
+    dataStore: composeDataStoresWithOrderedStream(
```
vasco-santos, line 49: we want to read from S3 first, fallback to R2 if not there (prep for new world)

filecoin/store/data.js
```diff
@@ -0,0 +1,161 @@
...
+      for await (const chunk of out) {
+        chunks.push(chunk)
+      }
+      const blob = new Blob(chunks)
+      const cid = await CAR.codec.link(new Uint8Array(await blob.arrayBuffer()))
+
+      const putCmd = new PutObjectCommand({
+        Bucket: bucketName,
+        Key: `${cid.toString()}/${cid.toString()}.car`,
+        Body: bytes
```
vasco-santos, line 57: we only need this for the tests, so that we can put something on the store. Maybe I should change the API to have an augmented store just on the tests interface, but not the actual service store. WDYT?

filecoin/test/helpers/resources.js
vasco-santos: This was failing with a random error, fixed it for now to previous version

package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "tsc && eslint '**/*.js'",
     "clean": "rm -rf dist node_modules package-lock.json ./*/{.cache,dist,node_modules}",
     "test": "npm test -w billing -w upload-api -w carpark -w replicator -w satnav -w roundabout -w filecoin",
-    "test-integration": "ava --verbose --serial --timeout=600s test/*.test.js",
+    "test-integration": "ava --verbose --serial --timeout=660s test/*.test.js",
```
vasco-santos, line 16: not sure if needed, but now tests look a bit slower with extra queue

filecoin/store/data.js
```diff
@@ -0,0 +1,161 @@
...
+    for (const stream of streamFunctions) {
+      result = await stream(e)
+      if (result.ok) {
+        return result
+      }
+    }
+    if (result === undefined) {
+      throw new Error('no result received')
+    }
+    return result
```
Gozala, line 159: Nit: result should probably aggregate errors as opposed to just returning the last error.

filecoin/functions/handle-filecoin-submit-message.js
Gozala: Nit: this would read more naturally if it was something like

```js
cons s3 = createDataStore(AWS_REGION, s3BucketName)
const r2 = createDataStore(R2_REGION, r2BucketName, {
    endpoint: r2BucketEndpoint,
    credentials: {
      accessKeyId: r2BucketAccessKeyId,
      secretAccessKey: r2BucketSecretAccessKey,
    },
})

const store = S3.or(r2)
```

Gozala (approved):

filecoin/functions/handle-filecoin-submit-message.js
vasco-santos: good call, we actually already use this pattern on `upload-api` https://github.com/web3-storage/w3infra/blob/main/upload-api/functions/ucan-invocation-router.js#L185C52-L186C1 and that is why I kept. But probably we could get more declarative on a follow up

filecoin/store/data.js
vasco-santos: true, given this is a temporary thing I am tempted to keep it, given otherwise would be difficult to align the types

PR #349, opened 2024-04-10
feat: add snapshot size to dry run report

alanshaw:
This adds space size (in bytes) to the dry run report.

Also adds the new `did:web:free.web3.storage` plan (that NFT.Storage is using) to the pricing calculator.


vasco-santos (approved):
lgtm 👌🏼

PR #350, opened 2024-04-12
fix: make filecoin accept receipt cron run more times

vasco-santos:
We saw some backlog growing while issuing `filecoin/accept` receipts. This is due to not have pagination in the query that goes through submmited pieces pending a receipt.

Looking at the CRON execution duration, we can conclude for the last 5 days that average duration goes under 300k ms (5min). Which means we were having CRON alive 5 min and then 5 minutes sleeping. While we do not implement pagination on the query this seems a quick win.

Also note that if the CRON function ends up sometimes timing out, it is not really problematic. Looking at the [code we run](https://github.com/web3-storage/w3up/blob/main/packages/filecoin-api/src/storefront/events.js#L187), we just go through table and, for each entry go through receipt chain to see if we can update the status (which makes a new lambda run to issue receipt). Therefore, if we timeout and restart the job, we will keep going where we were

image

### DIFF
--- a/stacks/filecoin-stack.js
+++ b/stacks/filecoin-stack.js
@@ -120,7 +120,7 @@ export function FilecoinStack({ stack, app }) {
    */
   const dealTrackCronName = getCdkNames('deal-track-cron', stack.stage)
   new Cron(stack, dealTrackCronName, {
-    schedule: 'rate(10 minutes)',
+    schedule: 'rate(8 minutes)',
     job: {
       function: {
         handler: 'filecoin/functions/handle-cron-tick.main',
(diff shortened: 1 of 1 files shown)


stacks/filecoin-stack.js
```diff
@@ -120,7 +120,7 @@ export function FilecoinStack({ stack, app }) {
    */
   const dealTrackCronName = getCdkNames('deal-track-cron', stack.stage)
   new Cron(stack, dealTrackCronName, {
-    schedule: 'rate(10 minutes)',
+    schedule: 'rate(8 minutes)',
```
vasco-santos, line 123:
```suggestion
    schedule: 'rate(6 minutes)',
```

stacks/filecoin-stack.js
```diff
@@ -132,7 +132,7 @@ export function FilecoinStack({ stack, app }) {
           AGGREGATOR_DID,
           PROOF: STOREFRONT_PROOF,
         },
-        timeout: '9 minutes',
+        timeout: '8 minutes',
```
vasco-santos, line 135:
```suggestion
        timeout: '6 minutes',
```

vasco-santos:
merging this to help out to go through the backlog existing at the moment

PR #352, opened 2024-04-17
fix: get service url also working for dev deployments

vasco-santos:


stacks/config.js
```diff
@@ -150,14 +150,15 @@ export function setupSentry (app, stack) {
...
+export function getServiceURL (stack, customDomain) {
   // in production we use the top level subdomain
   if (stack.stage === 'prod') {
     return 'https://up.web3.storage'
-  // in staging and PR environments we use a sub-subdomain
-  } else if (stack.stage === 'staging' || stack.stage.startsWith('pr')) {
-    return `https://${stack.stage}.up.web3.storage`
+  // Derive from custom domain if there is one, which is used in staging, PR envs and dev envs
+  } else if (customDomain) {
+    return customDomain.domainName
```
vasco-santos, line 161: this way we also include dev setups, while supporting staging and PRs. Only when a custom domain is not set, we need to go via env var

travis (approved):

PR #353, opened 2024-04-17
feat: add blob protocol to infra

vasco-santos:
This PR creates stores and wires up new `upload-api` running `blob/add`, `web3.storage/blob/allocate`, `web3.storage/blob/accept` and `ucan/conclude` capabilities. Tests are also imported from `upload-api` implementation and run here.

As agreed on https://github.com/web3-storage/w3up/issues/1343 , there won't be any deduping between old world and new world. Therefore, we have new `allocations` table, and use different key schema in `carpark`. We are writing blobs keyed as `base58btc` as previously discussed as `${base58btcEncodedMultihash}/${base58btcEncodedMultihash}.blob`. I added `.blob` suffix but I am happy to other suggestions. Depending on how we progress with the reads side, we can consider creating a new bucket to fully isolate new content?

The `receipts` and `tasks` storage end up being more complicated as they need to follow https://github.com/web3-storage/w3infra/blob/main/docs/ucan-invocation-stream.md#buckets, and is essentially the same as what happens on https://github.com/web3-storage/w3infra/blob/main/upload-api/ucan-invocation.js#L66 but at a different level as this is a proactive write of tasks and receipts.

Integration tests are in progress in branch https://github.com/web3-storage/w3infra/tree/feat/add-blob-protocol-to-service-integration-test. They currently run in my local dev setup, and have full functionality from `blob/add`, followed by HTTP PUT to bucket and `ucan/conclude` containing the receipt of the put. This makes scheduled `blob/accept` run and a receipt with a location claim being available. You can see the tests in branch, on file `test/blob.test.js`. I want to only add them here once we can rely on the client to do the manual hack I did here


upload-api/stores/allocations.js
```diff
@@ -0,0 +1,211 @@
...
+
+  return useAllocationsStorage(dynamoDb, tableName)
+}
+
+/**
+ * @param {DynamoDBClient} dynamoDb
+ * @param {string} tableName
+ * @returns {AllocationsStorage}
+ */
+export function useAllocationsStorage(dynamoDb, tableName) {
```
vasco-santos, line 43: This is quite similar with `storeTable` with few modifications given new columns

upload-api/stores/blobs.js
```diff
@@ -0,0 +1,134 @@
...
+  return useBlobsStorage(s3, bucketName)
+}
+
+/**
+ *
+ * @param {S3Client} s3
+ * @param {string} bucketName
+ * @returns {BlobsStorage}
+ */
+export function useBlobsStorage(s3, bucketName) {
```
vasco-santos, line 39: This is quite similar with `buckets/car-store` with few modifications given new key schema

upload-api/scheduler.js
```diff
@@ -0,0 +1,43 @@
...
+    this.getServiceConnection = getServiceConnection
+  }
+
+  /**
+   * @param {ServiceInvocation} invocation
+   * @returns {Promise<Result>}
+   */
+  async schedule(invocation) {
+    const connection = this.getServiceConnection()
+    const [res] = await connection.execute(invocation)
```
Gozala, line 34: Should this write a receipt into receipt storage ?

upload-api/stores/allocations.js
```diff
@@ -0,0 +1,211 @@
...
+     */
+    insert: async ({ space, blob, invocation }) => {
+      const insertedAt = new Date().toISOString()
+      const multihash58btc = base58btc.encode(blob.digest)
+
+      const item = {
+        space,
+        multihash: multihash58btc,
+        size: blob.size,
+        invocation: invocation.toString(),
```
Gozala, line 109: I realize this just implements what has be previously added to upload-api, however sadly I'm only recognizing this here. It is not a blocker, but perhaps a good idea to do in a followup if we agree.

I am realizing that this should probably link to the signed receipt as opposed to invocation. In a way all our stores should simply be indexes over receipts because receipts are signed commitments from us, also they do link back to the corresponding invocation.

Rational here is that anyone with access to table could create valid record like this, however if it was linking to signed receipt they would not be able to without signing key.

upload-api/stores/blobs.js
Gozala: Might be worth adding this as code comment actually.

upload-api/stores/blobs.js
```diff
@@ -0,0 +1,134 @@
...
+  return {
+    /**
+     * @param {Uint8Array} multihash
+     */
+    has: async (multihash) => {
+      const encodedMultihash = base58btc.encode(multihash)
+      const cmd = new HeadObjectCommand({
+        Key: `${encodedMultihash}/${encodedMultihash}.blob`,
```
Gozala, line 47: Do you recall reason why had same named dir and file within it ? I am pretty sure there was one, but I can't seem to recall what it was. If you do could you please add code comment mentioning it so future us can be grateful

upload-api/stores/blobs.js
```diff
@@ -0,0 +1,134 @@
...
+      const encodedMultihash = base58btc.encode(multihash)
+      const multihashDigest = digestDecode(multihash)
+      const checksum = base64pad.baseEncode(multihashDigest.digest)
+      const cmd = new PutObjectCommand({
+        Key: `${encodedMultihash}/${encodedMultihash}.blob`,
+        Bucket: bucketName,
+        ChecksumSHA256: checksum,
+        ContentLength: size,
+      })
+      const expiresIn = 60 * 60 * 24 // 1 day
```
Gozala, line 78: Nit: Should we expose this as an optional argument ? Seems like something we may want to be able to change in some cases.

Gozala (approved):

upload-api/scheduler.js
vasco-santos: it is not needed. Sadly this looks then a bit confusing and looking forward to have some kind of receipt storage part of ucanto/scheduler later on. Here this does a HTTP request executing the invocation, which goes through the `ucan-invocation-router` and is hooked up with the receipt storage: https://github.com/web3-storage/w3infra/blob/main/upload-api/functions/ucan-invocation-router.js#L280

I will add a comment to explain this

upload-api/stores/allocations.js
vasco-santos: I understand your point and this looks like it make sense. I will create an issue to address this as follow up because I am not sure on how we can do this in practise given this is `blob/add` invocation CID. And its receipt link would not be possible to compute when we run `blob/allocate` given it is also dependent on this receipt itself.

upload-api/stores/blobs.js
vasco-santos: yes, it is related to how Bucket rate limiting happens (at least in S3), where rate limits are per path, rather than entire bucket. With this, we protect ourselves from rate limiting by having folder names as hashes with unique content inside

upload-api/stores/blobs.js
vasco-santos: yes, good call! I will expose this in interface and here

upload-api/stores/allocations.js
vasco-santos: https://github.com/web3-storage/specs/issues/122

Gozala: Ah sorry I was not very clear here, I meant that it should be receipt to `blob/allocate` not to receipt of the `blob/add`

PR #354, opened 2024-04-18
feat: add blob protocol to infra integration tests

vasco-santos:
Adds integration tests for blob implementation started on https://github.com/web3-storage/w3infra/pull/353 .

This test aims to test all the main flows that MUST happen once a Blob is written:
- [x] Blob is written and stored on target
- [x] Blob is written via `w3up-client`
- [ ] Blob index is stored on user space
- [x] Blob is retrievable from Roundabout
- [ ] Blob blocks are retrievable from Hoverbord
- [ ] Blob is retrievable by its root from w3link
- [x] Blob makes its way into Filecoin pipeline after a `filecoin/offer`

In addition, `store-add` tests are added independently to make sure old clients keep operation for writes + reads!


test/blob.test.js
```diff
@@ -0,0 +1,85 @@
+import { testBlob as test } from './helpers/context.js'
+
+import { base58btc } from 'multiformats/bases/base58'
+import { HeadObjectCommand } from '@aws-sdk/client-s3'
+import { Assert } from '@web3-storage/content-claims/capability'
```
vasco-santos, line 5: needs devDep on top level `package.json`

PR #356, opened 2024-04-24
feat(billing): support blob capabilities

alanshaw:
* Use the blob allocate receipt to track space size additions.
* Use blob remove receipt to track space size reductions.

### DIFF
--- a/billing/test/lib/ucan-stream.js
+++ b/billing/test/lib/ucan-stream.js
@@ -1,4 +1,7 @@
 import { Schema } from '@ucanto/core'
+import * as ServiceBlobCaps from '@web3-storage/capabilities/web3.storage/blob'
+import * as BlobCaps from '@web3-storage/capabilities/blob'
+import * as StoreCaps from '@web3-storage/capabilities/store'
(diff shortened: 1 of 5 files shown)


billing/test/lib/ucan-stream.js
```diff
@@ -24,14 +24,60 @@ export const test = {
...
+     */
     const receipts = [{
       type: 'receipt',
       carCid: randomLink(),
       invocationCid: randomLink(),
       value: {
         att: [{
-          with: await randomDID(),
+          with: await randomDIDKey(),
+          can: 'web3.storage/blob/allocate',
```
vasco-santos, line 42: Nit, could we use exported `.can`  from capabilities repo instead?

vasco-santos (approved):
lgtm

billing/test/lib/ucan-stream.js
alanshaw: I did reach for that first but then I realised I didn't do it for the rest of them in the other tests and it would be a bigger change. Worth doing in retrospect.

PR #360, opened 2024-04-29
fix: upgrade filecoin api with content store to rely on roundabout

vasco-santos:
Today Storefront relies directly on a bucket (in S3) to read bytes to derive PieceCID for validation. With blob protocol, we are moving to write to R2 only 🙌🏼  therefore, we aim to make filecoin Storefront able to read from anywhere, but also not increase the infra costs with this change.

This PR changes Storefront to rely on Roundabout to read Blob bytes to derive Piece CID from anywhere instead of `contentStore` being directly hooked to a bucket. See [usage](https://github.com/w3s-project/w3up/blob/main/packages/filecoin-api/src/storefront/events.js#L43)

Relying on Roundabout introduces a layer of indirection, but in current implementation where this validation runs in AWS and content will likely be read from CF, makes it cost optimal to read from Roundabout.

Closes https://github.com/w3s-project/w3up/issues/1349


sst.config.ts
```diff
@@ -41,6 +41,7 @@ export default {
 
     app.stack(BusStack)
     app.stack(UploadDbStack)
+    app.stack(RoundaboutStack)
```
vasco-santos, line 44: FilecoinStack now requires Roundabout endpoint deployed

filecoin/store/content.js
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Stream Blob bytes for a given CID.
+     *
+     * @param {import('@ucanto/interface').UnknownLink} cid 
+     */
+    stream: async (cid) => {
+      // create URL for the link to be fetched
+      const getUrl = new URL(`/${cid.toString()}`, storeHttpEndpoint)
+
+      // Retry a few times as it looks like R2 sometimes takes a bit to make it available
```
vasco-santos, line 24: context: This was making integration tests sometimes fail before I added the retry

filecoin/store/content.js
```diff
@@ -0,0 +1,60 @@
...
+      let res
+      try {
+        res = await pRetry((async () => {
+          const fetchRes = await fetch(getUrl, {
+            // Follow potential redirects
+            redirect: 'follow',
```
alanshaw, line 30: This is the default no?

filecoin/store/content.js
```diff
@@ -0,0 +1,60 @@
...
+          })
+          if (fetchRes.status === 404) {
+            throw new RecordNotFound(`blob ${cid.toString()} not found in store`)
+          } else if (fetchRes.status > 299 || !fetchRes.body) {
```
alanshaw, line 34:
```suggestion
          } else if (!fetchRes.ok || !fetchRes.body) {
```

filecoin/store/content.js
```diff
@@ -0,0 +1,60 @@
...
+        /** @type {RecordNotFound | StoreOperationFailed} */
+        // @ts-ignore
+        const error = err
+        return {
+          error
+        }
+      }
+
+      // To satisfy typescript
+      if (!res.body) {
```
alanshaw, line 49: Return `fetchRes.body` above where you have already checked it is not null and then you don't need this check.

alanshaw (approved):

PR #363, opened 2024-05-01
feat: add dependencies for `index/add` invocations

alanshaw:
Upgrades to `@web3-storage/upload-api` v13 and adds dependencies for `index/add` invocations.

Adds multihashes to the Elastic IPFS SQS multihashes queue and adds block index data to Dynamo `blocks-cars-position` table.

Note: I did not also generate dudewhere/satnav indexes. Hoverboard reads directly from dynamo and Freeway uses location claims (generated from dynamo). I don't think we _need_ them anymore.


stacks/upload-api-stack.js
```diff
@@ -66,6 +66,7 @@ export function UploadApiStack({ stack, app }) {
     customDomain,
     defaults: {
       function: {
+        timeout: '30 seconds',
```
alanshaw, line 69: Seems to max out at ~14s, setting to 30s to be safe.

Screenshot 2024-05-02 at 15 13 10

upload-api/test/helpers/ipni-service.js
```diff
@@ -0,0 +1,45 @@
...
+    client: dynamo,
+    name: tableName
+  })
+
+  const messages = new Set()
+  return Object.assign(
+    useIPNIService(multihashesQueue, carsBlocksPositionStore),
+    {
+      /** @param {import('multiformats').MultihashDigest} digest */
+      async query (digest) {
```
Gozala, line 30: Nit: I think calling this `has` or `contains` would be far less confusing

upload-api/test/helpers/queue.js
```diff
@@ -0,0 +1,55 @@
...
+      }, {
+        retries: 3,
+        minTimeout: 100,
+        onFailedAttempt: console.warn
+      })
+    }
+  }
+
+  return messages
+}
```
Gozala, line 55:
```suggestion
}

```

Gozala (approved):
Provided some feedback inline mostly around naming suggestions and alike. I think it looks great regardless though! Thanks for making it happen.

upload-api/external-services/ipni-service.js
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * @param {import('multiformats').MultihashDigest[]} digests
+   * @returns {Promise<import('@ucanto/interface').Result<import('@ucanto/interface').Unit, import('@ucanto/interface').Failure>>}
+   */
+  async addAll (digests) {
+    try {
+      const items = [...digests]
+      while (true) {
+        const batch = items.splice(0, 10)
```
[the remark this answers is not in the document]
alanshaw: It's the max number of messages you can send to an SQS queue in a batch.

upload-api/test/helpers/ipni-service.js
alanshaw: Yes I did consider this. However I don't believe there's an explicit IPNI API for that, and I thought at some point we might want query to actually return the result of querying IPNI for a multihash.

Should have changed in w3up types before now...

PR #364, opened 2024-05-14
fix: remove unnecessary dependency 

travis:
I'm not sure why I added this, but suspect it was a debugging attempt gone awry - let's remove it!


vasco-santos (commented):
@travis I don't understand why this is in main `package.json`? I believe is not needed IIRC, and `upload-api/package.json` is already upgraded

travis:
hm that's a good point... I'll try removing it!

travis:
ok this seems to work just fine! @vasco-santos if you're happy with this I can merge

PR #365, opened 2024-05-15
chore: update upload-api dep

alanshaw:


package.json
```diff
@@ -33,7 +33,7 @@
     "@web-std/fetch": "^4.1.0",
     "@web3-storage/data-segment": "^5.1.0",
     "@web3-storage/filecoin-client": "^3.3.3",
-    "@web3-storage/w3up-client": "^12.5.3",
+    "@web3-storage/w3up-client": "^14.0.0-rc.0",
```
vasco-santos, line 36: this will break integration tests, maybe you can skip them as I am working on this

alanshaw:
Test failures are because `blob-index` is using an old version of `carstream` and it does so because it doesn't declare the version it needs in it's dependencies 🤦 

Fix here: https://github.com/w3s-project/w3up/pull/1467

vasco-santos (approved):

PR #366, opened 2024-05-16
fix: add blob metrics

vasco-santos:
Adds blob metrics for admin and space to track:
- count `blob/add`
- sum `blob/add` (`nb.blob.size`)
- count `blob/remove`
- sum `blob/remove` allocated size


alanshaw (approved):
LGTM!

PR #369, opened 2024-05-16
feat: add blob protocol to infra integration tests

vasco-santos:
Adds integration tests for blob implementation started on https://github.com/web3-storage/w3infra/pull/353 .

This test aims to test all the main flows that MUST happen once a Blob is written:
- [x] Blob is written and stored on target
- [x] Blob is written via `w3up-client`
- [x] Blob index is stored on user space
- [x] Blob is retrievable from Roundabout
- [x] Blob blocks are retrievable from Hoverbord
- [x] Blob is retrievable by its root from w3link
- [x] Blob makes its way into Filecoin pipeline after a `filecoin/offer`

In addition, `store-add` tests are added independently to make sure old clients keep operation for writes + reads!

Follow ups:
- [ ] @vasco-santos to create an issue to create ENV VARs for the testing names added like `w3s.link` hostname


test/blob.test.js
```diff
@@ -0,0 +1,251 @@
...
+    throw new Error('failed to archive DAG index', { cause: indexBytes.error })
+  }
+  // Store the index in the space
+  const resIndex = await Blob.add(serviceProps.conf, indexBytes.ok, { connection: serviceProps.connection })
+  const indexLink = Link.create(carCodec.code, resIndex.multihash)
+
+  // Register the index with the service
+  await Index.add(serviceProps.conf, indexLink, { connection: serviceProps.connection })
+  // Register an upload with the service
+  await Upload.add(serviceProps.conf, root, shards, { connection: serviceProps.connection })
```
alanshaw, line 141: Why are we not using the `@web3-storage/w3up-client` for this?

test/store.test.js
```diff
@@ -0,0 +1,239 @@
...
+          shards.push(meta.cid)
+        },
+      })
+    )
+
+  if (root === undefined) throw new Error('missing root CID')
+  t.is(shards.length, 1)
+
+  // Invoke upload/add
+  await Upload.add(serviceProps.conf, root, shards, { connection: serviceProps.connection })
```
alanshaw, line 94: Why not use an older version of the client for this?

test/blob.test.js
vasco-santos: See comment above

```
// Add blob using custom client to be able to access receipts
```

I want to have the CIDs for the Receipts so that I can make sure they exist. The upload client is fully used on `integration.test.js`, this is a more internal testing. Also, this way, I have access to the CID of the index that I can also check is written

vasco-santos: I will improve docs though

test/store.test.js
vasco-santos: I attempted to do so, but there were some dependencies collision and I opted to get this similarly to `blob` tersts

test/blob.test.js
vasco-santos: probably, I could also do a follow up where in the client we export a function that does this that client also uses internally so that we do not need to maintain this code here

PR #372, opened 2024-05-17
fix: claim correct location

alanshaw:
When IPNI service writes to dynamo, use the `BlobsStorage` to get the actual download URL for a given multihash.


vasco-santos (approved):
LGTM

PR #380, opened 2024-05-25
feat: agent store integration

Gozala:
Integrates agent-store into the infra and removes custom invocation execution and storage that was used previously. Requires https://github.com/w3s-project/w3up/pull/1479


read-pipeline.tldr
```diff
@@ -0,0 +1 @@
..."stylesForNextShape":{},"brush":null,"scribbles":[],"cursor":{"type":"default","rotation":0},"isFocusMode":false,"exportBackground":true,"isDebugMode":false,"isToolLocked":false,"screenBounds":{"x":0,"y":0,"w":1213,"h":852},"insets":[false,false,false,false],"zoomBrush":null,"isGridMode":false,"isPenMode":false,"chatMessage":"","isChatting":false,"highlightedUserIds":[],"canMoveCamera":true,"isFocused":true,"devicePixelRatio":2,"isCoarsePointer":false,"isHoveringCanvas":false,"openMenus":["main menu","main-menu-sub.file"],"isChangingStyle":false,"isReadonly":false,"meta":{},"duplicateProps":null,"id":"instance:instance","currentPageId":"page:page","typeName":"instance"},{"x":660.3196135010137,"y":783.1571105809057,"z":0.47983117693813965,"meta":{},"id":"camera:page:page","typeName":"camera"},{"editingShapeId":null,"croppingShapeId":null,"selectedShapeIds":[],"hoveredShapeId":null,"erasingShapeIds":[],"hintingShapeIds":[],"focusedGroupId":null,"meta":{},"id":"instance_page_state:page:page","pageId":"page:page","typeName":"instance_page_state"},{"id":"pointer:pointer","typeName":"pointer","x":-590.1858969619184,"y":-696.4566948941541,"lastActivityTimestamp":1714667796351,"meta":{}}]}
```
vasco-santos, line 1: can you also add svg so that is visible here? Also I think we should add this to the docs folder

vasco-santos: we have all the other diagrams in the infra, I have no idea at the moment what this diagram is, but I wonder if we should just have a repo for all the diagrams instead

upload-api/functions/ucan-invocation-router.js
```diff
@@ -152,72 +155,127 @@ export async function ucanInvocationRouter(request) {
...
+  const serviceSigner = getServiceSigner({
+    did: UPLOAD_API_DID,
+    privateKey: PRIVATE_KEY,
+  })
 
-  const tasksStorage = createTasksStorage(AWS_REGION, invocationBucketName, workflowBucketName)
-  const receiptsStorage = createReceiptsStorage(AWS_REGION, taskBucketName, invocationBucketName, workflowBucketName)
-  const allocationsStorage = createAllocationsStorage(AWS_REGION, allocationTableName, {
-    endpoint: dbEndpoint,
+  const agentStore = AgentStore.open({
```
vasco-santos, line 163: For past experiences on testing, we end up always allowing to pass the bucket clients so that either Lambdas, or tests give them to the implementation of a Store. I would recommend this pattern to keep consistency in the project, as well as make testing easy then

upload-api/functions/ucan-invocation-router.js
```diff
@@ -152,72 +155,127 @@ export async function ucanInvocationRouter(request) {
...
-  const subscriptionTable = createSubscriptionTable(AWS_REGION, subscriptionTableName, {
-    endpoint: dbEndpoint
-  });
+  const delegationBucket = createDelegationsStore(
+    r2DelegationBucketEndpoint,
+    r2DelegationBucketAccessKeyId,
+    r2DelegationBucketSecretAccessKey,
+    r2DelegationBucketName
+  )
+  const subscriptionTable = createSubscriptionTable(
```
vasco-santos, line 211: can we do any format changes in a different PR? This is already going to be quite big, and will make super difficult to review if all format is changing

vasco-santos (commented):
can we please move the formatting chanegs to other PR? This makes reviewing again super more difficult

read-pipeline.tldr
Gozala: I did not intend to publish it here

upload-api/functions/ucan-invocation-router.js
Gozala: You can actually pass `store.connection.channel` instead of `store.connection.address` to de exactly that.

upload-api/functions/ucan-invocation-router.js
Gozala: Yeah I'll undo the formatting changes, they were not my intention it's just vscode formatted everything with prettier because I forgot to disable it on this repo after deleting all my configs in attempt to solve the issue that was caused by docker images 😭

upload-api/stores/agent/stream.js
```diff
@@ -0,0 +1,154 @@
...
+  const records = []
+  for (const member of message.index) {
+    if (member.invocation) {
+      const { task, invocation, message } = member.invocation
+      records.push({
+        Data: UTF8.fromString(
+          JSON.stringify({
+            // This is bad naming but not worth a breaking change
+            carCid: message.toString(),
+            task: task.toString(),
```
Gozala, line 102: @vasco-santos I have realized only when updating tests a few minutes ago that this was called `invocationCid` and maybe something depends on this. My intention was to prepare things for a change where task and invocation cid aren't the same, which is why I put `task` link here and removed `invoactionCid` which appeared obsolete given that `value.cid` was the same.

I'm not sure what this was used for and whether it meant to refer to task or invocation, please let me know and we can change this as needed.

vasco-santos: I think invocationCID is used on several places like Billing that I am not familiar with. This also implies changes on the kinesis consumers then, and I would like to not tie changes in this PR so deep into other things. Can we keep this with the same and change those as follow up?

upload-api/functions/validate-email.jsx
```diff
@@ -87,6 +87,7 @@ function createAuthorizeContext () {
     PROVIDERS = '',
     STRIPE_PRICING_TABLE_ID = '',
     STRIPE_PUBLISHABLE_KEY = '',
+    UCAN_LOG_STREAM_NAME = '',
```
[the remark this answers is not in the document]
vasco-santos: ok it is there: https://github.com/w3s-project/w3infra/blob/main/stacks/upload-api-stack.js#L118

upload-api/stores/agent/stream.js
```diff
@@ -0,0 +1,154 @@
...
+      })
+    }
+
+    if (member.receipt) {
+      const { task, receipt, message } = member.receipt
+      // Prior implementation used to resolve the invocation and include it's
+      // details the stream record. If invocation was not found it threw
+      // exception which would result in HTTP status 500.
+
+      // 🔬 Need to figure if there are consumers downstream that depend on
```
vasco-santos, line 123: IIRC it is currently needed. There are consumers that need the information of the invocation to track metrics. For instance, when we receive a `store/add` receipt with 'ok', we need the invocation info to know the `size` value to increment the size os bytes received by the service

I would wish we just attempt these changes in follow up, or we are going to make this quite complicated

upload-api/stores/agent.js
```diff
@@ -0,0 +1,118 @@
...
+    return this
+  }
+
+  /**
+   * @param {API.ParsedAgentMessage} message
+   * @returns {Promise<API.Result<API.Unit, API.WriteError>>}
+   */
+  async write({ data, index, source }) {
+    const message = { data, source, index: [...index] }
+    const save = Store.write(this.connection.store, message)
```
vasco-santos, line 48: There is some non compatibility with previous solution that is making `/receipt` endpoint not working anymore. [Here is the lambda for receipt endpoint](https://github.com/w3s-project/w3infra/blob/main/upload-api/functions/receipt.js). In short, it reads from invocation bucket (baf....in) and redirects to the workflow bucket. So, with the new key paths it actually works, but what is being stored in the workflow looks somewhat different now compared to before.

This is how we previously were storing the Body of the Message request: 
- https://github.com/w3s-project/w3infra/blob/main/upload-api/ucan-invocation.js#L71-L76
- https://github.com/w3s-project/w3infra/blob/main/upload-api/ucan-invocation.js#L73
- https://github.com/w3s-project/w3infra/blob/main/upload-api/ucan-invocation.js#L229

while in this PR we are storing the message format above (which actually looks exactly the same).

However, when we try to read workflow from the [filecoin integration tests](https://github.com/w3s-project/w3infra/blob/main/test/filecoin.test.js#L131-L150), now it fails while reading with "'CBOR decode error: encountered invalid minor (31) for major 1'"

vasco-santos (commented):
I want to raise that except for the 2 problems still above and once we also merge here https://github.com/w3s-project/w3infra/pull/381, everything else looks good to me and I am confident we can merge this tomorrow

alanshaw (approved):
Looks great, I left a couple of comments but I'm super happy for this!

upload-api/functions/ucan.js
```diff
@@ -27,25 +22,34 @@ const AWS_REGION = process.env.AWS_REGION || 'us-west-2'
 async function handlerFn(request) {
   const {
     INVOCATION_BUCKET_NAME: invocationBucketName = '',
-    TASK_BUCKET_NAME: taskBucketName = '',
```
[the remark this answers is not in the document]
Gozala: It was never used and me and @vasco-santos realized that it would introduce extra indirection which we could avoid by using new pathing in form of


```
task/invocation@message.in
task/revocation@message.out
```

upload-api/ucan-invocation.js
```diff
@@ -47,206 +37,28 @@ export async function processUcanLogRequest(request, ctx) {
...
-      carCid,
-      invocationCid,
-      value: normalizeInvocation(invocation),
-      ts: Date.now(),
-      type: STREAM_TYPE.RECEIPT,
-      out: receipt.out,
+    const result = await ctx.agentStore.messages.write({
+      source: request,
+      data: message,
+      index: AgentMessage.index(message)
```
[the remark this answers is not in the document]
Gozala: Unfortunately this is yet another endpoint we have that simply receives invocations / receipts and simply writes them into a store and indexes them, so all of the stuff added in upload-api was does not cover this 🥺

I thine we could clean this up or perhaps deprecate this endpoint all together, but until we do this seemed like a best compromise I could think of

upload-api/stores/agent.js
vasco-santos: this was fixed

vasco-santos (approved):
LGTM

PR #383, opened 2024-05-30
feat: upgrade deps for new capabilities

vasco-santos:
Updates capabilities names and final dependencies for `upload-api` and `filecoin-api` with also the agentStore integrated


billing/test/lib/ucan-stream.js
```diff
@@ -83,6 +83,7 @@ export const test = {
           with: await randomDIDKey(),
           can: StoreCaps.add.can,
           nb: {
+            // @ts-expect-error different CID type per dep versions
```
vasco-santos, line 86: I will try to remove this in a different PR as follow up after deployment

alanshaw (approved):

PR #384, opened 2024-06-04
fix: upgrade w3up client rc with receipt poll for accept

vasco-santos:
Tests blob client with https://github.com/w3s-project/w3up/pull/1459


alanshaw (approved):
LGTM

PR #388, opened 2024-06-06
fix: stop writing to SATNAV

alanshaw:
No need to write to this bucket anymore - both freeway and hoverboard use content claims.


vasco-santos (approved):
LGTM! hurray 🎉 to much infra code gone

PR #390, opened 2024-06-11
feat: update upload-api dependency

alanshaw:
Updates `upload-api` dependency, bringing in a breaking change to blob storage (digest parameter type changed from `Uint8Array` to `MultihashDigest`), a change to `index/add` handler that publishes an index claim to content claims (needed implementation of content claims service for tests) and finally a change that drops writing to DUDEWHERE R2 bucket since it is no longer used in the reads pipeline.

see https://github.com/storacha-network/w3up/pull/1501


test/store.test.js
```diff
@@ -172,13 +163,22 @@ test('store protocol integration flow', async t => {
 
   // Verify w3link can resolve uploaded file
   console.log('Checking w3link can fetch root', root.toString())
-  const w3linkResponse = await fetch(
-    `https://${root}.ipfs-staging.w3s.link`,
-    {
-      method: 'HEAD'
+  const gatewayURL = `https://${root}.ipfs-staging.w3s.link`
+  const gatewayRetries = 5
+  for (let i = 0; i < gatewayRetries; i++) {
```
alanshaw, line 168: Since the CAR has to be indexed from a bucket event it may take a while to appear at the gateway. Adding retries with a short timeout seems to work.

vasco-santos (approved):
LGTM

PR #392, opened 2024-06-14
feat: add script to copy stores and uploads to a new space

travis:
There are many ways a user might lose control of a space. Because we can't fake the crypto our system is built upon, the simplest solution to this is to have them create a new space that they do control and then ask us to copy the contents of the old space into the new.

Of course this doesn't require copying bits around - we just need to update some indices in DynamoDB. This script finds all entries in the store and upload tables for a space and creates new entries for each of those items in the new space. This should result in the new space being functionally identical to the old

This is a somewhat sensitive operation - we should only do this when we trust the customer a great deal - the contents of a particular space may be sensitive and this is a potential attack vector to discover them.

This has not been tested extensively, though I have run it in my development AWS env and it seems to work - we should do some validation that using this results in a new space that is functionally identical to the old.

@alanshaw particularly interested in your thoughts on whether this will mess up billing at all?

### DIFF
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -45,6 +45,7 @@ cli
 
 cli
   .command('rebuild-space-metrics <space-did>', 'Rebuild store and upload metrics for a space.')
+  .option('-s, --snapshot', 'If set, create a billing snapshot after updating the space metrics.')
   .action(rebuildSpaceMetrics)
 
 cli.parse(process.argv)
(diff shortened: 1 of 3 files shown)


travis:
ok this is in good shape - my one remaining concern is how this will interact with billing - @hannahhoward maybe you have some insight since you've been in that code in the past few days?

juliangruber:
> ok this is in good shape - my one remaining concern is how this will interact with billing - @hannahhoward maybe you have some insight since you've been in that code in the past few days?

I was wondering about this as well, and was thinking that capabilities would need to be removed from the old account? But then I was wondering who pays for a file, if there are multiple accounts with access?

travis:
closing to reclaim IAM roles

PR #395, opened 2024-06-21
fix: publish to IPNI async

alanshaw:
This PR fixes an issue where a call to `space/index/add` does not have enough time to complete with the timeout forced by AWS API Gateway (29s).

When a CAR file has many thousands of blocks, it can take >29s to write all the blocks to the old E-IPFS dynamo table and to the E-IPFS multihashes queue (which feeds IPNI advertisements).

Previously looking at the exectuion times for lambdas I considered this possible, but clearly I did not encounter an upload with thousands fo blocks in the time window I observed.

The PR introduces an IPNI stack. It currently consists of 2 queues and 2 lambda consumers. 1 queue is for IPNI and the other is for writing to dynamo.

### Before

```mermaid
graph TD;
  subgraph w3infra
  upload-api[upload-api index/add handler]
  end
  subgraph E-IPFS
  upload-api-->|batch of up to 10x multihashes| multihashes-queue[SQS multihashes-queue];
  upload-api-->|batch of up to 25x entries| blocks-cars-position[DynamoDB blocks-cars-position table];
  end
  multihashes-queue-->|...| IPNI;
```

The issue is:

1. The existing multihashes SQS queue expects 1 multihash per message and using batch send will only allow 10 messages max.
2. Writing to the DynamoDB table allows only 25 puts to be batched.

This times out for indexes of many thousands of entries and causes the API Gateway service to respond with 503.

### After

```mermaid
graph TD;
  subgraph w3infra
  upload-api[upload-api index/add handler]-->|1 message with up to 3000x multihashes| block-advert-publisher-queue[SQS block-advert-publisher-queue];
  block-advert-publisher-queue-->block-advert-publisher-consumer(ʎ-consumer);
  upload-api-->|1 message with up to 2000x index entries| block-index-writer-queue[SQS block-index-writer-queue];
  block-index-writer-queue-->block-index-writer-consumer(ʎ-consumer);
  end
  subgraph E-IPFS
  block-advert-publisher-consumer-->|300x batch of 10 multihashes| multihashes-queue[SQS multihashes-queue];
  block-index-writer-consumer-->|80x batch of 25 entries| blocks-cars-position[DynamoDB blocks-cars-position table];
  end
  multihashes-queue-->|...| IPNI;
```

The solution is to create queues that allow messages to be sent to them with _many items per message_. The queue consumers then use batch sends to write to existing E-IPFS mutlihashes queue and dynamo table.

Each lambda consumer gets 15 minutes to execute which should be more than enough time to write to their targets.

**Note this essentially re-introduces an unknown async time between index/add and the data becoming available on the gateway due to the queue.** ...which I had hoped to avoid with the transition to the blob protocol.


sst.config.ts
```diff
@@ -46,6 +48,8 @@ export default {
     app.stack(UcanInvocationStack)
     app.stack(BillingStack)
     app.stack(FilecoinStack)
+    app.stack(ElasticIpfsStack)
+    app.stack(IpniStack)
```
vasco-santos, line 52: deploys are already so long, I would suggest to avoid adding more stacks. It also makes it closer to limits of AWS like roles, etc.

Perhaps a middle ground solution could be to have an `IndexingStack`, where both these would live? by the end of the day, the `ElasticIpfsStack` is just references to other infra and hopefully we can remove that code soon.

test/helpers/up-client.js
```diff
@@ -55,14 +55,18 @@ export async function setupNewClient (uploadServiceUrl, options = {}) {
...
+    (async () => {
+      // click link in email
+      const latestEmail = await mailslurp.waitForLatestEmail(inboxId, timeoutMs)
+      const authLink = getAuthLinkFromEmail(latestEmail.body, uploadServiceUrl)
+      const res = await fetch(authLink, { method: 'POST' })
+      if (!res.ok) {
+        throw new Error('failed to authenticate by clickling on auth link from e-mail')
+      }
+    })()
+  ])
```
alanshaw, line 69: For me, waiting for an email to arrive (which never did) was smothering the already rejected `authorizePromise`.

sst.config.ts
alanshaw: Hmm, yes indexer stack is a good shout and doesn't tie to IPNI.

hannahhoward:
@hannah will review today!

hannahhoward (approved):
Generally, this works as a "Right now" Solution, still reviewing the code, but appears to solve the immediate problem, even if it re-upps the urgency of nailing the indexing in a better way. More review to come but would not block.

hannahhoward (commented):
Second run through, still LGTM. There's actually a lot more moving around existing than I thought other than the 2 new queues x 2 new lambdas combo, but that is straight forward.

I agree with Vasco on naming ... and I don't fully understand the diff between EIPFSService & IPNIService

alanshaw:
> ... and I don't fully understand the diff between EIPFSService & IPNIService

The IPNI stack was consuming the E-IPFS stack. I just separated them out incase we wanted to use any E-IPFS stack components in other stacks. In hindsight I think it's unlikely.

In general we separate out stacks to avoid circular dependencies. e.g. we have an `upload-api` stack and a `upload-db`. The `billing` stack _uses_ the `upload-db` stack. The `upload-api` stack uses the `billing` stack AND the `upload-db` stack. This wouldn't be possible if the DB was not a separate stack.

PR #399, opened 2024-07-04
feat(billing): reuse DynamoDB and SQS clients

alanshaw:
As per recommendations in [Best Practices with Lambda](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.Lambda.BestPracticesWithDynamoDB.html) this PR ports a small chunk of code from content claims that caches DynamoDB and SQS clients after they are created. This allows existing connections to be reused for non-cold start invocations and should lower lambda running time.


hannahhoward (approved):
LGTM

PR #401, opened 2024-07-05
feat: reuse all AWS clients

alanshaw:
This is a fairly mechanical change to replace all calls to `new SQSClient/DynamoDBClient/S3Client` with a call to a function that creates and instance and caches it - similarly to https://github.com/storacha-network/w3infra/pull/399

It also moves `mustGetEnv` into the same shared lib folder - we have copied this function an insane amount of times, it was very wet, not DRY at all.


alanshaw:
Going to merge before becomes stale. Please shout if any objections.

PR #403, opened 2024-07-05
fix(filecoin): increase metrics consumer parallelization factor

alanshaw:
The filecoin metrics consumer has been observed to take up to 10 seconds to execute and appears to be the reason that iterator age is continuously increasing.

This PR increases the parallelization factor so that multiple executions can consume the stream in parallel. It also increases the timeout for the lambda in line with other timeouts in the same stack. This is because errors occur when it takes 10 seconds, and it does not ever take longer - hinting that it sometimes needs longer.

Note: I believe it is fine to do this in parallel because the lambda simply issues increment (ADD) instructions to dynamo for the metrics, not SET.

Screenshot 2024-07-05 at 11 51 40

Screenshot 2024-07-05 at 11 55 30


hannahhoward (approved):
LGTM

PR #406, opened 2024-07-15
feat(billing): add consumer list cache to UCAN stream consumer

alanshaw:
This adds an LRU cache to the UCAN stream consumer for billing to allow us to process data for spaces with frequent writes a lot quicker (will also reduce our dynamo reads bill a little).

Previously we filtered lambda invocations items to only process receipts (instead of also invocations, which we simply dropped). This was working for 3 days but now iterator age has started increasing again.

Screenshot 2024-07-15 at 11 26 53

The lambda simply reads from the consumer table and writes to the space diff table. The data from the consumer table almost never changes (it actually cannot right now - we haven't built this feature yet) so an in-memory cache should speed up execution for spaces with high write frequency. We're currently seeing runtimes of ~250ms (when we actually need to do some work - i.e. a lot of lambda invocations are not doing any space diff accounting because they are for receipts that are not for `store/add` or `blob/add`):

Screenshot 2024-07-15 at 11 29 58

### DIFF
--- a/billing/functions/ucan-stream.js
+++ b/billing/functions/ucan-stream.js
@@ -114,7 +114,7 @@ const withConsumerListCache = (consumerStore) => {
       const cached = cache.get(cacheKey)
       if (cached) return cached
       const res = await consumerStore.list(key, options)
-      if (res.ok) cache.set(key.consumer, res)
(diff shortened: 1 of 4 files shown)


alanshaw:
Going to merge - please shout if objections.

PR #407, opened 2024-07-15
feat: add filters for metrics consumers

alanshaw:
The admin/space metrics deal only with receipts and discard anything that is not a receipt.

This should (hopefully) reduce iterator age and lambda invocations bill.

e.g. current admin metrics iterator age:

Screenshot 2024-07-15 at 12 45 22

### DIFF
--- a/stacks/upload-api-stack.js
+++ b/stacks/upload-api-stack.js
@@ -6,7 +6,7 @@ import {
   use
 } from 'sst/constructs'
 
-import { StartingPosition } from 'aws-cdk-lib/aws-lambda'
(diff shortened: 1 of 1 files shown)


vasco-santos:
@alanshaw could you test this works? I have tried this in the past and was not possible, but maybe it got fixed when we upgraded SST

alanshaw:
It does work!

This is the config shown in the console for the UCAN stream consumer for the billing stack:

Screenshot 2024-07-15 at 14 53 22

...and when I deployed this our iterator age started reducing significantly.

PR #408, opened 2024-07-15
feat: partition kinesis messages by task CID

alanshaw:
This PR adds a partition key to kinesis messages that allows consumers to consume messages in parallel using `parallelizationFactor`.

Context: We added `parallelizationFactor` but it had no effect. This is because we set the same partition key for every message.


alanshaw:
Going to merge this so we can make progress.

PR #415, opened 2024-08-02
feat: return receipt bytes do not redirect

alanshaw:
Avoids weird issues with CORS.


alanshaw:
Ideally we don't want to do this.

PR #416, opened 2024-08-19
fix: disable UCAN log expiration

travis:
it looks like we set this bucket up to expire items - in retrospect I think we should have created a separate bucket for the UCAN log because this stream log store bucket was probably originally meant to store streaming items temporarily

disable expiration for now, but uncomment with a note


hannahhoward (approved):
LGTM

PR #422, opened 2024-09-20
Add a manual put firehose

hannahhoward:
# Goals

Add a second firehose for putting records manually. This is for replaying the kinesis event stream.

I turned prettier on in my editor that's why this is so long but the only thing that actually changed is the new manual delivery stream.


hannahhoward:
closing, this is just a one off and shouldn't live in the permaneant infra setup.

PR #424, opened 2024-09-27
feat: add PSA stack

alanshaw:
Adds a temporary stack for the old Pinning Service API (PSA) data.

The stack creates 2 functions:

- `hash` given a root CID, find a CAR file in one of the configured buckets and calculate it's hash.
- `download` given a root CID, find a CAR file in one of the configured buckets and return a signed URL allowing temporary access to the data.

This is for the PSA migration tool in console.

1. Users list pinned root CIDs.
2. Users get hash of CAR file containing pinned data by calling `hash`.
3. Users `blob/add` CAR file.
4. (Maybe) users need to upload the CAR, so they call `download` to get a signed URL and then download the data, and upload it to Storacha.

### DIFF
--- a/psa/config.js
+++ b/psa/config.js
@@ -1,0 +1,26 @@
+import { base32 } from 'multiformats/bases/base32'
+
+/** @type {import('../lib.js').Bucket[]} */
+export const buckets = [
+  {
+    name: process.env.S3_DOTSTORAGE_0_BUCKET_NAME,
+    region: process.env.S3_DOTSTORAGE_0_BUCKET_REGION,
+    toKey: root => {
+      const s = root.toV1().toString(base32)
+      return `complete/${s}/${s}.car`
+    }
+  },
+  {
+    name: process.env.S3_DOTSTORAGE_1_BUCKET_NAME,
+    region: process.env.S3_DOTSTORAGE_1_BUCKET_REGION,
+    toKey: root => {
+      const s = root.toV1().toString(base32)
+      return `complete/${s}/${s}.car`
+    }
+  },
+  {
+    name: process.env.S3_PICKUP_BUCKET_NAME,
+    region: process.env.S3_PICKUP_BUCKET_REGION,
+    toKey: r => `pickup/${r}/${r}.root.car`
+  }
+]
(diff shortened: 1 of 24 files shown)


hannahhoward:
hahhah PSA
image

psa/config.js
```diff
@@ -0,0 +1,26 @@
+import { base32 } from 'multiformats/bases/base32'
+
+/** @type {import('../lib.js').Bucket[]} */
+export const buckets = [
+  {
+    name: process.env.S3_DOTSTORAGE_0_BUCKET_NAME,
+    region: process.env.S3_DOTSTORAGE_0_BUCKET_REGION,
```
hannahhoward, line 7: is this just the names we have 0 & 1?

psa/functions/download.js
```diff
@@ -0,0 +1,22 @@
...
+
+  let root
+  try {
+    root = Link.parse(searchParams.get('root'))
+  } catch (err) {
+    return errorResponse('Invalid "root" search parameter', 400)
+  }
+
+  try {
+    const url = await getDownloadURL(buckets, root)
```
hannahhoward, line 17: buckes doesn't need an import?

stacks/psa-stack.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const downloadFunction = new Function(stack, 'download', {
+    handler: 'shardutil/functions/download.handler',
+    url: { cors: true, authorizer: 'none' }
+  })
+
+  downloadFunction.attachPermissions(['s3:GetObject'])
+
+  stack.addOutputs({
+    hashFunctionURL: hashFunction.url,
```
hannahhoward, line 31: do these need a domain name?

hannahhoward (approved):
LGTM though my personal knowledge of old web3 storage is not sufficient to tell if this is the right place to look for car files + hashes.

psa/config.js
alanshaw: Yeah, for historical reasons we had 2 buckets that received some pinning service data. They are in different regions.

psa/functions/download.js
alanshaw: It certainly does!

stacks/psa-stack.js
alanshaw: Yes, good shout - otherwise they will change when we deploy.

alanshaw: Going to have to setup a redirect for this manually since we want to allow execution of >30s (hashing big files). The AWS API gateway has a limit of 30s, which I think is now configurable, but I don't think our SST version supports it yet.

PR #429, opened 2024-10-09
chore: also log message receipt

travis:
so we can try to find the messages associated with these invocations and receipts


alanshaw (approved):

travis:
I realized this isn't really necessary, as we can get to the message ID by looking in the invocation-store bucket using the invocation CID I'm already logging

PR #435, opened 2024-11-12
fix: consumerTable.get returns customer

fforbeck:
**Context**

While testing the Egress Tracking System in Staging, an error occurred indicating that the router couldn't resolve the customer:

```
✘ [ERROR] Failed to record egress for space did:key:z6MkkYoZNccuPxDxc8kh6YgPSZwyXCqeS2QYqhtQtYyEKYTt {
    name: 'FieldError',
    message: 'Object contains invalid field "customer":\n' +
      '  - Expected value of type string instead got undefined'
    ...
```

**Issue**

The error arises because the `consumerTable.get` query does not include the `customer` field in its projection, despite the customer existing in `consumerTable` and being manually retrievable.

**Fix**

This PR updates the `consumerTable.get` query to ensure the `customer` field is projected correctly, allowing the router to resolve the customer as expected. It also fixes the related tests to reflect this change accurately.

### DIFF
--- a/billing/test/lib/egress-traffic.js
+++ b/billing/test/lib/egress-traffic.js
@@ -112,7 +112,7 @@ export const test = {
         }, {
           retries: maxRetries,
           minTimeout: delay,
-          factor: 2,
+          factor: 3,
           shouldRetry: err => (err instanceof Error && err.message === 'No aggregated meter event found yet'),
           onFailedAttempt: err => {
             console.log(`${err.message} - Attempt ${err.attemptNumber} failed. There are ${err.retriesLeft} retries left.`);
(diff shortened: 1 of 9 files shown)


alanshaw (approved):

billing/test/lib/egress-traffic.js
```diff
@@ -101,9 +121,36 @@ export const test = {
...
+
+/**
+ * Retry a function with exponential backoff
+ * 
+ * @param {Function} fn - The function to retry
+ * @param {number} maxRetries - Maximum number of retries
+ * @param {number} initialDelay - Initial delay in milliseconds
+ * @returns {Promise<any>} - The result of the function if successful
+ */
+async function retryWithExponentialBackoff(fn, maxRetries, initialDelay) {
```
travis, line 139: we use the `p-retry` library for this in a lot of places, would that work here?

travis (approved):
This is good to merge, but if you could swap out the exponential retry function for `p-retry` it'd be even good-er to merge.

billing/test/lib/egress-traffic.js
fforbeck: Good call. That works. I just updated the code with the `p-retry` lib. Thanks for flagging.

PR #436, opened 2024-11-13
fix: drop consumer index

fforbeck:
This PR modifies the consumerTable SST code to remove the existing `consumer` GSI configuration, causing the stack to delete the GSI.

We need to delete it first, and then we can recreate it with the same name but with the additional projected attributes.


fforbeck:
Replaced by https://github.com/storacha/w3infra/pull/438

PR #437, opened 2024-11-13
fix: create consumer index

fforbeck:
Once the GSI is deleted (via #436), we can merge this PR to reconfigure the `consumer` index from the `consumer` table with the new projected attributes: `provider, subscription, customer`.

We need this workaround because we can't simply modify an existing GSI.


fforbeck:
Replaced by https://github.com/storacha/w3infra/pull/438

PR #442, opened 2024-11-20
feat: add upload root CID sampler

alanshaw:
Adds a new endpoint `/sample` which outputs a `dag-json` encoded array of root CIDs registered by users, filtered by codecs the gateway renders natively (`dag-pb` and `raw`).

The endpoint accepts a `?size=` querystring to allow up to 10 CIDs to be retrieved.

This can be used for retrieval testing.


fforbeck (approved):
LGTM!

PR #443, opened 2024-11-22
fix(egress-tracking): new partition and sort key for dynamo table

fforbeck:
**Context**
I've identified an issue in our DynamoDB table where egress events were not being stored correctly. This was traced back to the configuration of our partition key (PK) and sort key (SK). In DynamoDB, the partition key is crucial for uniquely identifying each item, while the sort key helps organize data within a partition. 

This change ensures that the combination of partition key and sort key is unique for each record, preventing DynamoDB from overriding existing records in case of a collision.

**Changes Made**

1. **Modification of Partition Key (PK) & Sort Key (SK):** 
   - The PK was updated to ensure it uniquely identifies each record. Previously, key collisions occurred because multiple records were using the same PK value, leading to storage issues.

2. **Inclusion of `cause` in Stripe Idempotent Key:**
   - Added `cause` as part of the Stripe idempotent key to ensure the uniqueness of requests to Stripe, preventing duplicate processing.

3. **Renaming of Table from `egress-traffic` to `egress-traffic-events`:**
   - Since existing partition keys and sort keys cannot be updated, the table was renamed and the keys were updated. The existing table will be dropped to accommodate these changes.

**Impact**
- These changes resolve the storage issues for egress events and improve data integrity and retrieval efficiency.
- The renaming and restructuring of the table ensure that future records are stored and managed correctly.

Blocked by https://github.com/storacha/w3up/pull/1588

### DIFF
--- a/billing/test/helpers/context.js
+++ b/billing/test/helpers/context.js
@@ -191,7 +191,7 @@ export const createEgressTrafficTestContext = async () => {
     })
   }
 
-  const egressTrafficTable = await createTable(awsServices.dynamo.client, egressTrafficTableProps, 'egress-traffic-')
+  const egressTrafficTable = await createTable(awsServices.dynamo.client, egressTrafficTableProps, 'egress-traffic-events-')
   const egressTrafficEventStore = {
     ...createEgressTrafficEventStore(awsServices.dynamo.client, { tableName: egressTrafficTable }),
     ...createStorePutterClient(awsServices.dynamo.client, {
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "packageManager": "npm@10.8.2+sha256.c8c61ba0fa0ab3b5120efd5ba97fdaf0e0b495eef647a97c4413919eda0a878b",
   "type": "module",
   "scripts": {
-    "start": "sst start",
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "packageManager": "npm@10.8.2+sha256.c8c61ba0fa0ab3b5120efd5ba97fdaf0e0b495eef647a97c4413919eda0a878b",
   "type": "module",
   "scripts": {
-    "start": "sst start",
+    "start": "sst dev",
```
fforbeck, line 8: The `sst start` command is deprecated.

travis (approved):

billing/test/helpers/context.js
```diff
@@ -191,7 +191,7 @@ export const createEgressTrafficTestContext = async () => {
     })
   }
 
-  const egressTrafficTable = await createTable(awsServices.dynamo.client, egressTrafficTableProps, 'egress-traffic-')
+  const egressTrafficTable = await createTable(awsServices.dynamo.client, egressTrafficTableProps, 'egress-traffic-events-')
```
alanshaw, line 194: Why dash suffix?

PR #452, opened 2024-12-11
feat: add free trial pricing table to emails when appropriate

travis:
We are now offering free trials to users who come referred by other users, so we need to be able to display an alternate pricing table to those users.

This PR adds a check to the "email validated" screen that determines whether to offer users a free trial. It queries the referral service and renders the "free trial" pricing table if the user was referred.

This PR also removes some log lines that I accidentally introduced in a recent PR.


fforbeck (approved):
LGTM!

upload-api/functions/validate-email.jsx
```diff
@@ -215,21 +216,32 @@ export async function validateEmailPost(request) {
...
-  console.log("CHECKED PLAN", JSON.stringify(planCheckResult))
 
   if (!planCheckResult.ok?.product) {
-    stripePricingTableId = context.stripePricingTableId
     stripePublishableKey = context.stripePublishableKey
+    if (isReferred) {
+      stripePricingTableId = context.stripeFreeTrialPricingTableId
+    } else {
+      stripePricingTableId = context.stripePricingTableId
+    }
```
Peeja, line 243: **suggestion:** …I'm not totally sure what generates these errors for us, but whatever it is, I agree with it. 😄 Let's prefer the ternary to mutating a variable.

upload-api/html-storacha/index.jsx
```diff
@@ -220,9 +222,24 @@ export const ValidateEmail = ({
...
+              <>
+                <p>
+                  Congratulations! You are eligible for a free trial of our Lite or Business subscriptions. That means
+                  we won&apos;t charge you anything today.
+                  If you choose a Lite plan, you will get two months for free! If you choose Business, you will get one month for free!
+                  We do need you to provide a valid credit card before we can start your
+                  trial - pick a plan below and complete the checkout flow to get started!
+                </p>
+                <p>
+                  Please note that after your free trial ends, you will be charged 10 USD per month for Lite or 100 USD per month for Business tier.
```
Peeja, line 235: **question:** Is it safe to write these values here literally? They won't update if the pricing changes.

Peeja (approved):
Suggestions, but not blocking. I'll leave it to your judgement.

fforbeck (approved):

PR #453, opened 2025-01-13
feat!: consolitate upload service

alanshaw:
This PR implements https://github.com/storacha/RFC/pull/38

Env updates for staging/production deployment:

* rename `CONTENT_CLAIMS_DID` => `INDEXING_SERVICE_DID` and update value
* rename `CONTENT_CLAIMS_URL` => `INDEXING_SERVICE_URL` and update value
* set `INDEXING_SERVICE_PROOF` as SST secret (for filecoin publish `assert/equals` and `assert/index`)
* update `PROVIDERS` to include `did:web:up.storacha.network`
* update `ACCESS_SERVICE_URL` to `https://up.storacha.network`
* add `UPLOAD_API_ALIAS` and set to `did:web:web3.storage`

Depends on:

* [x] https://github.com/storacha/upload-service/pull/124


stacks/ucan-invocation-stack.js
```diff
@@ -81,15 +75,13 @@ export function UcanInvocationStack({ stack, app }) {
   })
 
   stack.addOutputs({
-    workflowBucketName: workflowBucket.bucketName,
-    invocationBucketName: invocationBucket.bucketName,
-    taskBucketName: taskBucket.bucketName
+    agentMessageBucketName: agentMessageBucket.bucketName,
+    agentIndexBucketName: agentIndexBucket.bucketName,
```
alanshaw, line 79: Simply renamed these buckets in the code so they make more sense.

stacks/upload-db-stack.js
```diff
@@ -28,16 +30,21 @@ export function UploadDbStack({ stack, app }) {
...
    * The allocation table tracks allocated multihashes per space.
    * Used by the blob/* service capabilities.
    */
   const allocationTable = new Table(stack, 'allocation', allocationTableProps)
 
+  /**
+   * The blob registry table contains information about blob registrations
+   * per space.
+   */
+  const blobRegistryTable = new Table(stack, 'blob-registry', blobRegistryTableProps)
```
alanshaw, line 46: This table will eventually replace the `allocation` table after migration.

Mostly a rename, but it is also written to at different points in the upload flow. The `allocation` table was written to after `(web3.storage/)blob/allocate` was invoked, _before_ the file was received. The `blob-registry` is written to after a successful `blob/accept` invocation.

stacks/upload-db-stack.js
```diff
@@ -105,8 +112,14 @@ export function UploadDbStack({ stack, app }) {
    */
   const spaceMetricsTable = new Table(stack, 'space-metrics', spaceMetricsTableProps)
 
+  /**
+   * This table tracks storage providers in the system.
+   */
+  const storageProviderTable = new Table(stack, 'storage-provider', storageProviderTableProps)
```
alanshaw, line 118: Every storage provider has a DID, a URL and a proof (a delegation to the upload service allowing `blob/allocate` and `blob/accept` invocations). They also have a weight, which allows some providers to be picked more often than others.

upload-api/external-services/router.js
```diff
@@ -0,0 +1,97 @@
...
+ * @param {import('@ucanto/interface').Signer} serviceID
+ * @returns {BlobAPI.RoutingService}
+ */
+export const create = (storageProviderTable, serviceID) => ({
+  selectStorageProvider: async () => {
+    const ids = await storageProviderTable.list()
+    if (!ids.length) return error(new CandidateUnavailableError())
+    const provider = parse(ids[getWeightedRandomInt(ids.map(id => id.weight))].provider)
+    return ok(provider)
+  },
```
alanshaw, line 22: Main routing logic.

upload-api/metrics.js
```diff
@@ -1,38 +1,28 @@
...
- * - BLOB_ADD_TOTAL: increment number of `blob/add` success receipts
- * - BLOB_ADD_SIZE_TOTAL: increment size of `blob/add` success receipts
  * - STORE_ADD_TOTAL: increment number of `store/add` success receipts
  * - STORE_ADD_SIZE_TOTAL: increment size of `store/add` success receipts
- * - UPLOAD_ADD_TOTAL: increment number of `upload/add` success receipts
- * - BLOB_REMOVE_TOTAL: increment number of `blob/remove` success receipts
- * - BLOB_REMOVE_SIZE_TOTAL: increment size of `blob/remove` success receipts
  * - STORE_REMOVE_TOTAL: increment number of `store/remove` success receipts
  * - STORE_REMOVE_SIZE_TOTAL: increment size of `store/remove` success receipts
- * - UPLOAD_REMOVE_TOTAL: increment number of `upload/remove` success receipts
```
alanshaw, line 32: Only store protocol related metrics are driven by the ucan stream now. When store protocol is decommissioned the ucan stream can also be removed.

upload-api/stores/agent/stream.js
```diff
@@ -82,6 +82,25 @@ export const write = async (connection, message) => {
...
+    // else find in store
+    const result = await Store.getInvocation(store, task)
+    if ('error' in result) {
+      console.error(result.error)
+      throw new NoInvocationFoundForGivenReceiptError(
+        `missing invocation: ${task}`
+      )
+    }
+    return result.ok
+  }
```
alanshaw, line 102: This change simply allows the invocation we're trying to find to be contained in the agent message. i.e. invocation and receipt in the same message. This wouldn't usually happen but in the blob protocol we create fake invocations to allow the old client to work with the new capabilities (e.g. old client expects a `web3.storage/blob/allocate` invocation but it has become `blob/allocate`).

upload-api/stores/blob-registry.js
```diff
@@ -0,0 +1,515 @@
...
+}
+
+/**
+ * @deprecated
+ * @param {BlobAPI.Registry} registry
+ * @param {import('@aws-sdk/client-dynamodb').DynamoDBClient} dynamoDb
+ * @param {string} tableName
+ * @returns {BlobAPI.Registry}
+ */
+export const useAllocationTableBlobRegistry = (registry, dynamoDb, tableName) => ({
```
alanshaw, line 377: Temporary until migration from `allocation` table to `blob-registry` completes.

upload-api/stores/blob-registry.js
```diff
@@ -0,0 +1,515 @@
...
+        await Promise.all([
+          metrics.space.incrementTotals({
+            [SPACE_METRICS_NAMES.BLOB_REMOVE_TOTAL]: [{ space, value: 1 }],
+            [SPACE_METRICS_NAMES.BLOB_REMOVE_SIZE_TOTAL]: [{ space, value: blobSize }]
+          }),
+          metrics.admin.incrementTotals({
+            [METRICS_NAMES.BLOB_REMOVE_TOTAL]: 1,
+            [METRICS_NAMES.BLOB_REMOVE_SIZE_TOTAL]: blobSize
+          })
+        ])
```
alanshaw, line 280: Note: metrics are now recorded here, instead of from the UCAN stream.

upload-api/tables/upload.js
```diff
@@ -134,13 +144,26 @@ export function useUploadTable(dynamoDb, tableName) {
...
+      // if new, increment total
+      if (raw.insertedAt === raw.updatedAt) {
+        await Promise.all([
+          metrics.space.incrementTotals({
+            [SPACE_METRICS_NAMES.UPLOAD_ADD_TOTAL]: [{ space, value: 1 }]
+          }),
+          metrics.admin.incrementTotals({
+            [METRICS_NAMES.UPLOAD_ADD_TOTAL]: 1,
+          })
+        ])
```
alanshaw, line 157: Note: upload metrics now recorded here, instead of via the UCAN stream.

Also more accurate - upload is mutable - we only want to count on the insert, not also update.

PR #456, opened 2025-01-17
fix: multiple space diffs

BravoNatalie:
### Description:
This PR addresses the issue https://github.com/storacha/project-tracking/issues/199, where duplicate space diff entries were being created when uploading a file using `space/blob/add`.

### Problem:
**Context:** When the UCAN invocation router receives a request, it executes the invocation and writes the receipts to the stream, which is then processed by `billing/ucan-stream.js` to populate the diff table.

Unlike `store/add`, `space/blob/add` spawns three additional invocations: `web3.store/blob/allocate`, `http/put`, and `web3.store/blob/accept` before finalizing.  
Due to the way receipts are extracted from agent messages—which are a collection of invocations and receipts without a clear hierarchy—this structure causes the receipt from `web3.store/blob/allocate` to be sent to the stream twice:  
1. Once when it is first invoked.  
2. Again because the `blob/allocate` receipt is embedded within the `blob/add` receipt.  

### Solution:
Change the capability used to track space usage deltas from `BlobAllocate` to `BlobAccept`, as `BlobAccept` does not appear multiple times.

### DIFF
--- a/billing/lib/ucan-stream.js
+++ b/billing/lib/ucan-stream.js
@@ -82,7 +82,6 @@ export const storeSpaceUsageDeltas = async (deltas, ctx) => {
     // could have multiple providers for the same consumer (space).
     const consumers = consumerList.ok.results
     console.log(`Found ${consumers.length} consumers for ${delta.resource}`)
-    console.log(`agentMessage (cause): ${delta.cause}`)
     for (const consumer of consumers) {
       diffs.push({
         provider: consumer.provider,
--- a/billing/lib/ucan-stream.js
+++ b/billing/lib/ucan-stream.js
@@ -19,9 +19,9 @@ export const findSpaceUsageDeltas = messages => {
     let resource
     /** @type {number|undefined} */
     let size
-    if (isReceiptForCapability(message, ServiceBlobCaps.allocate) && isServiceBlobAllocateSuccess(message.out)) {
+    if (isReceiptForCapability(message, ServiceBlobCaps.accept) && isServiceBlobAcceptSuccess(message.out)) {
       resource = message.value.att[0].nb?.space
-      size = message.out.ok.size
+      size = message.value.att[0].nb?.blob.size
     } else if (isReceiptForCapability(message, BlobCaps.remove) && isBlobRemoveSuccess(message.out)) {
       resource = /** @type {import('@ucanto/interface').DID} */ (message.value.att[0].with)
       size = -message.out.ok.size
@@ -82,6 +82,7 @@ export const storeSpaceUsageDeltas = async (deltas, ctx) => {
     // could have multiple providers for the same consumer (space).
     const consumers = consumerList.ok.results
     console.log(`Found ${consumers.length} consumers for ${delta.resource}`)
+    console.log(`agentMessage (cause): ${delta.cause}`)
     for (const consumer of consumers) {
       diffs.push({
         provider: consumer.provider,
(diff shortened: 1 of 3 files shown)


billing/lib/ucan-stream.js
```diff
@@ -82,6 +82,7 @@ export const storeSpaceUsageDeltas = async (deltas, ctx) => {
     // could have multiple providers for the same consumer (space).
     const consumers = consumerList.ok.results
     console.log(`Found ${consumers.length} consumers for ${delta.resource}`)
+    console.log(`agentMessage (cause): ${delta.cause}`)
```
alanshaw, line 85:
```suggestion
```

billing/lib/ucan-stream.js
```diff
@@ -19,9 +19,9 @@ export const findSpaceUsageDeltas = messages => {
     let resource
     /** @type {number|undefined} */
     let size
-    if (isReceiptForCapability(message, ServiceBlobCaps.allocate) && isServiceBlobAllocateSuccess(message.out)) {
+    if (isReceiptForCapability(message, ServiceBlobCaps.accept) && isServiceBlobAcceptSuccess(message.out)) {
       resource = message.value.att[0].nb?.space
-      size = message.out.ok.size
+      size = message.value.att[0].nb?.blob.size
```
alanshaw, line 24: Ah, and this works because `value` points to the original _invocation_ for blob accept, which _does_ include the size.

alanshaw: The only problem with this approach is that it doesn't take into account whether the blob is stored in the space already or not. The receipt for `blob/allocate`will have `size: 0` when the blob already exists in the space.

It's probably better than the current situation though...

BravoNatalie: If `size: 0`, we are not adding it to the diff table.

alanshaw: @hannahhoward WDYT shall we merge and release this? IMHO it's better than the status quo but still needs more work.

BravoNatalie: I understand now. I confirmed this in the code, and it's worse than you thought. We're actually inserting duplicate values of `bafyxyz` every time we store it. This happens because in this flow, the `blob/accept` is executed, resulting in two receipts for it.

BravoNatalie: If the blob already exists, we return success for `blob/allocate` with a size of 0 and `out.ok.address` set to `null`. However, if this occurs, we immediately return the receipt for `http/put`, and if it has succeeded, we execute `blob/accept`, leading to two receipts in the stream.

alanshaw: I think in that case we should just not generate those receipts early.

travis: Would love to catch up on this one - if it doesn't come up at sprint planning or we don't have enough time to get into it I'd love to sync briefly after that if y'all are free!

travis: OK - in our chat this morning we decided to ship this change since it will significantly decrease the impact of the duplicate diff issue, and to deprecate the ucan stream entirely as part of the upload-service work - @alanshaw should we create one more item under that upload service epic for this work? I'm happy to go file that and fill in some basic details if so...

alanshaw: I summarized the plan moving forward here: https://github.com/storacha/RFC/pull/39

alanshaw (approved):

PR #457, opened 2025-01-23
feat: a few improvements to the dry-run, allocations and snapshot update scripts

travis:
* simplify configuration by hard coding table names - I thought about trying to extract this to a constant but decided it's at least a separate project if it's worth doing at all - happy to jump into that in a quick followup if anyone strongly prefers it
* fix upsert snapshot script to work for more than 100 snapshots


fforbeck (approved):
LGTM! Do you remember if the batch size for the upsert operation in Dynamo supports 100 items?

travis:
> LGTM! Do you remember if the batch size for the upset operation in Dynamo supports 100 items?

ah yes - it's precisely 100 items iirc

BravoNatalie (approved):
LGTM

PR #460, opened 2025-01-31
fix: two important usage billing fixes

travis:
1) use a CID as an idempotency key - some Stripe usage billing reporting requests were failing because the space DID was much longer than we expected and the idempotency key cannot be more than 255 characters - we were sending ~450 characters in many cases

2) properly configure the "on failure" queue for the usage table handler - the previous config didn't work, and no "on failure" destination was set for this, which is, I suspect, why we can't find any trace of some usage table insert hook invocations


BravoNatalie (approved):
LGTM

alanshaw (approved):

PR #461, opened 2025-02-04
fix: switch to arm_64 architecture

alanshaw:
I did not realise we weren't already using `arm_64`. Turns out this change had been applied to `w3filecoin-infra` and `content-claims` but not here (aside from the PSA stack).

Switching to `arm_64` should reduce lambda costs.


volmedo (approved):

volmedo:
it looks like the test is failing because of a server didn't start on time, so maybe there is some flakiness

PR #462, opened 2025-02-05
feat: configurable principal mapping

alanshaw:
Allows the principal mapping to be configured from environment variables.

### DIFF
--- a/stacks/upload-api-stack.js
+++ b/stacks/upload-api-stack.js
@@ -146,6 +146,7 @@ export function UploadApiStack({ stack, app }) {
             INDEXING_SERVICE_URL,
             INDEXING_SERVICE_PROOF,
             HOSTED_ZONE: hostedZone ?? '',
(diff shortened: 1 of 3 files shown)


stacks/upload-api-stack.js
```diff
@@ -146,6 +146,7 @@ export function UploadApiStack({ stack, app }) {
             INDEXING_SERVICE_URL,
             INDEXING_SERVICE_PROOF,
             HOSTED_ZONE: hostedZone ?? '',
+            PRINCIPAL_MAPPING: process.env.PRINCIPAL_MAPPING ?? ''
```
travis, line 149: could you add an example of what this should look like in the docs and the env.tpl?

https://github.com/storacha/w3infra/blob/main/README.md#environment-variables
https://github.com/storacha/w3infra/blob/main/.env.tpl

travis (dismissed):
I'm also ok with moving docs into a separate task, but would be nice to have this before the new env var merges

PR #463, opened 2025-02-24
feat: login via OAuth

alanshaw:
refs https://github.com/storacha/RFC/pull/43

This PR adds a GitHub OAuth callback handler. It expects an `access/authorize` delegation to be passed in the `state` querystring paremeter. It completes the login flow by issuing a `access/confirm` invocation on behalf of the user, which in turn issues the requested delegations and attestations.

If the user is a new user they will be automatically enrolled in the "trial" plan (`did:web:trial.storacha.network`). This plan does not require payment details and so their `account` in the `customer` table is blank.

The billing run will not consider these customers and neither will egress accounting.

New signups with GitHub accounts must be at least 1 day old.


fforbeck (approved):
That looks great! I was wondering how we would link the Stripe customer with the GitHub account to be able to track the storage usage. I know it is not part of the scope, but just picturing how that would work. Is it correct to assume you solved that by adding the `details` attribute to the CustomerStore schema, so later we can query all the customers with free trial plans, get their emails, create the customer record in Stripe, and then run the report to calculate their storage usage, and somehow block and prompt them the paywall?

alanshaw:
As discussed, we store egress data in our own table so we don't need to send anything to Stripe before the user creates a Stripe account. This is also quite nice because it decouples from using Stripe as our only payment provider.

PR #464, opened 2025-04-07
fix: authorize gateway in integration tests

alanshaw:


alanshaw:
NFT 10k test should be fixed when https://github.com/storacha/indexing-service/pull/192 is released.

PR #466, opened 2025-04-10
feat: add Humanode verification callback

travis:
Add a Humanode OAuth callback that will give a "free storage with no credit card" plan to a user who uses Humanode to prove they are a unique person.

Screenshot 2025-04-10 at 3 42 22 PM
Screenshot 2025-04-10 at 3 43 09 PM
Screenshot 2025-04-10 at 3 44 11 PM
Screenshot 2025-04-10 at 3 43 19 PM

The one remaining concern I have with this code is the duplication in the templates - we now have 4 blobs of HTML like this across the two OAuth handlers and I'm not sure the best way to consolidate - any thoughts @alanshaw ?


hannahhoward (approved):
This looks pretty good, outside my domain but seems reasonable.

I would leave DRYing up templates / oath code w/ github for another PR.

PR #467, opened 2025-04-16
feat: capture x-client header for analytics

fforbeck:
### Context
We recently added the ability to pass custom headers to identify which Storacha client uses the Upload API (see: https://github.com/storacha/upload-service/pull/215). Now we need to capture the custom header for analytics.

### Changes
- Updated the `ucan-invocation-handler` to log additional information if the `X-Client` custom header is present. 
- The logs are structured in JSON format that's easy to query in CloudWatch Logs.
- The `clientId` field will contain the actual value of the `X-Client` header which indicates the name and version of the Storacha Client + service/project it is integrated to, e.g: `Storacha/X (js; MCP/Y)` - where `X` =  major version of the Storacha Client, and `Y` =  major version of the service/project integration.
- The `requestId` allows us to correlate this specific log entry with other logs from the same request, which helps debug issues.
- Then we build analytics based on CloudWatch Logs Insights using some query like this:
   ```
   fields @timestamp, clientId
   | filter message = "Client request"
   | stats count(*) as requestCount by clientId
   | sort requestCount desc
   ```

Related to: https://github.com/storacha/mcp-storage-server/issues/18

### DIFF
--- a/upload-api/functions/ucan-invocation-router.js
+++ b/upload-api/functions/ucan-invocation-router.js
@@ -118,17 +118,16 @@ export const knownWebDIDs = {
  * @param {import('aws-lambda').APIGatewayProxyEventV2} request
  */
 export async function ucanInvocationRouter(request) {
-  // Capture X-Client custom header for analytics - ignore if not present
-  const clientId = request.headers['X-Client']
-  if (clientId) {
-    console.log(JSON.stringify({
-      message: 'Client request',
-      clientId,
-      requestId: request.requestContext?.requestId || 'unknown',
-      timestamp: new Date().toISOString()
-    }))
-  }
-  
+  // Capture X-Client custom header for analytics
+  const clientId = Object.entries(request.headers)
+  .find(([key]) => key.toLowerCase() === 'x-client')?.[1] ?? 'Storacha/?'
+  console.log(JSON.stringify({
+    message: 'Client request',
+    clientId,
+    requestId: request.requestContext?.requestId || 'unknown',
+    timestamp: new Date().toISOString()
+  }))
+
   const {
     storeTableName,
     storeBucketName,
@@ -216,7 +215,7 @@ export async function ucanInvocationRouter(request) {
       credentials: {
         accessKeyId: carparkBucketAccessKeyId,
         secretAccessKey: carparkBucketSecretAccessKey,
-      }, 
+      },
     }),
     createBlobsStorage(AWS_REGION, storeBucketName),
   )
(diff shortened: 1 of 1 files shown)


upload-api/functions/ucan-invocation-router.js
```diff
@@ -118,6 +118,17 @@ export const knownWebDIDs = {
...
+  // Capture X-Client custom header for analytics - ignore if not present
+  const clientId = request.headers['X-Client']
+  if (clientId) {
+    console.log(JSON.stringify({
+      message: 'Client request',
+      clientId,
+      requestId: request.requestContext?.requestId || 'unknown',
+      timestamp: new Date().toISOString()
+    }))
+  }
```
alanshaw, line 130: I would just always log a value so that we can easily count known values as a percentage of all requests.

```suggestion
  // Capture X-Client custom header for analytics
  const clientId = request.headers['X-Client'] ?? 'Storacha/?'
  console.log(JSON.stringify({
    message: 'Client request',
    clientId,
    requestId: request.requestContext?.requestId || 'unknown',
    timestamp: new Date().toISOString()
  }))
```

upload-api/functions/ucan-invocation-router.js
```diff
@@ -118,6 +118,17 @@ export const knownWebDIDs = {
  * @param {import('aws-lambda').APIGatewayProxyEventV2} request
  */
 export async function ucanInvocationRouter(request) {
+  // Capture X-Client custom header for analytics - ignore if not present
+  const clientId = request.headers['X-Client']
+  if (clientId) {
+    console.log(JSON.stringify({
```
alanshaw, line 124: Do you need to strinigfy?

upload-api/functions/ucan-invocation-router.js
```diff
@@ -118,6 +118,17 @@ export const knownWebDIDs = {
  * @param {import('aws-lambda').APIGatewayProxyEventV2} request
  */
 export async function ucanInvocationRouter(request) {
+  // Capture X-Client custom header for analytics - ignore if not present
+  const clientId = request.headers['X-Client']
```
alanshaw, line 122: Just incase `request.headers` is lower-cased or the client sends with weird casing e.g. `X-client`.

```suggestion
  const clientId = new URLSearchParams(request.headers).get('X-Client')
```

alanshaw (changes requested):

upload-api/functions/ucan-invocation-router.js
fforbeck: This doesn't compile because this request.headers is type `APIGatewayProxyEventHeaders` which can't be passed to `URLSearchParams`

alanshaw (approved):

PR #468, opened 2025-04-23
feat: capture x-client header for analytics

fforbeck:
### Context
We recently added the ability to pass custom headers to identify which Storacha client uses the Upload API (see: https://github.com/storacha/upload-service/pull/215). Now we need to capture the custom header for analytics.

### Changes
- Updated the `ucan-invocation-handler` to log additional information if the `X-Client` custom header is present. 
- The logs are structured in JSON format that's easy to query in CloudWatch Logs.
- The `clientId` field will contain the actual value of the `X-Client` header which indicates the name and version of the Storacha Client + service/project it is integrated to, e.g: `Storacha/X (js; MCP/Y)` - where `X` =  major version of the Storacha Client, and `Y` =  major version of the service/project integration.
- The `requestId` allows us to correlate this specific log entry with other logs from the same request, which helps debug issues.
- Then we build analytics based on CloudWatch Logs Insights using some query like this:
   ```
   fields @timestamp, clientId
   | filter message = "Client request"
   | stats count(*) as requestCount by clientId
   | sort requestCount desc
   ```

Related to: https://github.com/storacha/mcp-storage-server/issues/18

### DIFF
--- a/upload-api/functions/ucan-invocation-router.js
+++ b/upload-api/functions/ucan-invocation-router.js
@@ -118,17 +118,16 @@ export const knownWebDIDs = {
  * @param {import('aws-lambda').APIGatewayProxyEventV2} request
  */
 export async function ucanInvocationRouter(request) {
-  // Capture X-Client custom header for analytics - ignore if not present
-  const clientId = request.headers['X-Client']
-  if (clientId) {
-    console.log(JSON.stringify({
-      message: 'Client request',
-      clientId,
-      requestId: request.requestContext?.requestId || 'unknown',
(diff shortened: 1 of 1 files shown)


fforbeck:
@alanshaw I've recreated the PR because I can't see any errors in the seedrun logs & AWS Cloud Watch for the Previous PR (#467) - I want to rebuild the test environment from scratch to see if we get any logs. It was failing with 503 in the Upload API POST request.

alanshaw:
Merge it - the error is unrelated to your work :)

PR #470, opened 2025-04-26
feat(agentstore): put index in dynamo

hannahhoward:
# Goals

I'm not sure why the agent store was setup to use S3 keys as an index of invocations, but the net effect is we end up using a ListObjectsV2 operation, which is a tier 1 op (see private note additional context)

Proposed approach to resolving this issue is:
1. Get upload service deployed to prod before anything
2. Test and merge this PR and through to prod
3. Write a script to move over all the old keys from S3 to Dynamo
4. Once complete, delete the fallback to S3, to remove ListObject operations (among other things, this is called on every single ucan invocation to check for a cached receipt that is usually empty)

# Implementation

- Add a dynamo table to replace the usage of S3 to manage an index of messages by invocation/receipt - folds into existing operations with minimal changes for now
- Convert many references to the agent store to use this dynamo reference.

### DIFF
--- a/stacks/ucan-invocation-stack.js
+++ b/stacks/ucan-invocation-stack.js
@@ -1,6 +1,7 @@
 import {
   Bucket,
   KinesisStream,
+  Table,
 } from 'sst/constructs'
 import { PolicyStatement, StarPrincipal, Effect } from 'aws-cdk-lib/aws-iam'
 
@@ -9,6 +10,7 @@ import {
   getKinesisStreamConfig,
   setupSentry
 } from './config.js'
+import { agentIndexTableProps } from '../upload-api/tables/index.js'
 
 /**
  * @param {import('sst/constructs').StackContext} properties
@@ -41,13 +43,15 @@ export function UcanInvocationStack({ stack, app }) {
       resources: [agentMessageBucket.cdk.bucket.arnForObjects('*')],
     })
   )
-
+  
   const agentIndexBucket = new Bucket(stack, 'invocation-store', {
     cors: true,
     cdk: {
       bucket: getBucketConfig('invocation-store', app.stage)
     }
   })
+
+  const agentIndexTable = new Table(stack, 'invocation-table', agentIndexTableProps)
 
   // TODO: keep for historical content that we might want to process
   new Bucket(stack, 'ucan-store', {
@@ -80,6 +84,7 @@ export function UcanInvocationStack({ stack, app }) {
   })
 
   return {
+    agentIndexTable,
     agentIndexBucket,
     agentMessageBucket,
     ucanStream
--- a/upload-api/functions/oauth-callback.js
+++ b/upload-api/functions/oauth-callback.js
@@ -220,11 +220,15 @@ const getContext = (customContext) => {
 
   const agentStore = openAgentStore({
     store: {
-      connection: { address: { region } },
+      dynamoDBConnection: { address: { region } },
+      s3Connection: { address: { region } },
       region,
       buckets: {
         message: { name: mustGetEnv('WORKFLOW_BUCKET_NAME') },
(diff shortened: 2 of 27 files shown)


upload-api/tables/index.js
```diff
@@ -220,3 +220,15 @@ export const storageProviderTableProps = {
...
 }
+
+
+/** @type TableProps */
+export const agentIndexTableProps = {
+  fields: {
+    task: 'string',
+    kind: 'string',
+    invocation: 'string',
+    message: 'string',
```
alanshaw, line 231: Would be good to have comments explaining what these fields store.

stacks/ucan-invocation-stack.js
```diff
@@ -41,14 +43,16 @@ export function UcanInvocationStack({ stack, app }) {
...
-
+  
   const agentIndexBucket = new Bucket(stack, 'invocation-store', {
     cors: true,
     cdk: {
       bucket: getBucketConfig('invocation-store', app.stage)
     }
   })
 
+  const agentIndexTable = new Table(stack, 'invocation-table', agentIndexTableProps)
```
alanshaw, line 54: Minor, but it's an index of items in agent messages, so I'd probably go for `agent-index` or `agent-message-index` as the table name. Whilst `invocation-table` aligns with the name of the old bucket (`invocation-store`), we have long since moved on from calling this `invocation-store` in the (majority of the) code. `agent-index` aligns with what we're calling it in the code today.

Also, I don't think we need need to suffix with `-table` - all dynamo tables are tables.

```suggestion
  const agentIndexTable = new Table(stack, 'agent-index', agentIndexTableProps)
```

upload-api/functions/oauth-callback.js
```diff
@@ -220,12 +220,16 @@ const getContext = (customContext) => {
...
-      connection: { address: { region } },
+      dynamoDBConnection: { address: { region } },
+      s3Connection: { address: { region } },
       region,
       buckets: {
         message: { name: mustGetEnv('WORKFLOW_BUCKET_NAME') },
         index: { name: mustGetEnv('INVOCATION_BUCKET_NAME') },
       },
+      tables: {
+        index: { name: mustGetEnv('INVOCATION_TABLE_NAME') },
```
alanshaw, line 231: Hmm, should this be `AGENT_INDEX_TABLE_NAME`?

upload-api/stores/lib.js
```diff
@@ -53,13 +54,35 @@ export async function getAgentMessage (messageCid, { workflowBucketName, s3clien
...
+export async function getAgentMessageCidWithInvocation (invocationCid, { dynamoDBTableName, invocationBucketName, s3client, dynamoDBClient, endsWith }) {
   const encodedInvocationKeyPrefix = `${invocationCid.toString()}/`
+
+  // Find agent message archive CID where this receipt was stored
+  // attempt to read first from dynamo
+  const dynamoCommand = new GetItemCommand({
+    TableName: dynamoDBTableName,
+    Key: {
+      "task": {
+        S: encodedInvocationKeyPrefix
```
alanshaw, line 72: I believe these are just CID strings (i.e. no `/` suffix)

```suggestion
        S: `${invocationCid}`
```

upload-api/stores/agent/store.js
```diff
@@ -328,12 +378,31 @@ const toIndexEntry = (path) => {
  */
 const list = async (connection, { prefix, suffix }) => {
   try {
+    // attempt to read first from dynamo
+    const dynamoCommand = new GetItemCommand({
+      TableName: connection.tables.index.name,
+      Key: {
+        "task": {
+          S: prefix
```
alanshaw, line 386: To remove the trailing slash:

```suggestion
          S: prefix.slice(0, -1)
```

alanshaw (changes requested):
🚀

PR #472, opened 2025-05-19
feat: render bsky.storage pricing table

travis:
If the `access/authorize` capability is invoked with `nb.app` set to `bsky-backups`, render the same pricing table we render on bsky.storage, which redirects them to that app rather than console.storacha.network after auth.

See https://github.com/storacha/upload-service/pull/264 for the other half of this.


hannahhoward (approved):
LGTM

PR #478, opened 2025-05-29
fix: legacy space claim publishing

alanshaw:
This PR upgrades the `upload-api` dependency.


alanshaw:
I'm going to merge this so that we can get the warm storage network deployed.

PR #482, opened 2025-06-10
fix: associate storacha account ID with humanode subject

travis:
this will be helpful for administration


BravoNatalie (approved):

alanshaw (approved):
Do we need an index so we can lookup subject by account?

PR #483, opened 2025-06-11
fix: blob deregister diff

alanshaw:
When de-registering a blob the diff size should be a negative number.


alanshaw:
Going to merge this as it's such a small change!

PR #487, opened 2025-06-12
fix: remove unused task store

alanshaw:
Removed in the new deployment but re-adding it back causes a naming conflict.


hannahhoward (approved):
approved, as long as we don't actually have any refs here to TASK_BUCKET_NAME

PR #488, opened 2025-06-12
fix: re-add tables from new deploy

alanshaw:
This PR adds the new tables present in the new deploy to the legacy deploy. They are unused here but should hopefully allow deploy without conflict.


hannahhoward (approved):
LGTM

PR #500, opened 2025-07-23
fix: better handling for upgrade path from "no credit card" tier

travis:
dramatically better handling of stripe webhooks, particularly in cases where a user previous signed up for the "no credit card tier" - previous in those cases the customer records would get overwritten.

now:

1) if a user is upgrading from the no-cc tier, we will preserve their existing customer record 
2) if a user is trying to change the stripe account associated with their customer id we'll throw an error 
3) if a user is trying to change the product associated with the stripe account we know is associated with their account, we'll do that

fixes https://github.com/storacha/ai-integrations/issues/55

### DIFF
--- a/billing/test/utils/stripe.js
+++ b/billing/test/utils/stripe.js
@@ -1,7 +1,8 @@
-import { 
-  accountIDToStripeCustomerID, 
+import {
+  accountIDToStripeCustomerID,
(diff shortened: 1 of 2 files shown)


fforbeck (approved):
LGTM

PR #510, opened 2025-08-30
feat(upload-api): return actual limits in getConsumer

hannahhoward:
# Goals

Enable plans to have enforced limits

# Implementation

- mark plans as allowing overages or not (all current plans for now allow overages)
- add a planLimit lookup based on customer ID which returns a value if an only if a customer has a plan that does not allow overages
- add code to calculate total space used in getConsumer properly, and also add call to get plan limit
- requires shifting around a bunch of dependencies
- pairs with https://github.com/storacha/upload-service/pull/386

# For discussion

I see two potential concerns with the proposed approach:

1. While the using the fields of allocated and limit is convenient cause they already exist, it leads to pretty big mixing of concerns, and a call to getConsumer that does a lot of calculation. It seems to me there should be a function/class that uses the provisionStore and the usageStore to provide a higher level call to get the usage + limit. Also perhaps usage per account should move into the usageStore per https://github.com/storacha/RFC/pull/64
2. Regardless of code quality, I'm concerned about performance calculating usage. I guess one move would be to not calculate usage UNLESS there's a limit on the plan. Again, this seems like it really just needs a different API than getConsumer, even if that was convenient to implement inside spaceAllocate. Is there a way we can cache current usage numbers? It seems like a weird thing to not be able to calculate. Especially since this is really a different question than billing: it's about the total current space used.

### DIFF
--- a/billing/lib/api.ts
+++ b/billing/lib/api.ts
@@ -101,6 +101,7 @@ export type CustomerStore =
   & StoreLister<{}, Customer>
   & {
     updateProduct: (customer: CustomerDID, product: string) => Promise<Result<Unit, Failure>>
+    planLimit: (customer: CustomerDID) => Promise<Result<number, Failure>>
   }
 
 /**
--- a/billing/lib/product-info.js
+++ b/billing/lib/product-info.js
@@ -1,0 +1,12 @@
+const MB = 1024 * 1024
+const GB = 1024 * MB
+const TB = 1024 * GB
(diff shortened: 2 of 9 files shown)


BravoNatalie (changes requested):
I agree with the concerns, especially around using `getConsumer`, since it adds confusion. I think implementing the new account-level usage would make things clearer.

PR #634, opened 2026-04-13
feat(roundabout): add `/piece/{pieceCID}` redirect endpoint

BravoNatalie:
Adds a dedicated GET/HEAD `/piece/{pieceCID}` route that resolves a Piece V2 CID to a presigned content URL via the Indexing Service, returning a `302` redirect. Returns `415` for Piece V1 CIDs and `400`for non-piece CIDs.

### DIFF
--- a/roundabout/functions/redirect.js
+++ b/roundabout/functions/redirect.js
@@ -36,7 +36,7 @@ export async function redirectGet(request) {
   }
 
   /** @type {string[]} */
-  const indexingServiceURLs = JSON.parse(process.env.ROUNDABOUT_INDEXING_SERVICE_URLS || '["https://indexer.storacha.network"]')
+  const indexingServiceURLs = JSON.parse(process.env.ROUNDABOUT_INDEXING_SERVICE_URLS || '[]')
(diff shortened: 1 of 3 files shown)


alanshaw:
@BravoNatalie something to be aware of - there are actually 2 R2 buckets (`carpark-prod-0` and `carpark-prod-1`) and I believe roundabout only considers 1 of them....

roundabout/functions/redirect.js
```diff
@@ -110,6 +110,58 @@ async function resolvePiece (cid, locateContent, indexingService) {
...
+  let indexingService
+  if (indexingServiceURLs.length > 1) {
+    const clients = indexingServiceURLs.map(u => new Client({ serviceURL: new URL(u) }))
+    indexingService = combine(clients)
+  } else {
+    const url = indexingServiceURLs[0] ? new URL(indexingServiceURLs[0]) : undefined
+    indexingService = new Client({ serviceURL: url })
+  }
+
+  const locateContent = contentLocationResolver({
```
alanshaw, line 155: So I think here you might need to check 2 buckets...

alanshaw (approved):

